PR #1, opened 2024-10-19
Introduce reccmp project file concept

madebr:
There are 3 kind of project files:
- `reccmp-project.yml`: should be part of the git repo
    e.g.
    ```yml
    targets:
      CONFIG:
        filename: CONFIG.EXE
        source-root: CONFIG
        hash:
          sha256: 864766d024d78330fed5e1f6efb2faf815f1b1c3405713a9718059dc9a54e52c
      ISLE:
        filename: ISLE.EXE
        source-root: ISLE
        hash:
      LEGO1:
        filename: LEGO1.DLL
        source-root: LEGO1
        hash:
          sha256: 14645225bbe81212e9bc1919cd8a692b81b8622abb6561280d99b0fc4151ce17
    ```
    It contains general information, is part of the repo and can never contain absolute paths.
- `reccmp-user.yml`: should not be committed to the repo
    e.g.
    ```yml
    targets:
      ISLE: # abc
        path: /home/maarten/games/isle/ISLE.EXE
      LEGO1:
        path: /home/maarten/games/isle/LEGO1.DLL
      CONFIG:
        path: /home/maarten/games/isle/CONFIG.EXE
    ```
    It contains all information that is user specific, and should remain unchanged for the duration of a decompilation project (unless you move binaries around)
- `reccmp-build.yml`: should not be committed to the repo
    e.g.
    ```yml
    project: /home/maarten/projects/isle  # directory of reccmp-project.yml
    targets:
      ISLE: # abc
        path: /home/maarten/projects/isle/build/ISLE.EXE
        pdb: /home/maarten/projects/isle/build/ISLE.pdb
      LEGO1:
        path: /home/maarten/games/isle/LEGO1.DLL
        pdb: /home/maarten/games/isle/LEGO1.pdb
      CONFIG:
        path: /home/maarten/games/isle/CONFIG.EXE
        pdb: /home/maarten/games/isle/CONFIG.pdb
    ```
    This file contains paths that can change from build to build.

`reccmp-project` and `reccmp-user.yml` are expected to be in the same directory. `reccmp-build.yml` can be in the build directory, and contains a reference to the project root.

The `reccmp-project` script can generate a new reccmp project with the command `reccmp-project create ...`
The `reccmp-project` script can create `project-user.yml` and `project-build.yml` with the command `reccmp-project detect ...`

I modified the various `reccmp-*` utilities such that it will look for `reccmp-build.yml` in the working directory.
It uses a similar search procedure as git uses for finding a `.git` folder: walk the tree until it find either a `reccmp-build.yml` file, or a `reccmp-project.yml` file, depending on whether it needs recompiled binaries.

There is also cmake support to automatically generate a `reccmp-build.yml` file. The `reccmp.cmake` file should be committed to foreign recompilation projects.

I modified the isle project to make use of the reccmp scripts in [this branch on my fork](https://github.com/madebr/isle/commits/use-reccmp-fork/).
I made it a git submodule in order to be able to run the unittests, that depend on isle executables.
Ideally, the isle project should just need to install `git+https://github.com/isledecomp/reccmp.git`.


foxtacles (approved):
Looks good to me!

disinvite:
Just starting to play with this now. I like the YAML config setup because it nicely solves the problem of "where do you put the files?" that (I assume) we've all had to figure out in our own way.

I'm less sure about the sample code files and cmake stuff. A new recomp project might not use C++ or might be a DOS program that would require a specific compiler under emulation. Any way to disable that from the `reccmp-project` options or have it as its own script?

jonschz:
I can't promise a detailed review before tomorrow, but the approach looks good so far! I'll try and see how well the Ghidra import works.

madebr:
> I'm less sure about the sample code files and cmake stuff. A new recomp project might not use C++ or might be a DOS program that would require a specific compiler under emulation. Any way to disable that from the `reccmp-project` options or have it as its own script?

I added a `--cmake-project` option to the `reccmp-project create` command.

Speaking about DOS. I added a `targets.<target-id>.pdb` key to `reccmp-build.yml`. Perhaps this is a bad name, and should be something else? pdb is MSVC-specific. Watcom provides .sym files.
Perhaps rename it to `debug-path`? This would future proof us for elf support where dwarf information can be either part of the original executable, or split to an external file.

Rename to `targets.<target>.debug-path`?

reccmp/tools/project.py
```diff
@@ -0,0 +1,444 @@
...
+
+    args = parser.parse_args()
+
+    argparse_parse_logging(args=args)
+
+    if args.action == "CREATE":  # FIXME: use enum or callback function
+        return create_project(
+            project_directory=args.create_directory,
+            original_paths=args.originals,
+            cmake=args.create_cmake,
```
disinvite, line 419: Should that be `args.detect_cmake` to match the argument on 385?

reccmp/tools/project.py
```diff
@@ -0,0 +1,444 @@
...
+from reccmp.project.logging import argparse_add_logging_args, argparse_parse_logging
+from reccmp.project.detect import (
+    RECCMP_BUILD_CONFIG,
+    RECCMP_PROJECT_CONFIG,
+    RECCMP_USER_CONFIG,
+    RecCmpProject,
+)
+
+
+TOOLS_DIR = Path(__file__).resolve().parent
```
disinvite, line 23: Did `reccmp.cmake` move to the assets dir in a forced push?

madebr: Yes, I think it's easier to handle all assets in a central `reccmp.assets` module.

reccmp/tools/project.py
madebr: Yes, I changed it to `args.create_cmake`.

reccmp/tools/project.py
```diff
@@ -0,0 +1,448 @@
...
+    if not original_paths:
+        print("Need at least one original binary", file=sys.stderr)
+        return 1
+    id_path = {}
+    project_config = {
+        "targets": {}
+    }
+    for original_path in original_paths:
+        if not original_path.is_file():
+            logger.error(f"Original binary (%s) is not a file", original_path)
```
disinvite, line 56: Should not be f-string

reccmp/tools/project.py
```diff
@@ -0,0 +1,448 @@
...
+        return create_project(
+            project_directory=args.create_directory,
+            original_paths=args.create_originals,
+            cmake=args.create_cmake,
+        )
+
+    if args.action == "DETECT":  # FIXME: use enum or callback function
+        project = RecCmpProject.from_directory(Path.cwd())
+        if not project:
+            parser.error("Cannot find reccmp project. Run %(prog)s create first.")
```
disinvite, line 429: This doesn't support % replacement and doesn't appear to recognize `%(prog)s` so I think this should be:
```python
parser.error(f"Cannot find reccmp project. Run {parser.prog} create first.")
```

reccmp/tools/project.py
```diff
@@ -0,0 +1,457 @@
...
+
+        // FUNCTION: {target_id} 0x10000000
+        SomeClass::~SomeClass() {{
+        }}
+
+        // GLOBAL: {target_id} 0x10102000
+        // STRING: {target_id} 0x10101f00
+        const char* g_globalString = "A global string";
+
+        {entry_function}
```
disinvite, line 351: I think this will work as:
```python
return textwrap.dedent("big string starting with #include") + entry_function
```
This string and `entry_function` should get dedented separately and then added together. In the current code, the top of the .cpp file keeps its indents. The lines from `entry_function` already have the padding removed, so we can't dedent further.

reccmp/tools/project.py
```diff
@@ -0,0 +1,457 @@
...
+def get_default_main_cpp(target_id: str, original_path: Path, hpp_path: Path) -> str:
+    target_type = executable_or_library(original_path)
+    match target_type:
+        case TargetType.EXECUTABLE:
+            entry_function = textwrap.dedent(
+                """\
+                #ifdef _WIN32
+                #include <windows.h>
+
+                // FUNCTION: {original_id} 0x10000020
```
disinvite, line 314: Should these both be f-strings with `{target_id.upper()}` to match `get_default_main_hpp()` ?

reccmp/tools/project.py
```diff
@@ -0,0 +1,457 @@
...
+
+    create_parser = subparsers.add_parser("create")
+    create_parser.set_defaults(action="CREATE")
+    create_parser.add_argument(
+        "--originals",
+        type=Path,
+        nargs="+",
+        metavar="ORIGINAL",
+        dest="create_originals",
+        required=True,
```
jonschz, line 382: According to the [documentation](https://docs.python.org/3/library/argparse.html#required), it is a best practice to not have required arguments starting with `-` or `--`. How about a syntax like ` reccmp-project create LEGO1.DLL` without `--originals`?

reccmp/project/detect.py
```diff
@@ -0,0 +1,356 @@
...
+        logger.warning("User config %s is missing target ids", RECCMP_USER_CONFIG)
+    if user_keys - project_keys:
+        logger.warning(
+            "User config %s contains too many target ids", RECCMP_USER_CONFIG
+        )
+    if project_keys - build_keys:
+        logger.warning("Build config %s is missing target ids", RECCMP_USER_CONFIG)
+    if build_keys - project_keys:
+        logger.warning(
+            "Build config %s contains too many target ids", RECCMP_USER_CONFIG
```
jonschz, line 70: Bug: Output `RECCMP_BUILD_CONFIG` instead of `RECCMP_USER_CONFIG` (two instances)

jonschz (approved):
Looks great overall! Here are some remarks on what I found:

- README:
  - [ ] It'd be great if you could add the descriptions you made in this PR to this repo's README, (either in the main README.md or in another referenced file). I'm also happy with doing that in a separate PR if you feel like the interface is still too unstable.
  - Make sure no README gets lost (for example: I can't find the info on recompiling BETA10 here or on your branch). I think we'll double check that anyway once we migrate `isle`, just wanted to raise the awareness.

- I encountered some issues when I tried to set up BETA10 (see e.g. [here](https://github.com/isledecomp/isle/blob/master/tools/README.md#re-compiling-a-beta-build-wip)):
  - I added the following to `reccmp-project.yml`:
    ```yaml
      BETA10:
        filename: BETA10.DLL
        source-root: LEGO1
    ```
  - Neither auto-generated `reccmp-build.yml` (in `build` or `build_debug`) contains an entry for BETA10. The manual fix is to add
    ```yaml
      BETA10:
        path: '[...]/build_debug/LEGO1.DLL'
        pdb: '[...]/build_debug/LEGO1.DLL'
    ```
    to `build/reccmp-build.yml`. `build_debug/reccmp-build.yml` is obsolete and can be removed. I am not sure if this workflow can be improved significantly, partly because the build artifact for BETA10 is also called `LEGO1.DLL`.
  - [ ] This is an actual bug: With BETA10 set up, I get hundreds of false positive `duplicate_offset` warnings in `isledecomp`, likely because of the shared codebase between BETA10 and LEGO1.

- Others:
  - I assume it's by design that you can only run scripts like `reccmp.tools.reccmp` from the `build` folder since it doesn't find the `reccmp-build.yml` otherwise? I don't see an easy way around this right now, just wanted to make sure that it's intentional.
  - [x] The Ghidra import does not work correctly. I'll open a ticket here to find a solution once this PR is merged. IMHO this should be fixed before we migrate `isle` -> #2
  - General question: What is your personal intended workflow if you want to work on the LEGO1 decompile and `reccmp` at the same time? Install them next to each other and do a `pip install -e`?

disinvite:
>   * [ ]  This is an actual bug: With BETA10 set up, I get hundreds of false positive `duplicate_offset` warnings in `isledecomp`, likely because of the shared codebase between BETA10 and LEGO1.

On this point: should `decomplint` read (or require) the reccmp project target from the command line? We currently use the `--module` argument to decide which annotations to focus on, but we could drop that and use the target name instead.

Another option is to use the base filename of the original binary. This is what the main `reccmp` script does.

jonschz:
> On this point: should `decomplint` read (or require) the reccmp project target from the command line?

Given the extremely low runtime of `decomplint` compared to all other scripts, I think it's fine for it to run on all targets. I'd try to minimise the effort at the moment since there is a lot to be done. If we ever want to change the behaviour of `decomplint`, I see some merit in having a uniform interface for all our scripts.

disinvite:
> > On this point: should `decomplint` read (or require) the reccmp project target from the command line?
> 
> Given the extremely low runtime of `decomplint` compared to all other scripts, I think it's fine for it to run on all targets. I'd try to minimise the effort at the moment since there is a lot to be done. If we ever want to change the behaviour of `decomplint`, I see some merit in having a uniform interface for all our scripts.

I agree, but we should solve the `duplicate offset` errors. The cause is here:
```python
args.paths = list(target.source_root for target in project.targets.values())
```
If no paths are specified at the command line, we combine all the source roots from the reccmp-project file. If your yaml looks like this then the files in the LEGO1 directory will appear twice:
```yaml
targets:
  beta10:
    filename: BETA10.DLL
    source-root: LEGO1
  lego1:
    filename: LEGO1.DLL
    source-root: LEGO1
```
If we isolate based on the target name (to match the other scripts) then this won't happen.

reccmp/tools/roadmap.py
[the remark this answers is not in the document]
madebr: The executable bits are now set.
```sh
chmod a+x reccmp/tools/roadmap.py
```

reccmp/tools/stackcmp.py
```diff
@@ -277,66 +284,43 @@ def virtual_address(value) -> int:
...
-
-    if not os.path.isfile(args.pdb):
-        parser.error(f"Symbols PDB {args.pdb} does not exist")
-
-    if not os.path.isdir(args.decomp_dir):
-        parser.error(f"Source directory {args.decomp_dir} does not exist")
+    argparse_parse_logging(args=args)
+    if args.loglevel != logging.DEBUG:
+        # Mute logger events from compare engine
+        logging.getLogger("isledecomp.compare.core").setLevel(logging.CRITICAL)
```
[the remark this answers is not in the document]
madebr: The `--verbose` option has different meaning in `stackcmp.py` and `reccmp.py`.
I think we should find out:why we have to disable the logging event from the isledecomp module in the first place.
Perhaps it should return a list of errors, throw an exception, or log with a lower priority (debug?).
If a module is too spammy by default, then logging becomes less useful.

reccmp/project/detect.py
madebr: This is fixed.

reccmp/tools/project.py
madebr: I made these positional.

reccmp/tools/project.py
madebr: The code does not use `original_id` anymore, but uses `target_id` instead.
(I also moved the project modification/generation out of the tool into a proper module, adding unit tests)

reccmp/tools/project.py
madebr: I did something like that, with an extra newline for style :)

madebr:
> If no paths are specified at the command line, we combine all the source roots from the reccmp-project file. If your yaml looks like this then the files in the LEGO1 directory will appear twice:

I fixed this by going through a python set:
```python
list(set(target.source_root) for target in project.targets))
```

reccmp/tools/stackcmp.py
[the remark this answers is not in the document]
jonschz: * I'll create a ticket for the future so this doesn't block the merge
* I remember that I added some additional mutes to one of them because it annoyed me while working on BETA10

jonschz:
> I fixed this by going through a python set:

I can confirm that the issue is gone, thank you!

PR #4, opened 2024-10-22
Handle duplicate thunks

disinvite:
Merging my PR isledecomp/isle#1113 over here. Original notes:

>Debug builds put their thunk functions right at the start of `.text`. To match these up, we just scan the list in orig and recomp and look for cases where each points to the same real function that has already been matched.
>
>I had assumed each thunk would only appear once, but there are duplicates in BETA10. (I would guess that only one ends up being used in most/all cases.)
>
>This changes `match_thunks()` so that we detect any duplicates and match in order of the thunk's virtual address. This is probably a naive way to do it, but it works fine for the code in its current state. (Open question: does thunk order depend on module order?)

### DIFF
--- a/reccmp/isledecomp/compare/core.py
+++ b/reccmp/isledecomp/compare/core.py
@@ -428,11 +428,6 @@ class Compare:
         points at a function we have already matched, we can find the matching
         thunk in recomp because it points to the same place."""
 
-        # Turn this one inside out for easy lookup
-        recomp_thunks = {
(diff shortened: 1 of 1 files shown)


jonschz:
Since this has been reviewed on the main branch, is there anything to be done here? I haven't looked into the effects of this change in detail yet, please tell me if you want me to do so.

jonschz (approved):

foxtacles (approved):

PR #9, opened 2024-10-24
Don't make built target available if it is missing original path, recompiled path, or recompiled pdb path

madebr:
addresses https://github.com/isledecomp/isle/pull/1116#issuecomment-2436332429

@jonschz 

The code will/should now ignore targets that are incomplete.


jonschz:
Looks good, thanks for the quick fix!

Issue #7, opened 2024-10-24
Project file hierarchy


madebr:
I created this issue to discuss project file hierarchies.

#1 requires 3 project files, of which only one should be committed to a project's source repo:
- `reccmp-project.yml` is supposed to be pushed to a git repo, and should not contain absolute paths
- `reccmp-user.yml` should not be part of a git repo, and is supposed to be modified by the user (although the tools can initialize it with defaults). This file can contain absolute paths.
- `reccmp-build.yml` should also not be commited to the repo, and is supposed to be modified by the build systems, and can contain absolute paths

There are several reasons for this design decision:
- split public (=`reccmp-project.yml`) information and private (=`reccmp-user.yml` & `reccmp-build.yml`) information.
  It should be up to a developer to choose where and how binaries are stored and/or built
- split information that usually remain stable in a project (`reccmp-user.yml`), and information that depends on a build (`reccmp-build.yml`). Another reason for splitting `-user.yml` and `-build.yml` is that it allows a CMake project to generate `reccmp-build.yml` files without a python/reccmp dependency and also without worrying to accidentally destroy information.

jonschz:
Finally getting around to responding here.

First of all, the split into a `reccmp-project.yml` that is checked into git, and a `reccmp-user.yml` that is not checked into git sounds very reasonable to me, no need to discuss that point from my side.

Regarding the `reccmp-build.yml` I am a bit torn. Here are some thoughts:
* I don't see a lot of good arguments to *remove* it now that we already have it. Biggest counterargument is that it can make setting up the local repo for a new user a tiny bit harder, but that's not a big issue in my opinion.
* For Lego Island I consider this extra file to be slightly overkill (its information could go into the user file, imho). There might be other projects where it makes more sense to have a separate file, though. I haven't been involved with other decompilation projects so I can't comment on that.
* Personally, I don't get any benefit from having two files since I need to set up a non-changing, custom `reccmp-build.yml` for my personal Lego Island setup. There reasons for that are as follows:
  * The Ghidra import script cannot know where the recompiled files are, so the `reccmp-build.yml` must be located next to the `reccmp-project.yml`, which invalidates the benefit of being able to generate the file during the build process within the build directory.
[...]

Issue #8, opened 2024-10-24
Target other toolchains/platforms


madebr:
#1 added very basic project files that only supports a Windows MSVC workflow, using PDB files.
It would be useful to extend/rework this such that we can target other platforms and/or toolchains.

I propose a format in which we can add various data sources for targets.
It would still consist of 3 files (`reccmp-project.yml`, `reccmp-user.yml` and `reccmp-build.yml`).
- `reccmp-project.yml` is supposed to be pushed to a git repo, and should not contain absolute paths
- `reccmp-user.yml` should still not be part of a git repo, and is supposed to be modified by the user (although the tools can initialize it with defaults). This file can contain absolute paths.
- `reccmp-build.yml` should also not be commited to the repo, and is supposed to be modified by the build systems.

I created https://github.com/isledecomp/reccmp/issues/7 to discuss file hierarchies.
The proposal in this issue can be developed independent of #7.

The search order is `reccmp-build.yml` -> `reccmp-user.yml` -> `reccmp-project.yml`
The priority order is cli-arguments > `reccmp-build.yml` > `reccmp-user.yml` > `reccmp-project.yml`

Different from #1, is that in the new proposal these work together, are fully equivalent and stack on top of each other.
Values of  reccmp-build.yml` can override values in `reccmp-user.yml`, which in turn can override values of `reccmp-project.yml`.
The basic structure is as follows:
```yaml
targets: # REQUIRED, value = map of targets
  <id-target1>:  # Only required when the current yml overrides properties
    <datasource1>: # Should be present for every data source property this yml overrides
      type: win32-dll|ghidra|ida|...  # REQUIRED: must be present (should not be overriden, but let's leave this unspecified)
      <various properties that depend on the type>
```reccmp-datacmp --type dll -f path/to/original.dll --source path/to/recompiled/source --pdb path/to/recompiled.pdb --type dll -f path/to/recompiled.dll


Example project file structure:
```yml
# reccmp-project.yml
targets:
  LEGO1:
    original:  # "original" can be whatever. It's a sub-id
      type: win32-dll  # can also be win32-exe (and also elf32, dylib32, ...)
                       # Perhaps also include arch here or somewhere?
      filename: LEGO1.DLL
      hash:
        sha256: 14645225bbe81212e9bc1919cd8a692b81b8622abb6561280d99b0fc4151ce17
    recompiled:
      type: win32-dll
      source-root: LEGO1
    ghidra:
      type: ghidra
[...]

PR #12, opened 2024-10-27
Fix Ghidra Import

### LINKED ISSUE #2: Fix the Ghidra import
- [ ] Wait for #1 to be merged
- [ ] Make the Ghidra import work again - currently, it fails to find the correct `.venv`
- [ ] Think about a good setup. One idea:
  - Add `isle/.venv/Lib/ghidra_import` to the list of script directories in Ghidra after `pip install`
  - Make the script search for both a `reccmp-project.yml` and a `.venv` subfolder on its path up

jonschz:
This change restores the ability to import a built project into Ghidra.

**Known issues and limitations**
- Currently, the `reccmp-build.yml` must be at an auto-detectable location from the Ghidra script, i.e. it should be located next to `reccmp-project.yml`. This aspect is relevant to the discussion at #7 .
- There is still one hard-coded reference to Lego Island directly, and I am not sure what the best solution is. I'd also be fine with merging this PR and opening a follow-up issue for this.

It'd be great if you could try to set it up based on the reccmp README and see if/how the explanations could be improved.

Closes #2


setup.cfg
```diff
@@ -17,7 +17,7 @@ install_requires = file:requirements.txt
 
 [options.packages.find]
 exclude =
-    reccmp.tests.*
+    tests*
```
jonschz, line 20: Drive-by fix: The exclusion of tests in `pip install` did not work correctly

reccmp/ghidra_scripts/lego_util/ghidra_helper.py
```diff
@@ -109,7 +109,9 @@ def sanitize_name(name: str) -> str:
     # Importing function names like `FUN_10001234` into BETA10 can be confusing
     # because Ghidra's auto-generated functions look exactly the same.
     # Therefore, such function names are replaced by `LEGO_10001234` in the BETA10 import.
-    if GLOBALS.module == SupportedModules.BETA10:
+
+    # FIXME: The identification here is a crutch - we need a more reusable solution for this scenario
```
jonschz, line 113: This is the one place that is still isle-specific. Please tell me if you want me to find a solution before merging.

One idea I had was to add a `ghidra` section to the targets of a project configuration like so:
```yaml
  BETA10:
    filename: BETA10.DLL
    [...]
    ghidra:
      function_prefix: "LEGO1"
```

jonschz (commented):
I also chose to relocate the `ghidra_scripts` directory into the `reccmp` project because I didn't see much of a point in installing two top-level packages into the virtualenv. If you have strong opinions one way or the other, please let me know.

reccmp/project/detect.py
```diff
@@ -118,9 +118,9 @@ def from_directory(cls, directory: Path) -> typing.Optional["RecCmpProject"]:
 class RecCmpBuiltProject:
     def __init__(
         self,
-        project_config_path: typing.Optional[Path],
-        user_config: typing.Optional[Path],
-        build_config: typing.Optional[Path],
+        project_config_path: Path,
```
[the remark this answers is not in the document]
madebr: This is an oversight on my part.
I started with making these members optional, but then you'd need to check for validity every time you use them.
An object of which every member can be None is of nu use.

reccmp/ghidra_scripts/lego_util/ghidra_helper.py
madebr: Feel free to add whatever attributes you think are needed.
It's also be useful for a "ghidra" type in https://github.com/isledecomp/reccmp/issues/8

reccmp/isledecomp/compare/core.py
```diff
@@ -67,12 +68,16 @@ def lookup(addr: int, size: int) -> Optional[bytes]:
 class Compare:
     # pylint: disable=too-many-instance-attributes
     def __init__(
-        self, orig_bin: IsleBin, recomp_bin: IsleBin, pdb_file: str, code_dir: str
+        self,
+        orig_bin: IsleBin,
+        recomp_bin: IsleBin,
+        pdb_file: Path | str,
```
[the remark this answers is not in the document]
madebr: Wouldn't making them `Path` throughout the code make more sense? It's a form of intent and self-documenting code.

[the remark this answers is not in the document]
disinvite: `argparse` must support creating a `Path` object from the command line, right? If so, then it should be `Path` all the way down. But we could leave `Path | str` for now since this is not the only lingering type issue.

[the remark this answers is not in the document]
madebr: Yes, simply add `type=Path`
Keeping `Path | str` is fine with me :)

reccmp/ghidra_scripts/README.md
```diff
@@ -4,18 +4,28 @@ The scripts in this directory provide additional functionality in Ghidra, e.g. i
...
 Since these scripts and its dependencies are written in Python 3, [Ghidrathon](https://github.com/mandiant/Ghidrathon) must be installed first. Follow the instructions and install a recent build (these scripts were tested with Python 3.12 and Ghidrathon v4.0.0).
 
 ### Script Directory
+This step differs slightly depending on your setup.
+
 - In Ghidra, _Open Window -> Script Manager_.
 - Click the _Manage Script Directories_ button on the top right.
-- Click the _Add_ (Plus icon) button and select this file's parent directory.
+- Click the _Add_ (Plus icon) button.
+    - Case 1: Select `Lib/site-packages/reccmp/ghidra_scripts` inside your _project's_ virtual environment.
```
disinvite, line 24: Maybe a note here that you have to enable "Show '.' files" under "File Chooser Options" if your environment is under ".venv"

disinvite:
This one tripped me up: I had a hand-edited reccmp-build.yml under the isle dir from before we cut over to the reccmp module. If the paths to the recomp and pdb are relative (i.e. `build/LEGO1.PDB`) then the import script can't locate them.

reccmp/ghidra_scripts/lego_util/ghidra_helper.py
disinvite: What if we don't overwrite the Ghidra function name if the one from `reccmp` has the `FUN_` prefix? (Or if it matches a `temp_function_prefix` key from the yml?) We can still populate all the other information.

I don't know how common it is for a decomp to have two different versions of the same binary, but this might be an unusual situation.

disinvite:
I'll keep looking at this tomorrow, but my imports are missing functions for both LEGO1 and ISLE. LEGO1 misses the very first function `Score::Score` and ISLE doesn't annotate much of anything. I'm sure it must be something I have misconfigured, but I'm not sure what.
[logdump-isle.txt](https://github.com/user-attachments/files/17536182/logdump-isle.txt)

reccmp/ghidra_scripts/lego_util/ghidra_helper.py
jonschz: When I was working on BETA10 I found it hard to tell if a function was not matched and had Ghidra's default name, or if it was matched and was named `FUN_...` by the Ghidra script. There are also some address collisions between LEGO1  and BETA10 functions.

The "clean" solution would be to use a different pattern in the source code for unnamed functions, but that would be a major change to the existing code. It would have one additional minor benefit: It'll be easier to tell apart non-stubbed functions from stubbed / implemented ones in LEGO1.

jonschz:
@disinvite could you also provide a LEGO1 log? I haven't tried ISLE.EXE yet since I didn't realise that it should be possible now.

disinvite:
Here's the log from LEGO1 after deleting items from those two header files. (I have one from last night too if you want it.) I didn't know which things were known issues; maybe `Score::Score` was never there and I didn't notice.
[legolog.txt](https://github.com/user-attachments/files/17550421/legolog.txt)

But certainly the script is working again now in the new module, and since that was the goal of this PR then I think we are good.

jonschz:
Thanks! This is a good reminder to document the known issues in the Ghidra import, even if there is no need to fix them any time soon. I will still look into the relative import issue you discovered, and there is one more pre-existing bug I'd like to understand before merging.

reccmp/ghidra_scripts/lego_util/ghidra_helper.py
jonschz: will be addressed in a follow-up, see #16

jonschz:
Ticket for known issues: #17 
Ticket for remaining isle references: #16 

>  If the paths to the recomp and pdb are relative (i.e. build/LEGO1.PDB) then the import script can't locate them.
Should be fixed now for the `reccmp-build.yml`. I haven't looked at `reccmp-user.yml` yet.

jonschz:
@disinvite could you take a final look at the most recent changes, and verify that the issues you found are fixed? Thanks!

disinvite:
Looks good to me!

PR #15, opened 2024-10-29
Match array elements all at once

disinvite:
This slightly changes the code that matches each array element (and struct member) from array variables. Instead of doing this as each variable is matched, do it all at once. This eliminates the performance overhead of searching the list of cvdump analysis nodes by building the dict for it up front. Now all the time is spent on DB updates, as expected.

No diff on BETA10 or LEGO1.

In `_add_match_in_array` we had set the `compare_type` option to either `SymbolType.DATA` or `SymbolType.POINTER`. It is immediately unset (in the database) by the following call to `set_pair` so I just removed it. The type needs to be null or we will do a redundant comparison in `datacmp`. I'm still working on my long-promised overhaul of the DB, and hidden side effects of the API (like that) are one of the things it will address.

### DIFF
--- a/reccmp/isledecomp/compare/core.py
+++ b/reccmp/isledecomp/compare/core.py
@@ -8,7 +8,6 @@ from typing import Any, Callable, Iterable, List, Optional
 from reccmp.isledecomp.bin import Bin as IsleBin, InvalidVirtualAddressError
 from reccmp.isledecomp.cvdump.demangler import demangle_string_const
 from reccmp.isledecomp.cvdump import Cvdump, CvdumpAnalysis
-from reccmp.isledecomp.cvdump.types import scalar_type_pointer
 from reccmp.isledecomp.parser import DecompCodebase
 from reccmp.isledecomp.dir import walk_source_dir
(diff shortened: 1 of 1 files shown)


reccmp/isledecomp/compare/core.py
```diff
@@ -248,68 +247,68 @@ def orig_bin_checker(addr: int) -> bool:
...
         Note that there is no recursion, so an array of arrays would not be handled entirely.
         This step is necessary e.g. for `0x100f0a20` (LegoRacers.cpp).
         """
 
+        # Helper function
         def _add_match_in_array(
             name: str, type_id: str, orig_addr: int, recomp_addr: int
         ):
+            # pylint: disable=unused-argument
+            # TODO: Previously used scalar_type_pointer(type_id) to set whether this is a pointer
```
jonschz, line 263: I can't find an argument why the `compare_type` would be needed. We can always re-introduce it if we need it in the future.

jonschz (approved):
Looks good! I have played around with it for a bit and I haven't found any change in behaviour.

PR #18, opened 2024-11-03
Improve DB performance and code clarity

disinvite:
Some tweaks to the database before rolling out a new feature (dynamic metadata) I've been working on. The goal here is to optimize the current design and then evaluate the performance shift of the upcoming changes.

- Bulk inserts of data. The biggest users of `set_recomp_symbol` are `load_cvdump` and `match_array_elements`, and both now run the `INSERT` statement using `executemany` instead of calling `execute` repeatedly. It seems cleaner to do it this way when possible, and I think we are cutting at least some overhead by directly reusing the prepared SQL statement instead of relying on the statement cache.

- Unique indices on `orig_addr` and `recomp_addr`. I remember testing this way back and found that performance dropped, so my conclusion was that it was just faster to test each row before inserting. The real problem is that the main search query checking `WHERE orig_addr IS NULL` now prefers the unique index on `orig_addr`, which is a full table scan if all are NULL. It should use the better index on the name field. If you nudge the optimizer to avoid `orig_addr` then it runs fine.

- Some methods return `cur.fetchall()` when returning a generator with `yield from` works better. I had the wrong type annotations in some of these places too.

- SymbolType enum is now an `enum.IntEnum`. This eliminates the call to the SymbolType constructor when fetching out of the database, along with some other code to set up query parameters. We can still do equality checks and test types using `mypy`. When converting the compare_type to its string representation in `match_name`, it is faster to just use a lookup. Maybe this means we should really have `enum.StrEnum` instead with the names of these types.

- `match_vtable` and `match_static_variable` were a little confusing to read. Both look better now.

### DIFF
--- a/reccmp/isledecomp/compare/core.py
+++ b/reccmp/isledecomp/compare/core.py
@@ -109,6 +109,10 @@ class Compare:
         )
         self.cvdump_analysis = CvdumpAnalysis(self.cv)
 
+        # Build the list of entries to insert to the DB.
+        # In the rare case we have duplicate symbols for an address, ignore them.
+        dataset = {}
+
         for sym in self.cvdump_analysis.nodes:
             # Skip nodes where we have almost no information.
             # These probably came from SECTION CONTRIBUTIONS.
@@ -124,6 +128,9 @@ class Compare:
 
             addr = self.recomp_bin.get_abs_addr(sym.section, sym.offset)
             sym.addr = addr
+
+            if addr in dataset:
+                continue
 
             # If this symbol is the final one in its section, we were not able to
             # estimate its size because we didn't have the total size of that section.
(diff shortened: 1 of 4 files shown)


foxtacles (approved):
Looks good to me!

reccmp/isledecomp/compare/core.py
```diff
@@ -260,22 +275,16 @@ def _match_array_elements(self):
...
         """
 
+        dataset = {}
+
         # Helper function
         def _add_match_in_array(
             name: str, type_id: str, orig_addr: int, recomp_addr: int
         ):
             # pylint: disable=unused-argument
             # TODO: Previously used scalar_type_pointer(type_id) to set whether this is a pointer
```
jonschz, line 285: I think this TODO can be removed

jonschz (approved):
While I (admittedly) didn't look over all the details of the change, what I looked at looks very good and I didn't find any regressions while experimenting with it. Thanks!

reccmp/isledecomp/compare/core.py
disinvite: I think I'll keep it until the next round. We may want to add type info to the database for these entries

Issue #13, opened 2024-10-27
Feature wishlist for Ghidra import


jonschz:
This is a collection of ideas how the Ghidra import could be improved. Feel free to add your thoughts here.

- [ ] Refactor isle-specific code (see https://github.com/isledecomp/reccmp/pull/12#discussion_r1817995984)
- [ ] Deny-list for types and functions that are not imported correctly and have been fixed by hand
  - Example: classes in BETA10 that differ from LEGO1
  - Where should this be configured? Dedicated file? New Ghidra section in the project configuration?
- [ ] Import vtables
  - Import vtables as `Structure` and add function pointer types for all functions.
  - A quick manual PoC looked quite promising
  - This should greatly increase the readability of virtual function calls

PR #19, opened 2024-11-16
Ghidra ignore list, config file pydantic refactor

jonschz:
Major changes:

### Pydantic
We now use [pydantic](https://docs.pydantic.dev/latest/) to parse and verify config files, as briefly discussed on Matrix
Note that some missing fields are now hard errors, which previously would (partially) lead to targets being skipped. This can be changed in case you encounter signficant issues, but an *incomplete* target should, in my opinion, raise an error while a missing target can be skipped.

### Ghidra import deny list

Implements one feature of #13 
We can now skip types and functions in the Ghidra import like so:
```yaml
targets:
  BETA10:
    ghidra:
      ignore-types:
        # these classes have been changed by hand to account for changes between LEGO1 and BETA10
        - LegoCarBuild
        - LegoWorld
        - LegoCarBuildAnimPresenter
      ignore-functions: []
        # for potential future use, currently empty
```
I will PR this change on `isle` when this one is merged.

### Linting and formatting for unit tests
The tests are now also covered by `pylint` and `black`. I'm open for discussion in case you don't like it.

### DIFF
--- a/.github/workflows/format.yml
+++ b/.github/workflows/format.yml
@@ -19,4 +19,5 @@ jobs:
       shell: bash
       run: |
         pylint reccmp
-        black --check reccmp
+        pylint tests
+        black --check .
--- a/reccmp/ghidra_scripts/import_functions_and_types_from_pdb.py
+++ b/reccmp/ghidra_scripts/import_functions_and_types_from_pdb.py
@@ -30,7 +30,6 @@ from pathlib import Path
 import traceback
 from typing import TYPE_CHECKING, Optional
 
-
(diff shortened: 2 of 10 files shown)


disinvite:
Looks good! We might have had linting/formatting on "tests/" and it just got lost in the conversion to this repo. Good to have that back. Agree on failing for incomplete targets but skipping unneeded ones.

@madebr any thoughts on this?

madebr (approved):
Nice simplification offered by pydantic :)

PR #20, opened 2024-11-20
Unified read function for decomp parser/linter

disinvite:
As discussed on matrix chat, I've been working on moving us from the current line-based C++ parser to a token-based parser. This will allow other teams to use whatever house style they want and get their annotations loaded without any surprises. It's a big change, so I'm going to add things in stages so it's easier to review.

First up is having just one `read()` method for both the parser and the decomp linter. The token-based parser needs to get the entire file's contents as a string, so file reads and unit tests will now use the same entrypoint.

The unit tests all used a list of strings to represent the code lines. These are now multiline strings, which I suppose is the best amongst the lot of not-aesthetically-pleasing options. I only used `dedent` where needed for the tests to pass -- this is typically due to not detecting the start or end of the function because we require the curly brace to be in a certain spot. The token-based parser won't have this problem so we can take those calls out later.

As expected, we are reading all the annotations exactly the same on the main isle repo before and after this change.

### DIFF
--- a/reccmp/isledecomp/parser/codebase.py
+++ b/reccmp/isledecomp/parser/codebase.py
@@ -19,7 +19,7 @@ class DecompCodebase:
         for filename in filenames:
             parser.reset()
             with open(filename, "r", encoding="utf-8") as f:
-                parser.read_lines(f)
+                parser.read(f.read())
 
             for sym in parser.iter_symbols(module):
                 sym.filename = filename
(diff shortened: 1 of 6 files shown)


jonschz (approved):
Looks good to me! The only possible issue I could think of was issues related to line endings, but a few quick tests of mine showed no issues there.

disinvite:
Thanks! I think we are okay if we always convert to `\n`, which already happens at the call to `open()` if the `newline` kwarg is `None`. Looks like you always get LF with multiline strings even if the _file_ uses CRLF.

PR #24, opened 2024-11-30
Add annotations to error messages

jonschz:
This small change makes it easier to find where an error message originates from.


jonschz:
Added feature: The same string can now be annotated multiple times without raising errors, as discussed on https://github.com/isledecomp/isle/pull/1177

disinvite (approved):

PR #25, opened 2024-12-01
Dynamic metadata for program database

disinvite:
This refactors the program database to store any kind of metadata, no matter how specific. For example:
- Bool to identify thunk, vtordisp, import functions
- Unique name for functions that look alike. (e.g. free functions like `operator!=`, or const vs non-const operators)
- Separate size for orig/recomp
- Back-reference for items inside a function like unwinds or jump tables

To implement this, we are using a JSON string to store key-value pairs. Any format would do, but [SQLite has a nice API](https://www.sqlite.org/json1.html) for working with JSON. Other options I looked at:
- A bunch of python dicts. Probably the fastest but getting there requires a lot of code to re-implement indices. You also lose atomicity and concurrency that we get for free with SQLite.
- `ALTER TABLE` statements to dynamically add columns. The problem here is that you need to maintain state at the python level to avoid sqlite exceptions, and _all_ your SQL statements need to be dynamic to account for which columns are used. (I may still explore this option, though.)

Not that JSON is without its drawbacks:
- We're doing our serialization/deserialization with the python `json` library, which adds non-trivial but not earth-shattering overhead. Comedy: this is faster than `json.loads` (but not enough to use):
    ```python
    dict(sql.execute("SELECT key, value FROM json_each(?)", (json_string,))
    ```
    There are drop-in replacement JSON modules that are intended to be faster, so maybe we can use one of those.
- On the topic of serialization: a faster way around it is to use a "stage" table to dump all your key/value pairs and then use the aggregate `json_group_object()` to collect them all in a JSON string and set into the listing. The problem here is that SQLite has no boolean type, so your python `True` is converted to a 1 before going into the JSON string. `json.dumps` maintains the bool so we don't need a conversion method.
- LEGO1 and ISLE use an odd string "\x00\x00" in `__crt` functions. The PDB points at its exact location in the recomp. The problem is that JSON does not distinguish between that and the empty string (and who could blame it). I worked around this for now by just dropping this specific string. (Accuracy went up by a tick.) We probably should compare const data (string and floats) by looking at the bytes. This will probably cover us with unicode strings too.

Other stuff:
- `set_orig_symbol` and `set_recomp_symbol` now use key args exclusively. This always should have been the case, but we need it now for the dynamic properties.
- `_match_array_elements` now updates existing entries, meaning that the asm will show `g_roiColorAliases[0].m_name` instead of just `g_roiColorAliases` for the first entry.
- The ORM object can access all the dynamic fields via `get()`. We could extend `__getitem__` so you can use `obj.whatever` instead. I added all the previous fields as `@property` methods for now to not introduce any breaking changes.
- New `bulk_match` method that does what the name indicates. I'm going to look into using this more to get extra performance.

The `match_options` table will get folded into the JSON string in a later update. (The ghidra scripts and main compare code use this and I didn't want to change those here.)

### DIFF
--- a/reccmp/isledecomp/compare/core.py
+++ b/reccmp/isledecomp/compare/core.py
@@ -168,10 +168,13 @@ class Compare:
                     decoded_string = raw.decode("latin1")
                     rstrip_string = decoded_string.rstrip("\x00")
 
-                    if decoded_string != "" and rstrip_string != "":
-                        sym.friendly_name = rstrip_string
-                    else:
-                        sym.friendly_name = decoded_string
+                    # TODO: Hack to exclude a string that contains \x00 bytes
+                    # The proper solution is to escape the text for JSON or use
+                    # base64 encoding for comparing binary values.
+                    # Kicking the can down the road for now.
+                    if "\x00" in decoded_string and rstrip_string == "":
+                        continue
+                    sym.friendly_name = rstrip_string
 
                 except UnicodeDecodeError:
                     pass
@@ -184,9 +187,7 @@ class Compare:
             }
 
         # Convert dict of dicts (keyed by addr) to list of dicts (that contains the addr)
-        self._db.bulk_cvdump_insert(
-            ({"addr": key, **values} for key, values in dataset.items())
-        )
+        self._db.bulk_recomp_insert(dataset.items())
 
         for (section, offset), (
             filename,
@@ -324,8 +325,13 @@ class Compare:
                                 recomp_element_base_addr + member.offset,
                             )
 
-        self._db.bulk_array_insert(
-            ({"recomp": key, **values} for key, values in dataset.items())
+        # Upsert here to update the starting address of variables already in the db.
+        self._db.bulk_recomp_insert(
+            ((addr, {"name": values["name"]}) for addr, values in dataset.items()),
+            upsert=True,
+        )
+        self._db.bulk_match(
+            ((values["orig"], addr) for addr, values in dataset.items())
         )
 
     def _find_original_strings(self):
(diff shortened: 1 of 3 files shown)


reccmp/isledecomp/compare/db.py
```diff
@@ -377,26 +397,28 @@ def _find_potential_match(
...
         # by adding the plus sign (Reference: https://www.sqlite.org/optoverview.html#uplus)
         if match_decorate:
-            sql = """
-            SELECT recomp_addr
-            FROM `symbols`
-            WHERE +orig_addr IS NULL
-            AND decorated_name = ?
-            AND (compare_type IS NULL OR compare_type = ?)
-            LIMIT 1
+            # TODO: Change when/if decorated becomes a unique column
```
jonschz, line 400: Is this something you keep track of? I wouldn't know what to make of it when I read this line.

jonschz (approved):
Overall, the change looks very good to me. It will also likely allow some beneficial refactorings in the Ghidra import, where I currently start with a match but I need to fetch additional information from the DB.

I also agree with the architectural decisions you have made - a JSON string field sounds like the best available option to me. Looking forward to the subsequent PRs!

reccmp/isledecomp/compare/core.py
```diff
@@ -168,10 +168,13 @@ def _load_cvdump(self):
...
                     rstrip_string = decoded_string.rstrip("\x00")
 
-                    if decoded_string != "" and rstrip_string != "":
-                        sym.friendly_name = rstrip_string
-                    else:
-                        sym.friendly_name = decoded_string
+                    # TODO: Hack to exclude a string that contains \x00 bytes
+                    # The proper solution is to escape the text for JSON or use
+                    # base64 encoding for comparing binary values.
+                    # Kicking the can down the road for now.
```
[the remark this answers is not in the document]
disinvite: Writing this up

reccmp/isledecomp/compare/db.py
disinvite: That note is a leftover from earlier designs. We could use the symbol (i.e. linker name, or "decorated name" in MSVC land) as another unique ID alongside the two addresses. The idea in this function is that a potential match using a symbol name should skip all these checks on non-unique fields. (The DB should optimize the query plan to do that anyway.)

I took this out because multiple symbols can point to the same address. Right now we deal with that by throwing away the other symbol(s). This really only comes up in the C runtime library with string functions, but it's something to revisit for programs with handwritten assembly.

PR #26, opened 2024-12-01
Run reccmp on binaries with no .reloc or .idata section

### LINKED ISSUE #22: Cannot compare binaries without `.reloc` section
Comparing binaries that do not have a `.reloc` section fails.
Such binaries can be created by [linking with `/FIXED`](https://library.thedatadungeon.com/msdn-1998-06/vccore/HTML/_core_.2f.fixed.htm).

backtrace:
```
  File "/home/maarten/.local/bin/reccmp-reccmp", line 8, in <module>
    sys.exit(main())
             ^^^^^^
  File "/home/maarten/projects/reccmp/reccmp/tools/asmcmp.py", line 224, in main
    with Bin(target.original_path, find_str=True) as origfile, Bin(
         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/maarten/projects/reccmp/reccmp/isledecomp/bin.py", line 192, in __enter__
    self._populate_relocations()
  File "/home/maarten/projects/reccmp/reccmp/isledecomp/bin.py", line 269, in _populate_relocations
    reloc = self.get_section_by_name(".reloc").view
            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/maarten/projects/reccmp/reccmp/isledecomp/bin.py", line 492, in get_section_by_name
    raise SectionNotFoundError
reccmp.isledecomp.bin.SectionNotFoundError
```

Can `_populate_relocations` be skipped when `.reloc` is not present?

I think we need to use the data directories to select what sections to use.
It's e.g. possible to have multiple `.data` sections.

I dabbled a bit in https://github.com/madebr/reccmp/tree/use-data-directories, but I think I broke reccmp

madebr:
Use the data directory instead

Fixes #22


jonschz:
Without having looked at the code yet: Maybe we want to add an example file to the testing suite to verify the correct handling of these kinds of binaries?

disinvite:
Very nice. Can you move all the dataclasses/enums to their own file? It's a little crowded with them all here. Instead of the "don't change the order" comment for each class, you could have one blanket comment for the entire file. I don't expect these will change much after this except to add new structs.

Maybe we should have a `binfile` module? Then this could be `reccmp/isledecomp/binfile/pefile.py` or something.

reccmp/isledecomp/bin.py
```diff
@@ -25,38 +29,337 @@ class InvalidVirtualAddressError(IndexError):
...
+class ImageSectionHeader:
+    name: str  # Name
+    virtual_size: int  # VirtualSize
+    virtual_address: int  # VirtualAddress
+    size_of_raw_data: int  # SizeOfRawData
+    pointer_to_raw_data: int  # PointerToRawData
+    pointer_to_relocations: int  # PointerToRelocations
+    pointer_to_line_numbers: int  # NumberOfRelocations
+    number_of_relocations: int  # NumberOfLineNumbers
+    number_of_line_numbers: SectionFlags  # Characteristics
```
disinvite, line 341: Should `characteristics` have its own member? I don't know if `number_or_relocations` or `number_of_line_numbers` are used for anything, but if we read both then the struct format should be `<8s6I2HI`

disinvite: Probably don't need the comments after each member either

reccmp/isledecomp/bin.py
```diff
@@ -25,38 +29,337 @@ class InvalidVirtualAddressError(IndexError):
...
+    e_lfarlc: int
+    e_ovno: int
+    e_res: tuple[int, int, int, int]
+    e_oemid: int
+    e_oeminfo: int
+    e_res2: tuple[int, int, int, int, int, int, int, int, int, int]
+    e_lfanew: int
+
+    @classmethod
+    def from_memory(cls, view: memoryview, offset: 0) -> tuple["ImageDosHeader", int]:
```
disinvite, line 57: `offset: int = 0` here

reccmp/isledecomp/bin.py
```diff
@@ -25,38 +29,337 @@ class InvalidVirtualAddressError(IndexError):
...
+        struct_fmt = "<30HI"
+        items = struct.unpack_from(struct_fmt, view[:64], offset)
+        result = cls(
+            *items[:14],
+            tuple(items[14:18]),
```
disinvite, line 62: The range from `items` is already a tuple, so this should just be:
```python
result = cls(*items[:14], items[14:18], *items[18:20], items[20:30], items[30])
```

madebr:
> Very nice. Can you move all the dataclasses/enums to their own file? It's a little crowded with them all here. Instead of the "don't change the order" comment for each class, you could have one blanket comment for the entire file. I don't expect these will change much after this except to add new structs.
> 
> Maybe we should have a `binfile` module? Then this could be `reccmp/isledecomp/binfile/pefile.py` or something.

Well, I started moving `bin` files to a `pe` module and before I knew it reccmp is able to detect multiple executable file formats (detect only, not load)

madebr:
@disinvite 
You're right, the diff was too big.
Can you have another quick glance?
I removed the indentation diff by doing `if True:` (+ a pylint ignore comment)
The diff should be clearer now.


About `find_str`, I think that should be moved out of `PEImage` entirely to a higher abstraction where we don't care about the actual format.
As you suggested, I'll move it into `PEImage` for now.

reccmp/isledecomp/formats/pe.py
```diff
@@ -0,0 +1,1005 @@
...
+
+    @property
+    def is_debug(self) -> bool:
+        return (
+            self.optional_header.directories[PEDataDirectoryItemType.DEBUG.value].rva
+            != 0
+        )
+
+    @property
+    def pdb_filename(self) -> Optional[str]:
```
madebr, line 578: FYI, it's now possible to get the (Windows) PDB file path from a PE image.

reccmp/isledecomp/compare/core.py
```diff
@@ -6,7 +6,10 @@
 import uuid
 from dataclasses import dataclass
 from typing import Any, Callable, Iterable, List, Optional
-from reccmp.isledecomp.bin import Bin as IsleBin, InvalidVirtualAddressError
+
+# from reccmp.isledecomp.bin import Bin as IsleBin, InvalidVirtualAddressError
```
disinvite, line 10: Can remove this

reccmp/tools/asmcmp.py
```diff
@@ -221,9 +223,17 @@ def main():
 
     logging.basicConfig(level=args.loglevel, format="[%(levelname)s] %(message)s")
 
-    with Bin(target.original_path, find_str=True) as origfile, Bin(
-        target.recompiled_path
-    ) as recompfile:
+    origfile = detect_image(filepath=target.original_path)
```
disinvite, line 226: Since there's only one argument to `detect_image`, we can probably drop the filepath key here and in other scripts. (Or use it consistently if the intent is to future-proof)

reccmp/isledecomp/formats/elf.py
jonschz, line 1: Have you created this based on some documentation or other code? In any case, a link to the reference / source you used would be great. Thanks!

reccmp/tools/vtable.py
```diff
@@ -68,9 +68,16 @@ def main():
...
-        target.recompiled_path
-    ) as recomp_bin:
+    orig_bin = detect_image(args.original)
+    assert isinstance(orig_bin, PEImage)
+
+    recomp_bin = detect_image(args.recompiled)
+    assert isinstance(recomp_bin, PEImage)
+    engine = IsleCompare(orig_bin, recomp_bin, args.pdb, args.decomp_dir)
+
+    # FIXME: remove "if True"
```
jonschz, line 78: thanks, this greatly increases the comparability. Feel free to remove it after @disinvite has taken another look

jonschz (commented):
Looks great! Assuming there are no regressions on `isle`, I'd also be fine with merging it and adding tests later. This shouldn't be stuck in limbo forever.

reccmp/isledecomp/formats/detect.py
```diff
@@ -0,0 +1,33 @@
+from pathlib import Path
```
disinvite, line 1: We could maybe merge this file into `__init__.py` if it is the main entrypoint for binary images

reccmp/isledecomp/formats/mz.py
```diff
@@ -0,0 +1,74 @@
...
+    @classmethod
+    def from_memory(cls, data: bytes, offset: int) -> tuple["ImageDosHeader", int]:
+        struct_fmt = "<30HI"
+        struct_size = struct.calcsize(struct_fmt)
+        items = struct.unpack_from(struct_fmt, data, offset)
+        result = cls(
+            *items[:14],
+            items[14:18],
+            *items[18:20],
+            tuple[20:30],
```
disinvite, line 47: items[20:30]

reccmp/isledecomp/formats/pe.py
```diff
@@ -0,0 +1,1005 @@
...
+from pathlib import Path
+import struct
+from typing import Iterator, Optional
+
+from .exceptions import InvalidVirtualAddressError, SectionNotFoundError
+from .image import Image
+from .mz import ImageDosHeader
+
+
+logger = logging.getLogger(__name__)
```
disinvite, line 15: I think we can just drop the logger at this point

reccmp/isledecomp/formats/pe.py
```diff
@@ -0,0 +1,1005 @@
...
+
+    @classmethod
+    def from_memory(cls, data: bytes, offset: int) -> Optional["CodeViewHeaderNB10"]:
+        struct_fmt = "<4sIII"
+        if not cls.taste(data, offset):
+            raise ValueError
+        items = struct.unpack_from(struct_fmt, data, offset)
+        offset_pdb_filename = offset + struct.calcsize(struct_fmt)
+        pos_null = data.index(0, offset_pdb_filename)
+        if pos_null == -1:
```
disinvite, line 346: `bytes.index()` throws `ValueError` if not found. Should use `find()` instead or wrap in try/catch. Same for the RSDS header

reccmp/isledecomp/formats/image.py
```diff
@@ -0,0 +1,12 @@
+import dataclasses
+from pathlib import Path
+
+
+@dataclasses.dataclass
+class Image:
+    filepath: Path
+    view: memoryview = dataclasses.field(repr=False)
+    data: bytes = dataclasses.field(repr=False)
```
disinvite, line 9: Do we need `data` if we have the memoryview of the file?

reccmp/isledecomp/formats/mz.py
```diff
@@ -0,0 +1,74 @@
...
+from .image import Image
+
+
+class MZHeaderNotFoundError(ValueError):
+    """MZ magic string not found"""
+
+
+# pylint: disable=too-many-instance-attributes
+@dataclass(frozen=True)
+class ImageDosHeader:
```
disinvite, line 14: Not that we can't look it up (and not that this isn't a decades-old format) but it would be nice to have more descriptive member names here that are self-documenting. The ones in the PE module are a good example.

reccmp/isledecomp/formats/mz.py
```diff
@@ -0,0 +1,74 @@
...
+    # Order is significant!
+    e_magic: int
```
disinvite, line 16: Maybe `bytes` here to match PE? And change the struct string in `from_memory()` and `taste()` to match

reccmp/isledecomp/formats/mz.py
```diff
@@ -0,0 +1,74 @@
...
+        struct_size = struct.calcsize(struct_fmt)
+        items = struct.unpack_from(struct_fmt, data, offset)
+        result = cls(
+            *items[:14],
+            items[14:18],
+            *items[18:20],
+            tuple[20:30],
+            items[30],
+        )
+        if result.e_magic != cls.MAGIC:
```
disinvite, line 50: We should follow the pattern set by the other modules and call `taste()` right at the beginning instead of checking it here

reccmp/isledecomp/formats/mz.py
```diff
@@ -0,0 +1,74 @@
...
+    e_cs: int
+    e_lfarlc: int
+    e_ovno: int
+    e_res: tuple[int, int, int, int]
+    e_oemid: int
+    e_oeminfo: int
+    e_res2: tuple[int, int, int, int, int, int, int, int, int, int]
+    e_lfanew: int
+
+    MAGIC = 0x5A4D  # "MZ"
```
disinvite, line 36: In general I think we can just use the constant where it's needed instead of saving it here. If all image files had exactly one magic number/signature then it would make sense, but that's not the case

reccmp/isledecomp/formats/elf.py
madebr: Based on [`/usr/include/elf.h` from glibc](https://sourceware.org/git/?p=glibc.git;a=blob;f=elf/elf.h;h=33aea7f743b885c5d74736276e55ef21756293ee;hb=HEAD)

reccmp/isledecomp/formats/mz.py
madebr: It's a pythoninized version of the one in the C headers: `IMAGE_DOS_HEADER`
https://github.com/hughbe/windows-sdk-headers/blob/89f151d343587fcd5b854a4851fcb8f7187f3ac4/Include/10.0.19041.0/km/ntimage.h#L69

I'd prefer to keep it.

reccmp/isledecomp/formats/mz.py
madebr: Yeah, probably.
The header has `USHORT` though, but that shouldn't stop us.
https://github.com/hughbe/windows-sdk-headers/blob/89f151d343587fcd5b854a4851fcb8f7187f3ac4/Include/10.0.19041.0/km/ntimage.h#L70

reccmp/tools/asmcmp.py
madebr: I think it'd be better to change the API by splitting binary format detection and loading.
Right now, `detect_file` does more then detecting: it also loads the file.

reccmp/isledecomp/formats/image.py
madebr: I added it for extracting the PDB filename from a PE image: `memoryview` has no `index` or `find` method.

reccmp/isledecomp/formats/elf.py
jonschz: Thanks! I should have worded it better - could you put this one-liner into the docstring of the file?

reccmp/isledecomp/formats/elf.py
```diff
@@ -0,0 +1,301 @@
...
+    EM_AMDGPU = 224  # AMD GPU
+    EM_RISCV = 243  # RISC-V
+    EM_BPF = 247  # Linux BPF -- in-kernel virtual machine
+    EM_CSKY = 252  # C-SKY
+    EM_LOONGARCH = 258  # LoongArch
+
+
+@dataclass(frozen=True)
+class ElfFileHeader:
+    # Based on elf.h from glibc (https://sourceware.org/git/?p=glibc.git;a=blob;f=elf/elf.h)
```
madebr, line 222: Is this ok docstring?
Or must it be
```suggestion
    """
    Based on elf.h from glibc: https://sourceware.org/git/?p=glibc.git;a=blob;f=elf/elf.h
    """
```

jonschz: Oh, I completely missed that. I suppose the code starting from line 221 is based on `elf.h`, but the code above isn't?

In that case, the comment is good enough in my opinion, but a source for the code above wouldn't hurt. Thanks!

reccmp/isledecomp/compare/core.py
```diff
@@ -174,7 +181,7 @@ def _load_cvdump(self):
                     # Kicking the can down the road for now.
                     if "\x00" in decoded_string and rstrip_string == "":
                         continue
-                    sym.friendly_name = rstrip_string
+                    sym.friendly_name = decoded_string
```
disinvite, line 184: What problem did you run into here? I now see a diff on `BETA10` for strings without a code annotation. The recomp side has `\x00` on the string where orig does not. Should this be `rstrip_string` to remove those?

reccmp/isledecomp/formats/image.py
disinvite: Ah you're right. Maybe we could just send this as the param to `CodeViewHeaderNB10` and `CodeViewHeaderRSDS`:
```python
debug_raw = bytes(self.view[debug_entry.pointer_to_raw_data : debug_entry.pointer_to_raw_data + debug_entry.size_of_data])
```

Could even use `rstrip` to capture the string if the pdb filename is all that's left.

reccmp/isledecomp/compare/core.py
madebr: I saw an error here but 2bc2a068cfa33098699bf838217a4b02114a3eb0 made it go away. I'll revert this change.

reccmp/isledecomp/formats/elf.py
madebr: I've added a top comment to all file formats.

reccmp/isledecomp/formats/image.py
madebr: That's possible, but then you'd do a memory copy of potentially a lot of data (but most of the times not much).

disinvite:
The only outstanding issue I see is the one with the `PEImageSectionHeader` [members](https://learn.microsoft.com/en-us/windows/win32/debug/pe-format#section-table-section-headers). After that, I think we can merge.

madebr:
> The only outstanding issue I see is the one with the `PEImageSectionHeader` [members](https://learn.microsoft.com/en-us/windows/win32/debug/pe-format#section-table-section-headers). After that, I think we can merge.

Was this addressed by https://github.com/isledecomp/reccmp/pull/26/commits/c2cedad4c5a492d07d1de44ae1424d15297658b1?

disinvite:
> > The only outstanding issue I see is the one with the `PEImageSectionHeader` [members](https://learn.microsoft.com/en-us/windows/win32/debug/pe-format#section-table-section-headers). After that, I think we can merge.
> 
> Was this addressed by [c2cedad](https://github.com/isledecomp/reccmp/commit/c2cedad4c5a492d07d1de44ae1424d15297658b1)?

No, it's line 304 in pe.py:
```python
number_of_line_numbers: PESectionFlags
```
I think that should be:
```python
number_of_line_numbers: int
characteristics: PESectionFlags
```
with struct format: `<8s6I2HI`

madebr:
You're right. I fixed it in 14d0ee8

madebr:
I didn't catch this error, because it seems like `number_of_line_numbers` is 0 in all binaries I tested it on.

disinvite:
Nice! After @jonschz gives a final look I think we are all set here.

reccmp/tools/roadmap.py
```diff
@@ -394,9 +394,15 @@ def main() -> int:
...
-    ) as recomp_bin:
+    orig_bin = detect_image(target.original_path)
+    assert isinstance(orig_bin, PEImage)
+
+    recomp_bin = detect_image(target.recompiled_path)
+    assert isinstance(recomp_bin, PEImage)
+
+    # FIXME: remove "if True"
+    # pylint: disable=using-constant-test
+    if True:
```
jonschz, line 405: This FIXME should still be addressed (and please look for others)

reccmp/tools/stackcmp.py
```diff
@@ -315,13 +316,20 @@ def main():
...
+    if not isinstance(origfile, PEImage):
+        raise ValueError(f"{target.original_path} is not a PE executable")
+
+    recompfile = detect_image(filepath=target.recompiled_path)
+    if not isinstance(recompfile, PEImage):
+        raise ValueError(f"{target.recompiled_path} is not a PE executable")
+
+    # FIXME: remove "if True"
+    # pylint: disable=using-constant-test
+    if True:
```
jonschz, line 329: Address FIXME

reccmp/tools/vtable.py
```diff
@@ -68,9 +68,16 @@ def main():
...
+    # pylint: disable=using-constant-test
+    if True:
```
jonschz, line 80: Address FIXME

jonschz (commented):
Great work! Will merge after the FIXMEs are removed

PR #29, opened 2024-12-07
Alert to ambiguous name match

### LINKED ISSUE #28: Alert to ambiguous name match
Re: isledecomp/isle#1192

When matching a comment annotation by its name[^1], we pick whichever entry is available first based on insertion order. [^2] This allows us to match multiple entries that have the same name. For example: `_Construct`. The order of match _attempts_ depends on the order we read each file; this is what we get when using `os.walk` to iterate on the code directory.

A function with different arguments or a different return type (including const vs. non-const) will appear to have the same name in the output from cvdump. The current way to distinguish between these functions is to match using the symbol name instead, which includes all of this information while not being very human-readable.

We can mitigate this problem in a few different ways:

1. For functions with the same name, display an alternate "unique name" that includes the distinguishing characteristics. This will help asm sanitize show a diff where there is one.
2. If we detect that a name match has multiple possible results, alert the user to what was matched. For example, a message like:
    ```
    Match at 0x10049160 is ambiguous, we chose: ?erase@?$_Tree@PAULegoPathCtrlEdge@@PAU1@U_Kfn@?$set@PAULegoPathCtrlEdge@@ULegoPathCtrlEdgeCompare@@V?$allocator@PAULegoPathCtrlEdge@@@@@@ULegoPathCtrlEdgeCompare@@V?$allocator@PAULegoPathCtrlEdge@@@@@@QAE?AViterator@1@V21@@Z
    ```
3. Sort match annotations by the original address so that matching does not depend on file or directory names.

[^1]: Unless we determine that the name given is a symbol, which means (for C++ symbols in MSVC) that "?" is the first character.
[^2]: Right now this is equivalent to recomp address because we sort the dict in `CvdumpAnalysis` before inserting to the database.

disinvite:
Resolves #28.

We use the existing `logger` to display the warning, so brace yourself for more messages. Sample:
```
[WARNING] Ambiguous match 0x10058980 on name '_Construct' to '?_Construct@@YAXPAPAVLegoAnimPresenter@@ABQAV1@@Z'
[WARNING] Ambiguous match 0x1004a780 on name '_Construct' to '?_Construct@@YAXPAPAULegoPathCtrlEdge@@ABQAU1@@Z'
[ERROR] Failed to find function symbol with annotation 0x10022360 and name '?_Construct@@YAXPAPAVMxCore@@ABQAV1@@Z'
[WARNING] Ambiguous match 0x10010c00 on name 'Mx3DPointFloat::operator=' to '??4Mx3DPointFloat@@QAEAAV0@ABV0@@Z'
[WARNING] Ambiguous match 0x10064b20 on name 'Mx4DPointFloat::operator=' to '??4Mx4DPointFloat@@QAEAAV0@ABV0@@Z'
[ERROR] Failed to find function symbol with annotation 0x100c7ef0 and name 'list<MxNextActionDataStart *>::insert'
[ERROR] Failed to find function symbol with annotation 0x100c1bc0 and name 'list<MxDSAction *,allocator<MxDSAction *> >::insert'
```

Other notes:
- I dropped the `_find_potential_match` method in favor of two more general purpose search methods (on name/type and symbol). These return `MatchInfo` objects, both matched and unmatched, so it's up to the caller to check before writing to the DB with a call to `set_pair`.
- We now build indices at the moment they are needed in the two aforementioned methods. This should improve performance (but perhaps only slightly)
- We could very easily create a generic search function that would take any combination of keywords. The problem I found in testing is that SQLite will not engage the indices if you parameterize the key extracted from the JSON string. For example:
    ```sql
    SELECT cols FROM symbols WHERE json_extract(kvstore,?) = ?
    ```
    The workaround is to use string concat instead of a parameter, but this of course leads to the dreaded SQL injection problem. We probably don't need to worry about this too much considering the lifespan and purpose of the data. I'll add this in a separate PR later when we need it.

### DIFF
--- a/reccmp/isledecomp/compare/db.py
+++ b/reccmp/isledecomp/compare/db.py
@@ -24,8 +24,6 @@ _SETUP_SQL = """
         value text,
         primary key (addr, name)
     ) without rowid;
-
-    CREATE INDEX `symbols_na` ON `symbols` (json_extract(kvstore, '$.name'));
 """
 
 
@@ -93,6 +91,7 @@ class CompareDb:
     def __init__(self):
         self._sql = sqlite3.connect(":memory:")
         self._sql.executescript(_SETUP_SQL)
+        self._indexed = set()
 
     @property
     def sql(self) -> sqlite3.Connection:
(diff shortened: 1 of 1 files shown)


reccmp/isledecomp/compare/db.py
```diff
@@ -388,52 +387,79 @@ def is_vtordisp(self, recomp_addr: int) -> bool:
...
+                    """SELECT count(rowid) from symbols
+                    where json_extract(kvstore,'$.name') = ?
+                    AND json_extract(kvstore,'$.type') = ?""",
+                    (name, compare_type),
+                ).fetchone()
 
-        return self.set_pair(addr, recomp_addr, compare_type)
+                if matched and count > 1:
+                    logger.warning(
+                        "Ambiguous match 0x%x on name '%s' to '%s'",
```
jonschz, line 454: What do you think about logging *all* matches for an ambiguous name instead of logging just the one that is chosen automatically? It might help finding the correct one faster, though you'd probably have to look into the cvdump in any case.

reccmp/isledecomp/compare/db.py
```diff
@@ -470,16 +496,16 @@ def match_vtable(
         for_vftable = f"{name}::`vftable'{{for `{for_name}'}}"
 
         # Try to match on the "vftable for X first"
-        recomp_addr = self._find_potential_match(for_vftable, SymbolType.VTABLE)
-        if recomp_addr is not None:
-            return self.set_pair(addr, recomp_addr, SymbolType.VTABLE)
+        for obj in self.search_name(for_vftable, SymbolType.VTABLE):
+            if obj.orig_addr is None and obj.recomp_addr is not None:
+                return self.set_pair(addr, obj.recomp_addr, SymbolType.VTABLE)
```
jonschz, line 501: Maybe add a comment here that we intentionally return the first match and ignore others. Same below

jonschz (approved):
Looks great overall! Feel free to merge after responding to the suggestions I made.

Just curious: What else would you like to see/implement in order to consider #28 solved? In my opinion, this PR is good enough.

reccmp/isledecomp/compare/db.py
disinvite: The way it is now, it will warn for each match if the name and type appear more than once. It will warn even if we pick the only unmatched option remaining. We don't warn if the match failed (because the orig. address was already used.)

Do you mean displaying a summary of each non-unique name/type combination at the end? For example: if there are 5 `_Construct` functions and we matched 3 of them, we may have picked incorrectly 3 times.

reccmp/isledecomp/compare/db.py
```diff
@@ -388,52 +387,79 @@ def is_vtordisp(self, recomp_addr: int) -> bool:
...
+                    (name, compare_type),
+                ).fetchone()
 
-        return self.set_pair(addr, recomp_addr, compare_type)
+                if matched and count > 1:
+                    logger.warning(
+                        "Ambiguous match 0x%x on name '%s' to '%s'",
+                        addr,
+                        name,
+                        obj.get("symbol", "__no_symbol__"),
```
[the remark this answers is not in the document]
disinvite: It also looks like a symbol itself, so bad choice all around. I'll fix it. As for the symbol not being there: it probably wouldn't happen in our case because we're matching from the S_PUB32/S_GPROC32 entries. But just in case, maybe writing `None` would be better here?

reccmp/isledecomp/compare/db.py
jonschz: Let's say we have one annotation with an ambiguous name `examplefn` (maybe a `// LIBRARY`), and there are two symbols with that name (lets say `?SYMBOL1` and `?SYMBOL2`). The way I understand the current code, it would output one warning `Ambiguous match 0xabcdabcd on name 'examplefun' to '?SYMBOL1`, but `?SYMBOL2` never makes it to the logger. I was wondering if it makes sense to log `Ambiguous match 0xabcdabcd on name 'examplefun' to '?SYMBOL1, could also refer to ?SYMBOL2, ...`.

The current code looks very reasonable if the number of annotations matches the number of symbols; my idea might improve the developer experience in other situations.

reccmp/isledecomp/compare/db.py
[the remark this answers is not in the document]
jonschz: `None` sounds like a good choice to me

reccmp/isledecomp/compare/db.py
disinvite: I added a docstring and swapped the search order around. I always have to remind myself how this function works so hopefully it's way more clear now. Let me know what you think!

reccmp/isledecomp/compare/db.py
```diff
@@ -388,52 +387,79 @@ def is_vtordisp(self, recomp_addr: int) -> bool:
...
-        # by adding the plus sign (Reference: https://www.sqlite.org/optoverview.html#uplus)
-        if match_decorate:
-            # TODO: Change when/if decorated becomes a unique column
-            for (recomp_addr,) in self._sql.execute(
-                "SELECT recomp_addr FROM symbols WHERE json_extract(kvstore, '$.symbol') = ? AND +orig_addr IS NULL LIMIT 1",
-                (name,),
-            ):
-                return recomp_addr
+    def search_symbol(self, symbol: str) -> Iterator[MatchInfo]:
+        if "symbol" not in self._indexed:
```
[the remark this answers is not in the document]
disinvite: If we eventually add the search function for any key from the JSON string, this would be populated from what's in the kwargs dict

reccmp/isledecomp/compare/db.py
disinvite: Yeah, that's a good idea. I also think we should _not_ warn if there are multiple options but you have eliminated all but one by matching on either the line number or a symbol (i.e. match types without ambiguity). This doesn't happen now because the code parser doesn't properly capture the function's name, but that will change with the lexer/token-based parser.

This would require changing the approach to display a summary at the end for each duplicate name/type in the DB if:
- At least one was matched by name and any are unmatched
- More than one matched by name

I think that's easier than it sounds so I could do it now, or we can push this as a stopgap to fix the outstanding problems. What do you think?

jonschz: I like your idea, but I think it makes more sense to implement it when the parser refactor has been completed.

jonschz:
Looks good, feel free to merge 👍

PR #30, opened 2024-12-09
Modernize roadmap, type cvdump parser

### LINKED ISSUE #10: Roadmap tool should use project file
`reccmp-roadmap` takes the same four args as `reccmp-reccmp`, so we should read the yaml file and accept `--target` here too.

jonschz:
* `roadmap` now uses the project config like all other scripts. Closes #10.
* `roadmap` was broken on master due to a regression, which I also fixed. Tests would probably be a good idea, but I'm not sure where to start and if it would be worth it.
* I typed parts of the code and the parsing code of `cvdump`.
* When trying to type the code I noticed that some of the `--order`-specfic code might be broken. I am not really sure how it is supposed to be used anyway. Input would be appreciated.

### DIFF
--- a/reccmp/tools/roadmap.py
+++ b/reccmp/tools/roadmap.py
@@ -9,19 +9,26 @@ in the original binary.
 import os
 import argparse
 import logging
+from pathlib import Path
 import statistics
 import bisect
 from typing import Iterator, List, Optional, Tuple
 from collections import namedtuple
 import reccmp
 from reccmp.isledecomp import Bin as IsleBin
-from reccmp.isledecomp.bin import InvalidVirtualAddressError
+from reccmp.isledecomp.bin import Bin, InvalidVirtualAddressError
+from reccmp.isledecomp.compare.db import MatchInfo
 from reccmp.isledecomp.cvdump import Cvdump
 from reccmp.isledecomp.compare import Compare as IsleCompare
 from reccmp.isledecomp.types import SymbolType
+from reccmp.project.detect import (
+    argparse_add_built_project_target_args,
+    argparse_parse_built_project_target,
+)
+from reccmp.project.error import RecCmpProjectException
+from reccmp.project.logging import argparse_add_logging_args, argparse_parse_logging
 
-# Ignore all compare-db messages.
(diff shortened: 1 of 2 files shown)


reccmp/tools/roadmap.py
```diff
@@ -33,9 +40,11 @@ class ModuleMap:
     """Load a subset of sections from the pdb to allow you to look up the
     module number based on the recomp address."""
 
-    def __init__(self, pdb, binfile) -> None:
-        cvdump = Cvdump(pdb).section_contributions().modules().run()
-        self.module_lookup = {m.id: (m.lib, m.obj) for m in cvdump.modules}
+    def __init__(self, pdb: Path, binfile: Bin) -> None:
```
disinvite, line 43: Could replace `Bin` with `IsleBin` here and condense lines 18 and 19. (This may change after #26 anyway)

disinvite (approved):

jonschz:
Thanks! I'll leave the PR open for another day or so in case someone wants to respond to the other points I raised. If not, I assume we'll address them when they become a problem.

disinvite:
Some comments:

First off: it must be said that this script is a little rough, and you could argue that it lacks a clear purpose. My reason for creating it was to identify smartheap functions in ISLE (see `library_smartheap.h`). It was helpful to see where there was a gap between known functions. The main `asmcmp` script doesn't show _everything_.

The `--order` feature came after. It makes an educated guess about how to order the modules in the makefile. For a given module, MSVC will put certain functions a long way from its neighbors. We try to remove the outliers to get the best average of where the module's contents should sit. You can compute this estimate for things other than functions, too. For example, vtables: `--order VTA`. AFAIK we haven't applied this information to `LEGO1` but since we are coming to the end of new functions, maybe now is the time.

As far as tests: aside from one to just make sure it runs, maybe we could add some around the display features? For example: there is a bug in `get_cmakefiles_prefix` since I am now using ninja to build and the slashes go the other way. I think we could use `pathlib` to normalize this.

On line 244: `print(prefix)`. I forget why this is there and it can probably be taken out.

PR #33, opened 2024-12-13
Merge match_options table into main

disinvite:
We had the `match_options` table as a second key-value store. It was only used to store the booleans for "stub" (don't compare asm) and "skip" (don't compare and hide this symbol from reccmp output). Those are now part of the JSON store in the main table, and you can access the fields from the `MatchInfo` object using `get()`.

Small change to behavior because the two API functions `mark_stub` and `skip_compare` will now _not_ create a new entry in the DB with the original address. I dropped the test that checked for this because the `get_match_options` function is gone. I think that's reasonable for these two options because you don't need to skip or ignore what isn't there.


jonschz (approved):
Looks good! One thing: Are there any merge conflicts with #26 ? Should we give that PR a chance to be merged, since it appears to be ready soon?

disinvite:
> Looks good! One thing: Are there any merge conflicts with #26 ? Should we give that PR a chance to be merged, since it appears to be ready soon?

Yeah, let's do that one first

PR #38, opened 2024-12-18
Cache DB lookups for asm sanitize

disinvite:
The assembly generate/sanitize/compare process is the last step in the pipeline for `reccmp`. We don't make any changes to the database between comparing each function, so it's no problem to cache the address -> string lookups across the entire session. This gets us about a 5% performance boost. I diffed the full asm output before and after (using `--debug`) and the result is the same.

I also consolidated the "name lookup" function to use a generator, as with the "binary read" and "check relocation table" functions. I have some ideas to improve this area coming soon.

### DIFF
--- a/reccmp/isledecomp/compare/asm/parse.py
+++ b/reccmp/isledecomp/compare/asm/parse.py
@@ -46,12 +46,13 @@ class ParseAsm:
     def __init__(
         self,
         relocate_lookup: Optional[Callable[[int], bool]] = None,
-        name_lookup: Optional[Callable[[int, bool], str]] = None,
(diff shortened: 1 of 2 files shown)


jonschz (approved):
The code looks good - I'll trust that you have tested this thoroughly. I like the refactor!

Just a question: Have you considered using `functools.cache`? Main downside I could think of is that it's harder to clear an existing cache - not sure how important that is in practice.

disinvite:
Thanks! `cache` is simpler and performance is the same or better. I suppose you would clear the cache by calling the generation function again to get a new `lookup` function. If we had some interactive component to `reccmp` you could re-run the asm sanitize process after making some changes to the database, but we don't have that feature (yet).

PR #40, opened 2024-12-24
Implement importing vftables into Ghidra

jonschz:
This is a feature I should have implemented earlier. Will apply the change to the shared Ghidra repos when it has been reviewed.

Merry Christmas!


disinvite:
Tested this earlier and it looks good! 🎄

PR #41, opened 2024-12-25
Bump dependencies and update CI

larziwau:


larziwau:
@foxtacles

jonschz (commented):
The changes look good so far. The new version of `pylint` appears to complain about an issue that the old version didn't notice, which causes the pipeline to fail. Could you briefly look into that, or defer the update on `pylint`?

@disinvite Could you take a second look? I have barely worked with GitHub actions in the past. It mostly looks like reformatting and using newer versions to me.

PR #42, opened 2024-12-26
Improve type safety, part 1

jonschz:
This is a first step in moving towards `mypy` compliance in this repository. Since the number of changes is quite large already, I will split the work over multiple PRs. I will leave comments on the parts where slight changes to the logic were made.

This change fixes most (but not all) mypy issues for `mypy --check-untyped-defs .\reccmp\isledecomp\`. Configuration was not yet set up deliberately - I will do this once `mypy` passes for most of the code.


reccmp/isledecomp/compare/asm/instgen.py
```diff
@@ -24,9 +23,20 @@ class SectionType(Enum):
...
+
+TabSectionType = Literal[SectionType.DATA_TAB] | Literal[SectionType.ADDR_TAB]
+
+
+class TabSection(NamedTuple):
+    type: TabSectionType
+    contents: List[Tuple[int, int]]
+
+
+FuncSection = CodeSection | TabSection
```
jonschz, line 39: This is effectively what the code did before. The difference is that you can check for `type` now and then deduce the type of `contents`.

reccmp/isledecomp/compare/asm/parse.py
```diff
@@ -201,15 +199,15 @@ def sanitize(self, inst: DisasmLiteInst) -> Tuple[str, str]:
 
         return (inst.mnemonic, op_str)
 
-    def parse_asm(self, data: bytes, start_addr: Optional[int] = 0) -> List[str]:
+    def parse_asm(self, data: bytes, start_addr: int = 0) -> list[tuple[str, str]]:
```
jonschz, line 202: The `Optional[]` was unnecessary.

reccmp/isledecomp/compare/core.py
```diff
@@ -24,6 +24,9 @@
 logger = logging.getLogger(__name__)
 
 
+NamedOrigEntry = TypedDict("NamedOrigEntry", {"orig": int, "name": str})
```
jonschz, line 27: Not too happy with this solution since this type is only used internally in one function. There's likely a better solution.

reccmp/isledecomp/compare/core.py
```diff
@@ -702,14 +712,17 @@ def _compare_function(self, match: MatchInfo) -> DiffReport:
         # Detect when the recomp function size would cause us to read
         # enough bytes from the original function that we cross into
         # the next annotated function.
+        assert match.orig_addr is not None
+        assert match.recomp_addr is not None
+        assert match.size is not None
```
jonschz, line 723: This is required a bunch of times. It might make sense to define another structure where all of these are not `None`, but that's out of the scope of this PR.

reccmp/isledecomp/formats/pe.py
```diff
@@ -995,3 +1028,33 @@ def read(self, vaddr: int, size: int) -> Optional[bytes]:
         # Reading off the end will most likely misrepresent the virtual addressing.
         _size = min(size, section.size_of_raw_data - offset)
         return bytes(section.view[offset : offset + _size])
+
+    def read_initialized(
```
jonschz, line 1032: This looks like it is a bigger change, but it's rather making explicit what was implicit earlier. Previously, if you got an unexpected `None` in `read()`, you would also run into an exception. I went over all uses of `read()` and replaced the ones that would have crashed on a `None` response.

reccmp/isledecomp/compare/core.py
```diff
@@ -199,11 +204,11 @@ def _load_cvdump(self):
                 if string_info.is_utf16:
                     continue
 
-                raw = self.recomp_bin.read(addr, sym.size())
+                size = sym.size()
+                assert size is not None
 
-                # read returns None when reading 0 bytes
```
jonschz, line 204: This comment was actually wrong - I added a unit test to verify that reading 0 bytes returns `b""`

reccmp/isledecomp/formats/pe.py
jonschz: obsolete now by #43

reccmp/isledecomp/compare/core.py
jonschz: see #45

jonschz:
- [ ] blocked by #45

reccmp/isledecomp/dir.py
```diff
@@ -27,23 +27,23 @@ def __init__(self, basedir) -> None:
...
+            self._basedir: PurePath = PureWindowsPath(
                 winepath_unix_to_win(str(self._realdir))
             )
         else:
             self._basedir = self._realdir
 
     def _memo_wrapper(self, path_str: str) -> str:
         """Wrapper so we can memoize from the public caller method"""
-        path = pathlib.PureWindowsPath(path_str)
+        path: PurePath = PureWindowsPath(path_str)
```
disinvite, line 46: The type hint is fine, but just a thought: I think this is a consequence of using the `path` variable as both a `pathlib.PurePath` and `pathlib.Path`. (The fact that one extends the other only adds to the confusion)

We should maybe have `PathResolver` return a completely different object on Windows and Unix to simplify this. (See #53)

reccmp/isledecomp/formats/detect.py
```diff
@@ -8,7 +8,7 @@
 from .pe import PEImage
 
 
-def detect_image(filepath: str | Path) -> Image:
+def detect_image(filepath: Path) -> Image:
```
disinvite, line 11: I think we should support both here because this is a "customer-facing" part of the interface. Internally we should continue to use `Path` but I think a user writing a custom script would expect to provide a string here and have it work.

reccmp/isledecomp/cvdump/analysis.py
```diff
@@ -93,7 +94,7 @@ class CvdumpAnalysis:
     """Collects the results from CvdumpParser into a list of nodes (i.e. symbols).
     These can then be analyzed by a downstream tool."""
 
-    verified_lines: Dict[Tuple[str, str], Tuple[str, str]]
+    verified_lines: dict[tuple[int, int], Tuple[str, int]]
```
disinvite, line 97: lower case `tuple` here

reccmp/isledecomp/parser/parser.py
```diff
@@ -264,7 +264,7 @@ def _vtable_marker(self, marker: DecompMarker):
             self._syntax_warning(ParserError.DUPLICATE_MODULE)
         self.state = ReaderState.IN_VTABLE
 
-    def _vtable_done(self, class_name: str = None):
+    def _vtable_done(self, class_name: str | None = None):
```
disinvite, line 267: Pretty sure this could not get called with `None`, so you could maybe remove the next three lines and leave this as just `class_name: str` with no default. There's a small bug here too that I will look into.

reccmp/tools/datacmp.py
```diff
@@ -247,7 +250,7 @@ def do_the_comparison(target: RecCmpBuiltTarget) -> Iterable[ComparisonItem]:
                         offset=0,
                         name="(raw)",
                         match=orig_raw == recomp_raw,
-                        values=(orig_raw, recomp_raw),
+                        values=(str(orig_raw), str(recomp_raw)),
```
disinvite, line 253: Would this do the same thing?
```python
class ComparedOffset(NamedTuple):
    offset: int
    # name is None for scalar types
    name: str | None
    match: bool
    values: tuple[str | bytes, str | bytes] # <-- this line
```

reccmp/tools/datacmp.py
```diff
@@ -130,6 +130,7 @@ def create_comparison_item(
     """Helper to create the ComparisonItem from the fields in the reccmp database."""
     if compared is None:
         compared = []
+    assert var.name is not None
```
disinvite, line 133: What about `var.name or ""` when creating the `ComparisonItem`?

reccmp/isledecomp/compare/core.py
```diff
@@ -176,11 +181,11 @@ def _load_cvdump(self):
...
-                if sym.size() == 0 and raw is None:
-                    raw = b""
+                # Assume an empty string for uninitialized memory
+                raw = self.recomp_bin.read(addr, size) or b""
```
disinvite, line 188: The latest change to `read()` should cover this and we can drop the `or b""`

reccmp/isledecomp/compare/core.py
```diff
@@ -299,8 +305,7 @@ def _match_array_elements(self):
         Note that there is no recursion, so an array of arrays would not be handled entirely.
         This step is necessary e.g. for `0x100f0a20` (LegoRacers.cpp).
         """
-
-        dataset = {}
+        dataset: dict[int, NamedOrigEntry] = {}
```
disinvite, line 308: What about this? (after importing `typing.Any`)
```python
# recomp -> key/val
dataset: dict[int, dict[str, Any]] = {}
# orig -> recomp
pairs: dict[int, int] = {}
```
Then:
```python
if recomp_addr not in dataset:
    dataset[recomp_addr] = {"name": name}
    pairs[orig_addr] = recomp_addr
```
Finally:
```python
self._db.bulk_recomp_insert(dataset.items(), upsert=True)
self._db.bulk_match(pairs.items())
```

reccmp/isledecomp/formats/pe.py
```diff
@@ -31,6 +31,10 @@ class UnknownPEMachine(ValueError):
     """The PE binary has an unknown machine architecture."""
 
 
+class UninitializedDataReadError(ValueError):
```
disinvite, line 34: No longer relevant?

reccmp/isledecomp/formats/pe.py
```diff
@@ -437,7 +444,9 @@ class DebugDirectoryEntryHeader:
     pointer_to_raw_data: int  # The file pointer to the debug data.
 
     @classmethod
-    def from_memory(cls, data: bytes, offset: int) -> "DebugDirectoryEntryHeader":
+    def from_memory(
+        cls, data: bytes, offset: int
+    ) -> "tuple[DebugDirectoryEntryHeader, int]":
```
disinvite, line 449: Does it work without quotes around the whole thing? Should be consistent with other `tuple["name_of_this_class"]` appearances

reccmp/isledecomp/compare/core.py
```diff
@@ -920,4 +931,4 @@ def compare_vtables(self) -> Iterable[DiffReport]:
         for match in self.get_vtables():
             diff = self._compare_match(match)
             if diff is not None:
-                yield self._compare_match(match)
+                yield diff
```
disinvite, line 934: lol

disinvite:
Very nice work! Agree on delaying the new CI action and replacing `Optional` with `| None` (and also delaying it).

reccmp/isledecomp/compare/core.py
jonschz: obsolete now

reccmp/isledecomp/dir.py
jonschz: Good point! I agree that it's fine like this, any refactors should be made in the fix of #53

reccmp/tools/datacmp.py
jonschz: As far as I can tell, it would. However, considering that the values are only used to display the values, I feel like my solution gets the intention across better. Another solution would be to use `bytes | None` and add the `"(uninitialized)"` at the place where it is printed.

reccmp/tools/datacmp.py
jonschz: Is there a known way to violat this assertion? I couldn't find one easily. If there isn't, I'd prefer to see an error rather than an unexpected empty string, since that will be much easier to debug.

reccmp/isledecomp/compare/core.py
jonschz: That would work, but the typing is weaker. For example, you don't get IDE support or type checking for dataset[reccmp_addr].name`.

However, I have found a refactor to a dataclass I like. Please tell me what you think (after I have pushed my change).

reccmp/isledecomp/formats/pe.py
jonschz: good catch!

reccmp/isledecomp/formats/pe.py
jonschz: Odd - I could have sworn that some tool (either `pylint` or `mypy` wanted me to change this, but now I can't reproduce the issue. Thanks!

jonschz:
Thanks for the thorough review! will merge once your have responded to my comments on the unresolved threads.

reccmp/isledecomp/compare/core.py
jonschz: see https://github.com/isledecomp/reccmp/pull/42/files#diff-6cfae8dfa8a79c361f570be094791dca228a63dfdb0bd655db9fc3889379b58dR297-R300

disinvite: Using a one-off dataclass for this feels like overkill if it's only there to squash the type checker alert. Does `mypy` complain about it?

jonschz: It accepts the solution I have, and presumably also the solution with the `Any`. I just try to avoid `Any` unless there's a lot of polymorphism or uncertainty, which we don't have here. I would have liked to use the `TypedDict` locally/anonymously, but that does not appear to be possible - see https://github.com/python/mypy/issues/9884. Maybe an ordinary `tuple` or a `NamedTuple` might also be an option here?

disinvite: That's a good point. What about just `dataset: dict[int, dict[str, str]]` for now, since `"name"` is the only key we're setting?

jonschz: We are also setting `orig_addr`, which is an `int`, so that wouldn't work. From what I can tell, `TypedDict` would be the correct annotation if we don't want to use dataclasses: https://typing.readthedocs.io/en/latest/spec/typeddict.html

jonschz: But I'd rather change it to `Any` before the discussion gets out of hand 😉

disinvite: I meant with the two dicts like this:
```python
dataset: dict[int, dict[str, str]] = {}
pairs: dict[int, int] = {}
```

PR #43, opened 2024-12-26
Allow reads on uninitialized data

disinvite:
Resubmit of #36. I added more comments and tests and took out the `va_space` pre-allocation. Since there is the potential for this to require a huge memory footprint, it's no longer worth the minor performance/simplicity benefit.

`get_relative_addr()` previously used `bisect` to more quickly find the section for the virtual address. This proved to be slower than linear search, so it's been removed.

The main change is to the `read()` method in `PEImage`. We previously treated reads on uninitialized data as a special case: return `None` if the data is entirely uninitialized, or crop the byte string returned. Since the real program doesn't make any distinction between bytes initialized to zero in the physical file and those set to zero during image load, we do the same. We will raise an exception if you read between sections in the PE file or past the end of the virtual address window.

`datacmp` has been changed accordingly. It still reports if a variable is partially uninitialized (and thus falls at the end of physical data for the section) but otherwise treats uninitialized data as if it were fully initialized to zero.

### DIFF
--- a/reccmp/isledecomp/formats/exceptions.py
+++ b/reccmp/isledecomp/formats/exceptions.py
@@ -5,3 +5,9 @@ class SectionNotFoundError(KeyError):
 class InvalidVirtualAddressError(IndexError):
     """The given virtual address is too high or low
     to point to something in the binary file."""
+
+
+class InvalidVirtualReadError(IndexError):
+    """Reading the given number of bytes from the given virtual address
+    would cause us to read past the end of the section or past the end
+    of the virtual address space."""
--- a/reccmp/isledecomp/formats/pe.py
+++ b/reccmp/isledecomp/formats/pe.py
@@ -5,7 +5,6 @@ Based on the following resources:
 - Debug information: https://www.debuginfo.com/examples/src/DebugDir.cpp
 """
 
-import bisect
 import dataclasses
 from enum import IntEnum, IntFlag
 from functools import cached_property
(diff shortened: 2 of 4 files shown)


madebr (approved):

jonschz:
Since I don't have that much of a background in how the binary file is loaded, a few basic questions:
- Are we sure that uninitialized sections are always loaded with zeros?
- Are we sure that it's up to the compiler whether a given variable is initialized? Is there a chance of introducing false negatives here? (variables that should be flagged by datacmp but are not)

jonschz (approved):
Can't test the code at the moment, but it looks good

disinvite:
> Are we sure that uninitialized sections are always loaded with zeros?

I'm basing that assumption off of [this page](https://learn.microsoft.com/en-us/windows/win32/debug/pe-format#section-table-section-headers), specifically:

> **VirtualSize**
> The total size of the section when loaded into memory. If this value is greater than SizeOfRawData, the section is zero-padded. 

> Are we sure that it's up to the compiler whether a given variable is initialized? Is there a chance of introducing false negatives here? (variables that should be flagged by datacmp but are not)
I'll give an example for this, but I don't have a reference. It's based on observations from what's in `LEGO1.DLL`.

Assume `.data` has these attributes:
- **VirtualAddress**: `0x1000`
- - **VirtualAddress**: `0x1000`

jonschz:
Sounds good, thanks for the explanation!

jonschz (approved):

Issue #3, opened 2024-10-20
New binary files for unit tests


disinvite:
The unit tests in `tests/test_islebin.py` depend on `LEGO1.DLL` and make sure we correctly interpret various points of interest from the file. We need a new generic file (or files) to adapt these tests for general use.

Some things to consider:
- [x] Decide which types of file we want to test. We currently only read PE-format files. Do we want to look at MZ or NE executables in the future?
- [x] Test for detecting compiler-specific metadata? e.g. MSVC debug information like build timestamp, etc
- [ ] Decide where the code for these dummy executables will live and how they will get built.
- [ ] Exclude these files from the python module install. They should only exist here for CI or on reccmp dev setups
- [ ] Right now we just test for reading the data structures out of the file. We could test more of the analysis tools as we add them (e.g. function/vtable range detection. ascii string detection)

disinvite:
#26 has raised this topic again. I think open questions 1 and 2 above (and probably 5) can be answered with "yes, we should do these things". The challenge now is creating these files. Or should we just find some files that would meet the criteria in the meantime? (Old shareware games or whatever)

madebr:
My intention in #26 was to add small executables completely under our control.

[win32_binaries.zip](https://github.com/user-attachments/files/17972540/win32_binaries.zip) built with MSVC 6
By avoiding the CRT, and linking with `/ENTRY`, it's possible to keep the file sizes low.

disinvite:
Pulling the discussion back over here. @madebr brought up open watcom as a modern way to build old exes without dosbox/dosemu and an old compiler. We have some sample files (for PE) to start with. But where do we put them?

- Build the files somewhere, bundle them here in the `reccmp` repo.
  **Pros:** Easy access, every contributor can test locally, can track changes to the binaries and companion tests.
  **Cons:** Awkward to ship lots of exe files with the module. Potential for virus scan alerts?

- Build files from source on a separate repo. 
  **Pros:** Clear separation of code and sample files.
  **Cons:** But now they are separate. Need to keep tests (in this repo) and binary changes in sync. Would need to download files locally to test locally.

And would we have separate binary files for each test case?

madebr:
I dislike it when binaries are added to git.
At libsdl-org/SDL_image and libsdl-org/SDL_mixer they added x86/x64 dlls for commen encoders/decoders.
The `.git` directory of those projects are close to 1GiB and 250 MiB because of this.

[...]

PR #44, opened 2024-12-28
Start on support for NE EXE

disinvite:
Got some NE format programs I want to start looking at. I followed the layout of the other file formats and used the same member names as the Windows header files.


jonschz:
The code looks fine from what I can tell. Is this a WIP or do you want to merge this in batches? Also, do you have an example file for experiments / also potentially for a unit test?

disinvite:
This is it for now. I had these files unstaged and thought I would finish them up and get them out of the way.

We should have tests, but there is still the question (from #3) of where to put them and how to build. NE might require an old compiler too.

madebr:
Do you know what compiler your executables were created with?
OpenWatCom v2 is maintained at https://github.com/open-watcom/open-watcom-v2/ and supports DOS, OS/2, 16-bit Windows and other vintage platforms.

jonschz (approved):
Then I guess it makes sense to merge this and fix any issues when we encounter them.

PR #45, opened 2024-12-28
Use 'Matched' ORM object if both addrs are set

disinvite:
My attempt to stamp out some of the type issues surrounding the `MatchInfo` object returned from the database. (See #42)

Database methods like `get_matches()` return matches (obviously) where both addresses are set. Others like `get_all()` return objects where one address can be null. We'll use two different objects for this situation. Since it would be awkward to have `MatchInfoMatch`, I named the objects `ReccmpThing` and `MatchedReccmpThing`. [Temporary name...](https://www.imdb.com/title/tt0079945/quotes/#:~:text=Why%20is%20any%20object%20we%20don%27t%20understand%20always%20called%20%22a%20thing%22%3F) or not?

Another change is to have the `size` helper property return 0 if size is null. I _think_ this is okay because it matches the user expectation. If size is not set, you can't really make a useful comparison on whatever item is under test. The exception is when the item has some intrinsic size, but then you would just set it. That being said, I recall this being a sticking point earlier when trying out various database changes, so maybe someone could double check my logic here.


jonschz:
Looks very promising overall! Some name suggestions:
- `MatchCandidate` and `Match`
- `ReccmpPartialMatch` and `ReccmpMatch`

disinvite:
What do you think about `ReccmpComponent` and `ReccmpMatch`? Could replace "component" for another word, but it's hard to find one that fits. e.g.
- Entry (of the database)
- Item (vague)
- Symbol (not all entries have linker symbols)

I think the superclass name should not have the word "match" to convey that matches are a subset of the total list of items.

jonschz:
How about "Entity"? Similar to "Entry" in its pros and cons, but maybe slightly better. ReccmpComponent and ReccmpMatch are fine with me, too.

disinvite:
I'll take it! We now have `ReccmpEntity` and `ReccmpMatch`. The only new change other than the rename is to use `ReccmpMatch` in `PdbFunction`. It gets created in `handle_matched_function` so that type seems to fit better.

reccmp/isledecomp/compare/db.py
```diff
@@ -44,12 +61,12 @@ def name(self) -> Optional[str]:
         return self.options.get("name")
 
     @property
-    def size(self) -> Optional[int]:
-        return self.options.get("size")
+    def size(self) -> int:
+        return self.options.get("size", 0)
```
jonschz, line 65: I'd add a docstring explaining that zero is returned when no size attributes is present

jonschz (approved):
Looks good!

PR #48, opened 2024-12-29
Minor improvements to reccmp project create

disinvite:
Added some comments and docstrings for the `create_project` method. I moved some failure checks to the front and added a couple unit tests to make sure we handle them as expected. New exception `RecCmpProjectAlreadyExistsError` with default error text.

Moved some code that made sure repeated base filenames would get unique target ids into a `unique_targets` utility method.

Fixed a bug where `create` and `detect` actions would always show the `"Missing command"` error.

@madebr


madebr (approved):

reccmp/tools/project.py
```diff
@@ -123,7 +123,8 @@ def main():
         except RecCmpProjectException as e:
             logger.error("Project detection failed: %s", e.args[0])
 
-    parser.error("Missing command: create/detect")
+    else:
+        parser.error("Missing command: create/detect")
 
     # try:
```
jonschz, line 129: Is this commented code still relevant? Maybe it can be removed during this refactor?

reccmp/tools/project.py
```diff
@@ -106,7 +106,7 @@ def main():
         except RecCmpProjectException as e:
             logger.error("Project creation failed: %s", e.args[0])
 
-    if args.action == "DETECT":  # FIXME: use enum or callback function
+    elif args.action == "DETECT":  # FIXME: use enum or callback function
```
jonschz, line 109: Just a suggestion: Maybe this is a good moment to change this to an enum?

jonschz (approved):
Looks good!

reccmp/tools/project.py
disinvite: Yeah I think the argparse subparser will error earlier than this.

disinvite: We do get here if you run `reccmp-project` with no subcommand

disinvite:
That should cover both comments. I renamed the `action` to `subcommand` just so it's distinct from the `action` kwarg used in other `argparse` methods.

PR #49, opened 2024-12-30
Refactor 32-bit asm sanitize, rewrite tests

disinvite:
Some refactoring around the asm sanitize function.

I rewrote all the tests (with a couple carried over) in a new file `test_sanitize32.py`. (16-bit asm sanitize will work differently.) I think the layout for these tests is a lot better and more clearly defines the behavior.

One new tweak has to do with placeholder numbers. (i.e. the `<OFFSET123>` things.) If we don't have a name and need to use a placeholder, the numbering includes the previously replaced addresses that may or may not have a name. This is to increase the readability of the diff. It makes more sense to mark "the third address used in the function" rather than "the first address that doesn't have a name".

The problem is that in some situations we don't use a placeholder at all. We will use a name if we have one, but otherwise we leave the hex value alone. These are:
- Instructions like `cmp eax, <value>`. We intentionally don't replace here because loops on arrays of structs might pick the address of the following variable as the range limit. If that's hidden behind a placeholder we might miss a diff.
- Jumps directly to another function. This could happen in handwritten assembly but where we've seen it is calls to destructor functions for objects `new`'d in the same function.

If we have a name for the value in these instructions, we will use it, but not set it in the `replacements` dict. This should keep the offset numbers consistent and increase match accuracy early on in a decomp. (The lookups are cached upstream in `create_name_lookup` so I expect no performance penalty.)

To give an example of what I'm talking about, here are the replacements from `Score::DeleteScript` at `0x10001340`.

|Address | Replacement | Notes |
|--------|-------------|-------|
|`100013f5`  | OFFSET1 | Unwind label |
|`100ad810`  | `MxDSAction::MxDSAction` ||
|`100f454c`  | `g_infoscorScript` ||
|`100ad1c0`  | `MxAtomId::operator=`  ||
|`100013ed`  | OFFSET5 | Unwind label |
|`100acf70`  | `DeleteObject` ||
|`100013ff`  | OFFSET7 | Unwind label |
|`100acfd0`  | `MxAtomId::~MxAtomId` | Unwind destructor |
|`100dd2a8`  | OFFSET9 | FuncInfo struct |
|`1008b020`  | `___CxxFrameHandler` | Unwind exception handler |
|`100ada80`  | `MxDSAction::~MxDSAction` | Unwind destructor |

If we do not have the name for the `MxAtomId` destructor, `100acfd0` won't be part of the replacement list. This decrements all future placeholder numbers, and so `<OFFSET9>` would become `<OFFSET8>`. This creates the appearance of a diff where there might not be one.

Going forward, the two destructors and `___CxxFrameHandler` (called by jumps) are not part of this list, so the placeholder number will be consistent.

More on SEH and unwind sections [here](https://www.openrce.org/articles/full_view/21).

I moved `create_name_lookup` into its own module because I want to expand this further. Short-term goal: remove the hack that allows us to replace the names for imported functions by doing it properly here.

The function that tests whether immediate operands are addresses was called `relocate_lookup`. It's now simply `addr_test`. We can use the relocation table data if we have it. Testing against the imagebase is also fine, or even using a db lookup.


foxtacles (approved):

reccmp/isledecomp/compare/asm/parse.py
```diff
@@ -58,40 +59,38 @@ def __init__(
...
         return None
 
-    def replace(self, addr: int) -> str:
-        """Same function as lookup above, but here we return a placeholder
-        if there is no better name to use."""
-        if (name := self.lookup(addr)) is not None:
+    def replace(self, value: int, exact: bool = False, test: bool = False) -> str:
+        """Provide a replacement name for the given address."""
+        if test and not self.is_addr(value):
+            return hex(value)  # TODO.
```
jonschz, line 79: What is the status of this TODO?

jonschz (commented):
I haven't worked with the sanitizer in detail, but the change looks good!

reccmp/isledecomp/compare/asm/parse.py
disinvite: Right now we convert the hex string to a number (using `int(value, 16)`) and then call `lookup` or `replace`. I was thinking about changing this so that these functions accept the string value instead. This will have to be the case with 16-bit asm because we need to split the `segment:offset` pair for `lcall` instructions and then convert to some normalized value.

I'm not totally committed to this yet, but `replace()` has to return something if the address test fails. The compromise is to do the reverse conversion here. This is sort of wasteful and depends on python `hex` and the `capstone` library having compatible output.

I can try and address this now before merging. I think the tests cover this okay (what you get in is what you get back) but I'm becoming less and less happy with this solution as I continue to type out this comment 😀

jonschz: I'd either fix it or add a link to the comment you just posted to the TODO comment. Personally, I don't like TODO comments without further explanations on `master` since there's always a chance they stay there for a long time and don't make sense to future contributors.

disinvite:
I wound up reverting some of my changes to drop the reverse conversion (and the `# TODO`). We're okay for now but I see where there's more work to be done.

One problem is with the edge case of `call dword ptr [0x1000]`. That address can be an imported function, or it can arbitrarily point to a vtable entry as we saw with some of the `MxMatrix` functions. Where I'm stuck is how to deal with the replacement cache so you get a useful `<OFFSETX>` entry. The same address value can be used in the function in different situations, and I think the cache has to take that into account. But that will happen in future PRs.

jonschz:
As far as I am concerned you could merge this and maybe open an issue based on your most recent comment 👍

PR #50, opened 2025-01-01
Ghidra scripts: detect version-specific `site-packages` path

disinvite:
Not sure if this is a `virtualenv` thing or a `homebrew` thing or what, but on OSX my `site-packages` directory is on the path `.venv/lib/python3.12/site-packages`. The ghidra scripts expect the path to be `.venv/Lib/site-packages/`.

`lib` is lowercase here, expecting `Path.glob` to do a case-insensitive search on Windows. (https://docs.python.org/3/library/pathlib.html#pathlib.Path.glob)


jonschz (approved):
Works fine on my machine (Windows 10)

PR #52, opened 2025-01-02
Fix bug with `--paths` command line option

### LINKED ISSUE #46: `reccmp-reccmp` with `--paths` option skips code dir
Found by @Ramen2X. Running `reccmp-reccmp` with the `--paths` option (not using the yml project files) apparently skips annotations from the code directory. You only get 130 matches for `LEGO1.DLL`, which are the exported functions.

disinvite:
Resolves #46. The fix is to set `target-id` to `None` if you used the `--paths` command-line option instead of `--target`. See `argparse_add_built_project_target_args`.

This is important because the target name is used to isolate the code annotations we want to read. If it is `None` we set it in `core.py` using the base filename of the orig binary.

I added some comments to the datatype classes so it's faster to understand which things do what.

### DIFF
--- a/reccmp/project/config.py
+++ b/reccmp/project/config.py
@@ -23,14 +23,30 @@ class GhidraConfig(BaseModel):
 
 @dataclass
 class RecCmpTarget:
-    target_id: str
+    """Partial information for a target (binary file) in the decomp project
+    This contains only the static information (same for all users).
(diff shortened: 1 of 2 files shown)


Ramen2X (approved):
Works great for me, thanks!

Issue #53, opened 2025-01-03
Refactor `PathResolver`


disinvite:
The `LINES` section of the PDB references code filenames used in the build. These are Windows paths and might contain back-references (`..`) that we need to resolve.

On Unix platforms we use `winepath` to convert the PDB paths to your local filesystem. This can be an expensive lookup so we try to limit the calls via the following:

1. In the constructor, convert the base code directory (Unix path) to its Wine representation. (This is `self._basedir`)
2. When we need to convert a Windows path back to Unix, check if it is relative to `self._basedir`. If it is, add the relative path to the Unix path (using `pathlib` to flip the slashes the correct way). If this points to a file, we are done.
3. If that trick didn't work, do the `winepath` lookup as normal.

We also cache the outputs because we will get the same Windows path multiple times for different functions. On Windows `PathResolver` is a passthrough.

This has been working for a while, but it's not without its idiosyncracies:

- Wonky caching approach with `self._memo_wrapper`. It's awkward to use `functools.cache` on an instance method so I went with this, but it's not great.
- This whole process assumes that the PDB would never contain a path reference _outside_ the base code dir. On `LEGO1` this has been true, but `BETA10` inlines nothing and so it contains path references to `msvc420/include`. We have no base path for this directory so every one of these is a `winepath` lookup.
- Does not support multiple code directories (but the project files don't support this yet either).
- Do we want `reccmp` looking in places outside the code directories you specify? If we did want to scan the compiler include files for some reason, this currently "just works" but if we made it a requirement to explicitly add it as a code directory, it would create a new problem for the project file structure. The compiler directory would be different for each user instead of relative to the git root.
- Some better variable names and more comments would help here, especially where we are using `PurePath` (not necessarily a real path) and `Path` (real path) interchangeably where their API overlaps.
- If `winepath` is not available this probably blows up with a not-totally-helpful error. We should maybe have a way to test for this and report that we cannot interpret line number annotations without it.

disinvite:
Resolved by #101.

PR #58, opened 2025-01-04
Expand line number match to a range of lines

### LINKED ISSUE #57: Line number annotation match fails on edge case
See isledecomp/isle#1309. We want to annotate a function like this (line numbers added):
```cpp
34     // FUNCTION: LEGO1 0x10002380
35     // FUNCTION: BETA10 0x1000fd90
36     float (*Matrix4::GetData())[4]
37     {
38     	return m_data;
39     }
```
The `LINES` section in the PDB shows that this function begins on line 36. All other functions so far have "started" on the line with the curly brace (37), and this is what we expect when parsing. The curly brace is always on its own line because we enforce Allman-style indenting via `clang-format`.

You may ask: What is special about this function that causes MSVC to behave differently? I don't think this is worth pursuing (except for one's own curiosity) because other compilers can do as they please (re: line-to-address reporting) and `reccmp` has to work with the data it gets.

I think the short term solution is to match any of the lines between the end of the marker comments and the curly brace. This is prone to problems if you edit the file without recompiling (and updating the PDB), but this is already a limitation of the parser.

disinvite:
Resolves #57.

I had planned to eventually drop the sample files tested in `test_parser_samples.py` but this is one of the only places we actually test that the line numbers are captured correctly. I had some comments warning about detecting the end line of the function; this seems to work okay (with Allman syntax) so I enabled the tests for those. I don't think it's possible for `end_line` to be `None` but I will save that refactor for later.

The lines database definitely needs improvement (related to #53) but that's not for this PR.

### DIFF
--- a/reccmp/isledecomp/compare/lines.py
+++ b/reccmp/isledecomp/compare/lines.py
@@ -75,6 +75,17 @@ class LinesDb:
         if len(possible_functions) == 1:
             return possible_functions[0]
 
+        # The file has been edited since the last compile.
+        if len(possible_functions) > 1:
+            logger.error(
+                "Debug data out of sync with function near: %s:%d",
+                path,
+                line_start,
+            )
+            return None
+
+        # No functions matched. This could mean the file is out of sync, or that
+        # the function was eliminated or inlined by compiler optimizations.
         logger.error(
             "Failed to find function symbol with filename and line: %s:%d",
             path,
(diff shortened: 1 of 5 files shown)


foxtacles (approved):

reccmp/isledecomp/compare/lines.py
```diff
@@ -46,25 +44,40 @@ def add_line(self, path: str, line_no: int, addr: int):
...
-            if my_samefile(path, source_path):
-                return addr
+
+        possible_functions = [
+            addr for source_path, addr in cur if my_samefile(path, source_path)
+        ]
+        if len(possible_functions) == 1:
+            return possible_functions[0]
 
         logger.error(
```
jonschz, line 78: Although that should "never happen", I'd add a special handler for the case of multiple matches.

foxtacles (approved):

PR #59, opened 2025-01-04
Report function alignment accuracy

disinvite:
`reccmp-reccmp` now reports how many functions have the same virtual address in both the original and recompiled binary.

For example:
```
Total effective accuracy 97.83% across 4362 functions (97.03% actual accuracy)
68 functions are aligned (1.56%)
```

Should we add this to SVG output also?

### DIFF
--- a/reccmp/tools/asmcmp.py
+++ b/reccmp/tools/asmcmp.py
@@ -264,6 +264,10 @@ def main():
 
     ### Compare everything.
 
+    # Count how many functions have the same virtual address in orig and recomp.
+    functions_aligned_count = 0
+
+    # Number of functions compared (i.e. excluding stubs)
     function_count = 0
     total_accuracy = 0
     total_effective_accuracy = 0
@@ -274,6 +278,12 @@ def main():
             print_match_oneline(
                 match, show_both_addrs=args.print_rec_addr, is_plain=args.no_color
             )
+
+        if (
+            match.match_type == SymbolType.FUNCTION
+            and match.orig_addr == match.recomp_addr
+        ):
+            functions_aligned_count += 1
 
         if match.match_type == SymbolType.FUNCTION and not match.is_stub:
             function_count += 1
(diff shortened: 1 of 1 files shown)


reccmp/tools/asmcmp.py
```diff
@@ -322,15 +332,20 @@ def main():
 
     implemented_funcs = function_count
 
+    # Substitute an alternate value for the total number of functions in the file
```
jonschz, line 335: I like that you added multiple explanations! How about _Substitute an alternate value for the total number of functions in the target (e.g. to account for functions that have not been stubbed yet but have been counted by some analysis tool like Ghidra)_? Or have I misunderstood the purpose of this argument?

reccmp/tools/asmcmp.py
```diff
@@ -322,15 +332,20 @@ def main():
...
     if args.total:
         function_count = int(args.total)
 
     if function_count > 0:
         effective_accuracy = total_effective_accuracy / function_count * 100
         actual_accuracy = total_accuracy / function_count * 100
+        alignment_accuracy = functions_aligned_count / function_count * 100
```
jonschz, line 342: Maybe `alignment_percentage`? I think `accuracy` is fine for the other variables since the code can be matched to a fractional degree, but alignment is binary

jonschz (approved):
Looks great!

reccmp/tools/asmcmp.py
disinvite: No, that's correct. I'll expand the comment.

reccmp/tools/asmcmp.py
jonschz: Yes 👍

reccmp/tools/asmcmp.py
jonschz: Yes 👍

PR #61, opened 2025-01-05
Refactor: Replace legacy type annotations

jonschz:
Python 3.8 is now EOL and Python 3.9 supports type annotations on generics. `Optional[]` and the capital `List[]`, `Dict[]`, etc. are now gone.

### DIFF
--- a/reccmp/tools/datacmp.py
+++ b/reccmp/tools/datacmp.py
@@ -123,7 +123,7 @@ class ComparisonItem(NamedTuple):
 
 def create_comparison_item(
     var: ReccmpMatch,
-    compared: list[ComparedOffset | None] = None,
+    compared: list[ComparedOffset] | None = None,
     error: str | None = None,
     raw_only: bool = False,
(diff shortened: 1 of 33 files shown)


reccmp/tools/datacmp.py
```diff
@@ -123,8 +123,8 @@ def result(self) -> CompareResult:
 
 def create_comparison_item(
     var: ReccmpMatch,
-    compared: Optional[List[ComparedOffset]] = None,
-    error: Optional[str] = None,
+    compared: list[ComparedOffset | None] = None,
```
disinvite, line 126: This should be `compared: list[ComparedOffset] | None = None`. The arg is `None` if we are reporting an error.

That's only there because the mutable list can't be a default value for `compared`.

jonschz: Good catch! I noticed this one earlier today but didn't have time to commit a fix. I think I made a mistake in my regex

disinvite (approved):

PR #62, opened 2025-01-05
Use regex lookaround in asm sanitize

disinvite:
The regular expressions used in asm sanitize looked like this:

```python
ptr_replace_regex = re.compile(r"\[(0x[0-9a-f]+)\]")
```

i.e. capturing the surrounding square brackets as part of the total match.

[`re.sub`](https://docs.python.org/3/library/re.html#re.sub) will replace the entire match with a string or the return value of a function. We use the function, but `match.group(0)` included more than just the hex value, so we used a hacky string replacement to get it working. The correct way is to use lookaround expressions to match with the surrounding characters but to not include them in `group(0)`.

I also removed an edge case that makes no sense. The expression that detects immediate value operands would allow a replacement of the first operand, which is the destination in intel syntax. But an immediate value can never be the destination. I think this was there to cover replacement for single operand instructions, but we already match the ones we want earlier in the function: `call`, `push`, and jumps.

### DIFF
--- a/reccmp/isledecomp/compare/asm/parse.py
+++ b/reccmp/isledecomp/compare/asm/parse.py
@@ -15,13 +15,12 @@ from .instgen import InstructGen, SectionType
 from .replacement import AddrTestProtocol, NameReplacementProtocol
 from .types import DisasmLiteInst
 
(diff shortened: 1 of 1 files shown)


jonschz (approved):
looks good!

PR #64, opened 2025-01-05
Fix type issues, part 2

jonschz:
This fixes all remaining basic `mypy` issues outside of the Ghidra scripts (which I will fix in a separate PR). I also set up a basic CI - improvements are possible (especially the name "Format" feels a bit incorrect now). ~~Leaving as draft until I've gone over it once again.~~

### DIFF
--- a/reccmp/isledecomp/formats/mz.py
+++ b/reccmp/isledecomp/formats/mz.py
@@ -44,7 +44,39 @@ class ImageDosHeader:
             raise ValueError
         struct_fmt = "<2s29HI"
         struct_size = struct.calcsize(struct_fmt)
-        items = struct.unpack_from(struct_fmt, data, offset)
+        items: tuple[
+            bytes,
+            int,
+            int,
+            int,
+            int,
+            int,
+            int,
+            int,
+            int,
+            int,
+            int,
+            int,
+            int,
+            int,
+            int,
+            int,
+            int,
+            int,
+            int,
+            int,
+            int,
+            int,
+            int,
+            int,
+            int,
+            int,
+            int,
+            int,
+            int,
+            int,
+            int,
+        ] = struct.unpack_from(struct_fmt, data, offset)
         result = cls(
             *items[:14],
             items[14:18],
--- a/reccmp/isledecomp/formats/mz.py
+++ b/reccmp/isledecomp/formats/mz.py
@@ -44,39 +44,11 @@ class ImageDosHeader:
             raise ValueError
         struct_fmt = "<2s29HI"
         struct_size = struct.calcsize(struct_fmt)
-        items: tuple[
-            bytes,
-            int,
-            int,
-            int,
-            int,
-            int,
-            int,
-            int,
-            int,
-            int,
-            int,
-            int,
-            int,
-            int,
-            int,
-            int,
-            int,
-            int,
-            int,
-            int,
-            int,
-            int,
-            int,
-            int,
-            int,
-            int,
-            int,
-            int,
-            int,
-            int,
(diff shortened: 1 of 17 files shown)


reccmp/tools/asmcmp.py
```diff
@@ -6,8 +6,9 @@
 import logging
 import os
 from datetime import datetime
+from pathlib import Path
 
-from pystache import Renderer
+from pystache import Renderer  # type: ignore[import-untyped]
```
jonschz, line 11: I haven't found any stubs here

reccmp/tools/project.py
```diff
@@ -102,7 +102,7 @@ def main():
     if args.subcommand == ProjectSubcommand.CREATE:
         try:
             # pylint: disable=unused-argument
-            project = create_project(
+            project: RecCmpProject | None = create_project(
```
disinvite, line 105: Since we aren't using the `project` variable for anything here, I'd say just remove it here and then move the type hint to the second use in the "detect" section.

reccmp/isledecomp/formats/mz.py
```diff
@@ -44,7 +44,39 @@ def from_memory(cls, data: bytes, offset: int) -> tuple["ImageDosHeader", int]:
             raise ValueError
         struct_fmt = "<2s29HI"
         struct_size = struct.calcsize(struct_fmt)
-        items = struct.unpack_from(struct_fmt, data, offset)
+        items: tuple[
```
disinvite, line 47: Any way around this? Aside from disabling `black` here so it all fits on one line.

jonschz: Not as far as I know. https://github.com/python/typing/issues/786 would be helpful here if it were supported.

Another option would be to declare c structs (see e.g. https://stackoverflow.com/q/1825715/10666668), but I haven't found a way to specify the endianness - I suppose this will cause issues if the decompilation subject and the host have different endianness.

The third option would be to just ignore the type errors here, but I'd prefer not to do that. Will try disabling `black` first.

disinvite: Could we move the type hints to their own file and import them here? That way they're out of sight until that typing change comes through. Then if you were going to disable `black` with `# fmt: off` you could do it there and not near the actual code.

jonschz: We could, in principle. However, I'd like to keep them in the same place as the format strings of the `struct.unpack()` calls since that is what they are based on.

The protection we get from these type hints is rather limited anyway, since the programmer must check manually that they match the struct format string. Personally, I'd prefer disabling the type checking for the one problematic line over  moving only the type annotation (without the format string) into a different file. And even moving both together feels a bit off to me - after all, `mz.py` and `pe.py` don't have much of a purpose besides providing a Python interface for a binary format, so that's the place where I would expect everything related to that format to go.

Long story short, here's my personal ordered preference:
1. Keep it as it is now (with locally disabled typecheck)
2. Drop the endless `int` tuple types altogether and `# type: ignore` the constructor calls
3. Move the types (and, potentially, the format strings) into separate files

disinvite: My main issue was the detriment to readability (and style) with the 30-line thing, but I think we're good now with it all on one line. We may end up wanting to move some of this `struct` stuff into its own file and then disable type checking there if it makes sense. There are a lot of enums and structs in these format modules now and I think it might be helpful to manage the binary-to-python conversion separately from the higher level stuff. But that's something for later.

disinvite (approved):

Issue #32, opened 2024-12-12
Import size of all orig functions from file


disinvite:
Most functions in LEGO1 are aligned to each 16 bytes. (i.e. the last digit in the hex address is 0). If the function's size is not evenly divisible, MSVC will pad it out with `0xcc` bytes. This is the `int3` instruction. The purpose is to break into the debugger if something caused the IP to jump outside of the expected code region.

We don't always know the exact size of a function in orig. We use the recomp function size to start, then take the minimum of that and the distance to the next orig function. If we don't have this known stopping point, we might read instructions from the following function. `instgen.py` collects the disassembled instructions from `capstone` and will stop if it sees an `int3` instruction.

You probably see where this is going. BETA10 `LegoROI::Read` has an intentional `int3` instruction (by way of the `_RPT1` macro) at `0x1018a0de`. So this is not a perfect method of approximating function size.

It's more difficult than it seems to accurately guess the range of each function. Even Ghidra doesn't do this perfectly all the time. So I would suggest that we add some way to read a simple file and a simple format (ala [python date format codes](https://docs.python.org/3/library/datetime.html#format-codes)) and insert into the database.

jonschz:
The idea sounds good in general. Just to be on the same page, what is your desired workflow here? Just collecting some ideas, feel free to edit this comment.

**Mandatory or optional end locations for every function?**

Advantages of mandatory:
- Likely more stable diff for incomplete functions

Disadvantages of mandatory:
- Higher barrier of entry since the end of each function must be found/generated first
- Ghidra + scripts will probably be necessary in order to start a new project
- Some maintenance work required for functions whose end is misdetected by Ghidra

**Where to get the initial offsets?**
- Will probably have to be a Ghidra script
- Especially important in the mandatory case

**Where to store the information?**
- Extra file?
  - Likely easier for a Ghidra workflow
  - We lose some of the benefits of having all annotations in code
- In code?
  - Likely easier to maintain (e.g. `// FUNCTION: LEGO1 0x10001000-0x1000abcd`)
  - Harder to set up initially (but a script-based refactoring to add them in via Ghidra sounds doable)
  - Likely more convenient for a workflow with optional end locations

madebr:
I'd make it optional, perhaps using `reccmp-project.yml` to make it mandatory. We want to keep it easy to start a re project.

Can this arguments also be made for data?
Does it also need to check the section name?
[...]

PR #65, opened 2025-01-07
Use string enum `EntityType`

### LINKED ISSUE #51: Reconsider `SymbolType` enum
We currently have this:
```python
class SymbolType(IntEnum):
    """Broadly tells us what kind of comparison is required for this symbol."""

    FUNCTION = 1
    DATA = 2
    POINTER = 3
    STRING = 4
    VTABLE = 5
    FLOAT = 6
```
We set this field for an entity based on assumptions from the PDB. If we read a match marker like `// VTABLE` we will set or replace this value based on the type of comparison that is requested.

We use the value to decide how to compare each entity. We also use it in the `match_name` function that presents a string describing the entity. This is used in asm sanitize and takes the form of: `{name | string text | float number} ({symbol type name})`

The problems with this are:
- Cannot insert the enum directly into the database, so it must be converted to a primitive type. This was solved this by extending `IntEnum` instead of just `Enum`.
- To write a custom query to (for example) "select all strings" the user has to know this internal value or use the enum as a parameter.
- Similarly, when loading data from a text file, there isn't a way to declare that an entity is a particular `SymbolType` without knowing the internal value (See #32)
- For proper type hygiene we should expect `SymbolType` anywhere the "type" field is needed. One reason for not doing this is that enums are slow. For what they are, there's more overhead than you would expect. (Again, we have sort of dodged this one by using `IntEnum`)
- On the topic of performance, to get a string representation of the `SymbolType` in `match_name()` it is much faster to use a `dict` than the "correct" way of `SymbolType(value).name`.

disinvite:
**Note:** `StrEnum` requires python 3.11 or later, so this effectively becomes our minimum version if this is merged. CI actions here and in isledecomp/isle use 3.12.

---

Resolves #51.

As of #45, items in the database (i.e. the components of the binaries) are "entities". This renames `SymbolType` to `EntityType` to match. It is a [`StrEnum`](https://docs.python.org/3/library/enum.html#enum.StrEnum) to support setting the value via a string (see #32). We use `enum.auto()` to establish the values. The expected behavior is that the raw value is a lower case string version of the enum constant.

There are three other changes. The main one is the `entity_type` property method of `ReccmpEntity` (formerly `compare_type`). It now always returns an `EntityType`. If none is set, return `EntityType.UNKNOWN`. If the raw value in the database is not part of the enum, return `EntityType.ERRTYPE`. This seems appropriate because this method is a helper anyway, and our current code assumes that an enum is returned. You can get the raw value with the `.get()` method.

`match_name()` from `ReccmpEntity` now takes advantage of the fact that `compare_type` is always an enum and so we use its string value directly.

Similarly, `match_type_abbreviation()` from `roadmap` uses the first three characters of the enum value.

### DIFF
--- a/.github/workflows/format.yml
+++ b/.github/workflows/format.yml
@@ -10,6 +10,10 @@ jobs:
     steps:
     - uses: actions/checkout@v4
 
+    - uses: actions/setup-python@v5
+      with:
+        python-version: '3.12'
+
     - name: Install python libraries
       shell: bash
       run: |
--- a/reccmp/isledecomp/compare/asm/replacement.py
+++ b/reccmp/isledecomp/compare/asm/replacement.py
@@ -1,7 +1,7 @@
 from functools import cache
 from typing import Callable, Protocol
 from reccmp.isledecomp.compare.db import ReccmpEntity
-from reccmp.isledecomp.types import SymbolType
+from reccmp.isledecomp.types import EntityType
 
 
 class AddrTestProtocol(Protocol):
@@ -29,7 +29,7 @@ def create_name_lookup(
             return m.match_name()
 
         offset = addr - getattr(m, addr_attribute)
-        if m.compare_type != SymbolType.DATA or offset >= m.size:
+        if m.entity_type != EntityType.DATA or offset >= m.size:
             return None
 
         return m.offset_name(offset)
--- a/reccmp/isledecomp/compare/core.py
+++ b/reccmp/isledecomp/compare/core.py
@@ -12,7 +12,7 @@ from reccmp.isledecomp.cvdump.demangler import demangle_string_const
 from reccmp.isledecomp.cvdump import Cvdump, CvdumpAnalysis
 from reccmp.isledecomp.parser import DecompCodebase
 from reccmp.isledecomp.dir import walk_source_dir
-from reccmp.isledecomp.types import SymbolType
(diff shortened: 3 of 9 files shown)


jonschz:
Disregarding the CI, doesn't this also break Python < 3.11 support for running locally? I don't have any personal stakes in supporting older, but still maintained Python versions, but I'm not sure we have a team agreement on which version are supposed to be supported.

disinvite:
Maybe the python version question should be its own PR rather than jammed into this one. If we're saying 3.11 is the minimum then we should probably test on that version (along with newer ones).

disinvite:
I might break this up into a few mutually exclusive PRs for simplicity. We don't absolutely need to use strings to allow loading from file and support `type=vtable`, so maybe hold off on `StrEnum` for now.

The items in flight are:
- Rename `SymbolType` to `EntityType` and `compare_type` property to `entity_type`. May as well rename the SQL table to "entities" while we're at it.
- `entity_type` property always returns enum instance.
- Run CI tests against multiple python versions.
- Establish minimum Python version. If we will use `StrEnum`, that's 3.11. If not, it's 3.10.
- "Format" task should use whatever minimum python version. The other tasks already use 3.12.

disinvite:
I'm going to close this for now. Most of the code here has been added in other PRs. The two issues are not solved yet but we can come back to those when it's time to figure out an import format.

PR #67, opened 2025-01-09
Rename `SymbolType` to `EntityType`

disinvite:
(Just the rename from #65.)

As of https://github.com/isledecomp/reccmp/pull/45, items in the database (i.e. the components of the binaries) are "entities". Therefore:

- `SymbolType` enum is now `EntityType`.
- `compare_type` property method of `ReccmpEntity` is now `entity_type`.
- `CompareDb` is now `EntityDb`.
- SQL table `symbols` is now `entities`.

### DIFF
--- a/reccmp/isledecomp/compare/core.py
+++ b/reccmp/isledecomp/compare/core.py
@@ -12,11 +12,11 @@ from reccmp.isledecomp.cvdump.demangler import demangle_string_const
 from reccmp.isledecomp.cvdump import Cvdump, CvdumpAnalysis
 from reccmp.isledecomp.parser import DecompCodebase
 from reccmp.isledecomp.dir import walk_source_dir
-from reccmp.isledecomp.types import SymbolType
+from reccmp.isledecomp.types import EntityType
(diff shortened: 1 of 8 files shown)


reccmp/isledecomp/compare/core.py
```diff
@@ -165,7 +165,7 @@ def _load_cvdump(self):
                     - sym.offset
                 )
 
-            if sym.node_type == SymbolType.STRING:
+            if sym.node_type == EntityType.STRING:
```
jonschz, line 168: Do we also want to rename the variable `sym` and/or the property `node_type`, and maybe `CvdumpNode` to `CvdumpEntity`? Just an idea, no strong opinion from my side

jonschz (approved):

reccmp/isledecomp/compare/core.py
disinvite: I'm going to leave those for now. Ultimately, what I think we should do with the cvdump loader is to have it insert things directly using our API rather than have this file extract the items from the `CvdumpAnalysis` structure. But that will take more than one PR to work out.

PR #68, opened 2025-01-10
Test on Python 3.10 through 3.13

### LINKED ISSUE #66: Set a minimum supported Python version
Which Python version should be the oldest supported by `reccmp`?

3.10 or 3.11 seem like the most logical options: https://devguide.python.org/versions/

---

**Background:**

For our CI task runners, if a specific version is not installed with `actions/setup-python@v5`, the default Python version for `ubuntu-latest` is 3.10.12. For `windows-latest` it is 3.9.13.
https://github.com/actions/runner-images/blob/main/images/ubuntu/Ubuntu2204-Readme.md
https://github.com/actions/runner-images/blob/main/images/windows/Windows2022-Readme.md

`ubuntu-latest` will soon (Jan 17) update to 24.04, with a default python version of 3.12. https://github.com/actions/runner-images/issues/10636

All the CI scripts on the `isle` repo that use Python specify Python 3.12. Our "format" script (black, pylint, mypy) is the only one to not specify a version. It runs `ubuntu-latest` so the version is 3.10 until next Friday when it will start using 3.12.

The effective minimum version for `reccmp` is 3.10 because we use [union type syntax](https://peps.python.org/pep-0604/) and [match/case statements](https://peps.python.org/pep-0634/).

disinvite:
Resolves #66.

We now test on Windows and Ubuntu for Python versions 3.10 through the latest 3.13. (3.14 is not finalized yet.)

The "format" task now runs 3.12. This will happen next Friday (Jan 17) regardless because we use `ubuntu-latest` as the runner. (See https://github.com/actions/runner-images/issues/10636)

I changed some of the text for the tasks if anyone has feedback.

### DIFF
--- a/.github/workflows/format.yml
+++ b/.github/workflows/format.yml
@@ -10,6 +10,10 @@ jobs:
     steps:
     - uses: actions/checkout@v4
 
+    - uses: actions/setup-python@v5
+      with:
+        python-version: '3.12'
+
     - name: Install python libraries
       shell: bash
       run: |
--- a/.github/workflows/unittest.yml
+++ b/.github/workflows/unittest.yml
@@ -4,17 +4,18 @@ on: [push, pull_request]
 
 jobs:
   fetch-deps:
-    name: Download original binaries
+    name: Prepare sample binary files
     uses: ./.github/workflows/legobin.yml
 
-  pytest-win:
-    name: 'pytest ${{ matrix.platform.name }}'
+  test:
+    name: '${{ matrix.platform.name }} ${{ matrix.python-version }}'
     runs-on: ${{ matrix.platform.os }}
     needs: fetch-deps
 
     strategy:
       fail-fast: false
       matrix:
+        python-version: ["3.10", "3.11", "3.12", "3.13"]
         platform:
           - { name: 'Windows', os: 'windows-latest' }
           - { name: 'Linux',   os: 'ubuntu-latest' }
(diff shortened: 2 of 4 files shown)


madebr:
I'd merge all the workflows, and add `if: always()` to some. We'd have to maintain less workflows, and the CO2 footprint will be smaller.

disinvite:
Do you mean having one Windows and one Ubuntu runner each to test all the Python versions? How do I do that?

madebr:
> Do you mean having one Windows and one Ubuntu runner each to test all the Python versions? How do I do that?

I was thinking about merging `format.yml` into `unittest.yml`, but perhaps that adds extra work (so ignore my previous remark).

How about [adding `cache: pip` on `actions/setup-python`](https://github.com/actions/setup-python?tab=readme-ov-file#caching-packages-dependencies)?

madebr (approved):

jonschz (approved):
Maybe we also want to add the supported versions to `pyproject.toml` like so: https://packaging.python.org/en/latest/guides/writing-pyproject-toml/#classifiers ? Looks good overall!

disinvite:
Looks like the `black` rules changed according to the python version in `pyproject.toml`. I'll check that out.

disinvite:
Here's why: https://black.readthedocs.io/en/stable/usage_and_configuration/the_basics.html#t-target-version

> By default, Black will infer target versions from the project metadata in pyproject.toml, specifically the [project.requires-python] field. If this does not yield conclusive results, Black will use **per-file auto-detection.**

PR #69, opened 2025-01-11
Fix `mypy` compliance in Ghidra scripts

jonschz:
The basic type safety should now be complete.

### DIFF
--- a/reccmp/ghidra_scripts/import_functions_and_types_from_pdb.py
+++ b/reccmp/ghidra_scripts/import_functions_and_types_from_pdb.py
@@ -31,8 +31,7 @@ import traceback
 from typing import TYPE_CHECKING
 
 if TYPE_CHECKING:
-    import ghidra
-    from lego_util.headers import *  # pylint: disable=wildcard-import # these are just for headers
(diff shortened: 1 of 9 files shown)


reccmp/ghidra_scripts/import_functions_and_types_from_pdb.py
```diff
@@ -70,6 +104,13 @@ def setup_logging():
 # This script can be run both from Ghidra and as a standalone.
 # In the latter case, only the PDB parser will be used.
 setup_logging()
+find_and_add_venv_to_pythonpath()
+reload_module("reccmp.ghidra_scripts.lego_util.statistics")
+reload_module("reccmp.ghidra_scripts.lego_util.globals")
+from reccmp.ghidra_scripts.lego_util.globals import GLOBALS
+
+logging.root.setLevel(GLOBALS.loglevel)
```
jonschz, line 112: This had to be reordered because we can't import `GLOBALS` before setting up the venv now. On the other hand, `mypy` doesn't allow me to import `from lego_util.globals` because the package can be referenced by two different paths.

disinvite (approved):

PR #71, opened 2025-01-12
Add Ghidra import for globals (data)

jonschz:
As the title says. Can be quite convenient to make sure everything is covered - I intend to experiment in that direction at some point.

I also fixed an issue with the Ghidra import where it would act up when a type of the same name already existed (imported by Ghidra from elsewhere). This used to require manual intervention.


disinvite (approved):
Sorry for the delay! Looks good. You could maybe add a docstring on `do_with_error_handling` just to explain the need to remove the duplicated code, but it is sort of self-explanatory.

Issue #75, opened 2025-01-16
PE: Get regions by type


disinvite:
We need some kind of helper function for PE images that will return regions of the file based on their characteristics. (This would solve one of the shortcomings in #70)

The simplest example is situations where we want to scan all code bytes. Right now we only use `.text`, which is not even the only code section in `LEGO1.DLL` and `BETA10.DLL`. There is also no requirement that the section be called `.text`. (I have a Borland-compiled game with `CODE` and no `.text`)

PE sections have a characteristics field and we already have an `IntEnum` to parse them.

[This table on Microsoft's site](https://learn.microsoft.com/en-us/windows/win32/debug/pe-format#special-sections) shows common section names and their expected characteristics. (Scroll down past the bullet list.)

So... just look at the characteristics? This works for code because we can test for `IMAGE_SCN_MEM_EXECUTE` or `IMAGE_SCN_CNT_CODE`. Searching for const data is harder because almost all sections in that table have `IMAGE_SCN_CNT_INITIALIZED_DATA` and `IMAGE_SCN_MEM_READ` set.

We could refer to the data directory and exclude those regions, but there is no guarantee that the directory entry points at the beginning of a section. The directory can also point to a data structure inside a segment used for other things. For example, exports in `LEGO1` are in `.rdata`. There is no `.edata` section.

Do we refer to the "standard" section names as an exclusion list? For example, if I want to look at const data, get all sections with the "init_data" and "readable" flags set, but where `section.name not in {".debug", ".edata", ".idata", ".pdata", ...}`.

Is there a better way that I'm missing?

disinvite:
Duplicate of #72.

PR #78, opened 2025-01-22
New analysis module for `find_float_consts`

disinvite:
Moving the `find_float_consts` search into its own module. Same performance improvements from #70. (I'm working on the thunk piece separately.)

Instead of having all analysis on the binary jammed into `PEImage`, put the relevant stuff in its own module that calls functions on the image.

I think all the review comments are addressed from that previous PR, and we now have some rudimentary tests targeted to just this analysis function. We still have the `TODO` to use more sections than just `.text` and `.rdata` but we are better suited to add the new function when it is ready. (#75)

The module is just under `analysis` for now. Thoughts on x86-specific or PE-specific sub-modules?

As for the floating point instructions: this is hopefully now better explained in the code and comments. You can run this excerpt to see the candidate instructions:
```python
from capstone import Cs, CS_ARCH_X86, CS_MODE_32
disassembler = Cs(CS_ARCH_X86, CS_MODE_32)

deadbeef = b"\xef\xbe\xad\xde"
for opcode in range(0xd8, 0xe0):
    for ext in range(256):
        code = bytearray([opcode, ext, *deadbeef])
        inst = next(disassembler.disasm(code, 0), None)
        if inst and "deadbeef" in inst.op_str:
            print(f"{opcode:02x} {ext:02x}   {inst.mnemonic:10} {inst.op_str}")
    print()
```

That's all of them that operate on a single address. We can ignore anything that stores a value since those pointers are not constants. We also don't care about things that set or load status registers on the FPU. I ignored integer instructions like `fiadd` for now but we may want them later. We actually wound up targeting _fewer_ instructions than before.

### DIFF
--- a/reccmp/isledecomp/analysis/float_const.py
+++ b/reccmp/isledecomp/analysis/float_const.py
@@ -11,7 +11,6 @@ Then filter on pointers into read-only sections.
 """
 import re
 import struct
-from collections.abc import Buffer
 from typing import Iterator, NamedTuple
 from reccmp.isledecomp.formats import PEImage
 
@@ -63,10 +62,11 @@ class FloatInstruction(NamedTuple):
 
 
 def find_float_instructions_in_buffer(
-    buf: Buffer, base_addr: int = 0
+    buf: bytes, base_addr: int = 0
 ) -> Iterator[FloatInstruction]:
     """Search the given binary blob for floating-point instructions that reference a pointer.
     If the base addr is given, add it to the offset of the instruction to get an absolute address.
+    TODO: Uses `bytes` as the generic type for the Buffer protocol. See PEP 688 added in Python 3.12.
     """
     for match in FLOAT_INSTRUCTION_RE.finditer(buf):
         inst = match.group(1)
--- a/reccmp/isledecomp/analysis/float_const.py
+++ b/reccmp/isledecomp/analysis/float_const.py
@@ -1,0 +1,124 @@
+"""Analysis related to x86 floating point instructions.
+All floating point instructions use two byte opcodes. The first byte is in the range D8 to DF.
+The second indicates the operation and pointer or registers used.
+
+We are interested in floating point constants, so we want to exclude instructions that:
+- access the status register or environment (FLDCW, FLDENV)
+- store a value (FST, FSTP)
+- refer to integers (FI*)
+
+Then filter on pointers into read-only sections.
+"""
+import re
+import struct
+from typing import Iterator, NamedTuple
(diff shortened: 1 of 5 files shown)


reccmp/isledecomp/analysis/float_const.py
```diff
@@ -0,0 +1,124 @@
...
+    pointer: int
+
+
+def find_float_instructions_in_bytes(
+    raw: bytes, base_addr: int = 0
+) -> Iterator[FloatInstruction]:
+    """Search the given binary blob for floating-point instructions that reference a pointer.
+    If the base addr is given, add it to the offset of the instruction to get an absolute address.
+    """
+    for match in FLOAT_INSTRUCTION_RE.finditer(raw):
```
madebr, line 69: I'm not sure what python version is required, but I think you can run regular expressions on memoryviews.

disinvite: Thanks! Per [PEP 688](https://peps.python.org/pep-0688/) the common type for `bytes` and `memoryview` is `collections.abc.Buffer`, so I used that here.

disinvite: Ah. 3.12 only.

disinvite: To clarify: it works with both `bytes` and `memoryview`, we just can't use the preferred type hint until 3.12. For now it's just `bytes` which `mypy` accepts.

reccmp/isledecomp/analysis/float_const.py
```diff
@@ -0,0 +1,123 @@
...
+    instructions and check whether the pointer address refers to
+    read-only data."""
+
+    # Multiple instructions can refer to the same float.
+    # Return each float only once from this function.
+    seen = set()
+
+    # TODO: Should check all code and const data sections.
+    code_sections = (image.get_section_by_name(".text"),)
+    const_sections = (image.get_section_by_name(".rdata"),)
```
madebr, line 98: Does this check also test write-able data sections?
e.g. code that does;
```c
float g_Gravity = 9.8f;
void set_gravity(float g) { g_Gravity = g; }
```

disinvite: It would not identify `9.8f` if the value is in a writable section. If it were never modified (and in `.rdata`) then we would return it, but the correct behavior is to add the variable annotations first and then not replace `g_Gravity` with `EntityType.FLOAT`.

jonschz (approved):
The readability has improved greatly, thanks! I took a brief look and couldn't find any regressions.

PR #79, opened 2025-01-25
Truncate arrays if they are longer in recomp

disinvite:
`_match_array_elements` adds offset entities for each array variable. We need this if array entries are referenced by pointers in other variables because `datacmp` only checks the database for a pointer match. (e.g. `g_skBMap` and `g_skeletonKickPhases`)

This function assumes the arrays are the same size in orig as in recomp. This is not always true for `BETA10` and so the offsets we generate will clash with the ones from the following variable. We can detect when this occurs by looking at the next orig address. If recomp is larger, we add the extra offsets on the recomp side only and not link them to orig.

You can use `roadmap --verbose` to see what changed. I'll attach the `BETA10` diff here. No change to `LEGO1`.
[truncate-array-roadmap-beta10.txt](https://github.com/user-attachments/files/18547325/truncate-array-roadmap-beta10.txt)

IIRC the ghidra scripts do not reference these offsets and use the cvdump data directly to build out the types. Is that right @jonschz?

### DIFF
--- a/reccmp/isledecomp/compare/core.py
+++ b/reccmp/isledecomp/compare/core.py
@@ -25,6 +25,9 @@ from .diff import combined_diff, CombinedDiffOutput
 from .lines import LinesDb
 
 
+# pylint: disable=too-many-lines
+
+
 logger = logging.getLogger(__name__)
 
 
@@ -311,6 +314,19 @@ class Compare:
         dataset: dict[int, dict[str, str]] = {}
         orig_by_recomp: dict[int, int] = {}
 
+        # Helper function
+        def _add_match_in_array(
+            name: str, type_id: str, orig_addr: int, recomp_addr: int, max_orig: int
+        ):
+            # pylint: disable=unused-argument
+            # TODO: Previously used scalar_type_pointer(type_id) to set whether this is a pointer
+            if recomp_addr in dataset:
+                return
+
+            dataset[recomp_addr] = {"name": name}
+            if orig_addr < max_orig:
+                orig_by_recomp[recomp_addr] = orig_addr
+
         # Indexed by recomp addr. Need to preload this data because it is not stored alongside the db rows.
         cvdump_lookup = {x.addr: x for x in self.cvdump_analysis.nodes}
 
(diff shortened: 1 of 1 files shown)


reccmp/isledecomp/compare/core.py
```diff
@@ -335,29 +325,50 @@ def _add_match_in_array(
...
-                            )
+            if data_type["type"] != "LF_ARRAY":
+                continue
+
+            # Check whether another orig variable appears before the end of the array in recomp.
+            # If this happens we can still add all the recomp offsets, but do not attach the orig address
+            # where it would extend into the next variable.
+            upper_bound = match.orig_addr + match.size
+            if (next_orig := self._db.get_next_orig_addr(match.orig_addr)) is not None:
+                if next_orig < upper_bound:
```
jonschz, line 336: maybe use an `and` instead of two consecutive `if`s

reccmp/isledecomp/compare/core.py
```diff
@@ -335,29 +325,50 @@ def _add_match_in_array(
...
+
+                if array_element_type.members is None:
+                    # If array of scalars
+                    name = f"{match.name}{array_element.name}"
+                    orig_addr = orig_element_base_addr
+                    recomp_addr = recomp_element_base_addr
+
+                    dataset[recomp_addr] = {"name": name}
+                    if orig_addr < upper_bound:
+                        orig_by_recomp[recomp_addr] = orig_addr
```
jonschz, line 360: How about adding back the helper function and passing `upper_bound` as an additional parameter? I like the added comments what each case corresponds to.

jonschz (approved):
Looks good! You are correct, these array elements are not used by the Ghidra import. I originally added them because `datacmp` would raise errors on correct data.

disinvite:
Thanks! Let me know how it looks now. I had to add the dreaded `# pylint: disable=too-many-lines` to the file, but I have some upcoming changes planned that will shrink this back down.

jonschz:
looks good!

PR #84, opened 2025-02-01
New matching and event reporting module

disinvite:
In #81 we changed the database to support merging two entities during a match. The natural progression is to separate loading of entity data (i.e. orig annotations) and matching. This is the first step towards doing that.

The modularization now makes it possible to test the effect of each match function. This was not really feasible before because everything was enclosed in the `Compare` structure. As a result, we now have a boatload of tests that demonstrate the idiosyncrasies for each type of match -- particularly vtables and static variables.

Until now, if we failed to match an orig annotation to a recomp entity, the orig address would not exist in the database. With these entities now there, we can use the orig name during asm sanitize even if we failed to match. I think this is better because it exposes a diff that would be hidden by the `<OFFSET>` placeholder. (Also: your annotation could be wrong, so this will show where it creates problems.) The impact to `LEGO1` is minimal because we have just about everything mapped out, but there are two new diffs.[^1]

If an orig vtable annotation fails to match, its name will be just the class name. It would appear in the asm output as just `'Pizza (VTABLE)'` instead of `Pizza:'vftable' (VTABLE)`. We could mock up a name here if the vtable fails to match, but I left it alone for the moment. This happens because we had been overloading the "name" attribute in the database for a lot of different purposes. Now that we are free to use any attribute for matching, we should change this.

I also added a simple event reporting protocol in `event.py` and each of the match functions can call it to report a failed match. We had relied on the `logging` module for this which left no way to react to a failed match in code or store it for later. (We can now also test when a match should fail.) For now we just redirect the events back to the main logger from `core.py`. I changed the error text slightly in a few places but the information should be the same.

Lastly, there are two new SQL views to help keep the queries concise and a wrapper around the `COUNT` query. When matching we now always read both addresses in order. This will keep matching consistent for cases (like function name match) where there are multiple options. (i.e. it does not depend on database insertion order or filename order when reading annotations.)

[^1]: We don't properly grab the `g_skeletonKickPhases` variable from the PDB so it's not possible to match it right now. The orig annotation now sets it in the database so there now appears to be a diff in `LegoRaceCar::HandleSkeletonKicks`. The other diff is from an MSVC library function which might be a typo in the annotation.


reccmp/isledecomp/compare/db.py
jonschz, line 1: Maybe I am missing something, but it looks to me as if there are a lot of unused functions in here right now (like `match_variable()`. Is there a good reason to not delete them in this PR?

reccmp/isledecomp/compare/core.py
```diff
@@ -262,51 +273,82 @@ def orig_bin_checker(addr: int) -> bool:
...
+
+        # For now, just redirect match alerts to the logger.
+        report = create_logging_wrapper(logger)
+
+        # Now match
+        match_symbols(self._db, report)
+        match_functions(self._db, report)
+        match_vtables(self._db, report)
+        match_static_variables(self._db, report)
+        match_variables(self._db, report)
```
jonschz, line 322: I really like this new structure

reccmp/isledecomp/compare/core.py
```diff
@@ -262,51 +273,82 @@ def orig_bin_checker(addr: int) -> bool:
...
+                    string_correct = string.name == orig
+                except UnicodeDecodeError:
+                    string_correct = False
+
+                if not string_correct:
+                    logger.error(
+                        "Data at 0x%x does not match string %s",
+                        string.offset,
+                        repr(string.name),
+                    )
```
jonschz, line 340: Shouldn't this be a report? Maybe it also makes sense to go over all remaining log calls and see which of those should be reports.

reccmp/isledecomp/compare/db.py
```diff
@@ -16,6 +16,16 @@
...
+
+    CREATE VIEW orig_unmatched (orig_addr, kvstore) AS
+        SELECT orig_addr, kvstore FROM entities
+        WHERE orig_addr is not null and recomp_addr is null
+        ORDER by orig_addr;
+
+    CREATE VIEW recomp_unmatched (recomp_addr, kvstore) AS
+        SELECT recomp_addr, kvstore FROM entities
+        WHERE recomp_addr is not null and orig_addr is null
+        ORDER by recomp_addr;
```
jonschz, line 28: I assume you had good reasons to directly add SQL to `match_msvc.py` instead of writing abstractions for all the SQL queries. If you don't intend to add abstractions, how about adding a comment here that these views are used in other files? At first glance, `db.py` may look like it is self-contained (and I think it was before).

reccmp/isledecomp/compare/match_msvc.py
```diff
@@ -0,0 +1,311 @@
...
+    def pop(self, key: str) -> int:
+        value = self._dict[key].pop(0)
+        if len(self._dict[key]) == 0:
+            del self._dict[key]
+
+        return value
+
+
+def match_symbols(db: EntityDb, report: ReccmpReportProtocol = reccmp_report_nop):
+    """Match all entities using the symbol attribute. We expect this value to be unique."""
```
jonschz, line 36: Maybe `Match all entities that have the ``symbol`` attribute set.`?

reccmp/isledecomp/compare/match_msvc.py
```diff
@@ -0,0 +1,311 @@
...
+                        msg=f"Matched 0x{orig_addr:x} using non-unique symbol '{symbol}'",
+                    )
+
+                batch.match(orig_addr, recomp_addr)
+
+            else:
+                report(
+                    ReccmpEvent.NO_MATCH,
+                    orig_addr,
+                    msg=f"Failed to match function at 0x{orig_addr:x} with symbol '{symbol}'",
```
jonschz, line 71: This probably isn't code you changed, but how can we be sure that we have a function here? And if we are sure, why don't we call this a function in the report above as well?

reccmp/isledecomp/compare/match_msvc.py
```diff
@@ -0,0 +1,311 @@
...
+
+    name_index = EntityIndex()
+
+    # TODO: We allow a match if entity_type is null.
+    # This can be removed if we can more confidently declare a symbol is a function
+    # when adding from the PDB.
+    for recomp_addr, name, symbol in db.sql.execute(
+        """SELECT recomp_addr, json_extract(kvstore, '$.name') as name, json_extract(kvstore, '$.symbol')
+        from recomp_unmatched where name is not null
+        and (json_extract(kvstore, '$.type') = ? or json_extract(kvstore, '$.type') is null)""",
```
jonschz, line 87: I am not sure about the precedence of `and` and `or` in SQL. How about adding parentheses to make this obvious?

reccmp/isledecomp/compare/match_msvc.py
```diff
@@ -0,0 +1,311 @@
...
+        # See also: warning C4786.
+        name = name[:255]
+        name_index.add(name, recomp_addr)
+
+        # Get the symbol for the error message later.
+        if symbol is not None:
+            recomp_symbols[recomp_addr] = symbol
+
+    # Report if the name used in the match is not unique.
+    # If the name list contained multiple addreses at the start,
```
jonschz, line 101: typo `addresses`

reccmp/isledecomp/compare/match_msvc.py
```diff
@@ -0,0 +1,311 @@
...
+
+    def pop(self, key: str) -> int:
+        value = self._dict[key].pop(0)
+        if len(self._dict[key]) == 0:
+            del self._dict[key]
+
+        return value
+
+
+def match_symbols(db: EntityDb, report: ReccmpReportProtocol = reccmp_report_nop):
```
jonschz, line 35: I am a bit confused about the dependencies here (this was likely the case before your changes):
* Do both `match_functions` and `match_static_variables` depend on this running first? If so, I'd document that in the respective docstrings.
* If not: What kind of symbols would `match_symbols` be good for, if it isn't among the previous two?

tests/test_match_msvc.py
jonschz, line 1: Very nice!

jonschz (commented):
Giving you a first batch of comments from what I noticed when reading the code. Will try and play around with it, then give some more feedback. Looks very promising! I especially like the improved testability and the report abstraction.

jonschz:
I tried a bit and failed to observe the diffs you mentioned. I see the same issue with `g_skeletonKickPhases` on the current `master` commit. Weirdly, `HandleSkeletonKicks` still shows a 100% match.

reccmp/isledecomp/compare/db.py
disinvite: No reason other than to keep this PR smaller. Let's take them out.

reccmp/isledecomp/compare/match_msvc.py
disinvite: This (reporting failed symbol match) is new and I just forgot to change the text copied from `match_functions`.

reccmp/isledecomp/compare/match_msvc.py
```diff
@@ -0,0 +1,311 @@
...
+    """Match all entities using the symbol attribute. We expect this value to be unique."""
+
+    symbol_index = EntityIndex()
+
+    for recomp_addr, symbol in db.sql.execute(
+        """SELECT recomp_addr, json_extract(kvstore, '$.symbol') as symbol
+        from recomp_unmatched where symbol is not null"""
+    ):
+        # Max symbol length in MSVC is 255 chars. See also: Warning C4786.
+        symbol_index.add(symbol[:255], recomp_addr)
```
madebr, line 45: That's correct for MSVC4 toolchain, but not for current MSVC.

disinvite: How about a parameter that decides whether to truncate? We don't have any way to enable or disable it (or detect the MSVC version) but these could be added later to the yml.

reccmp/isledecomp/compare/match_msvc.py
disinvite: The reason to match using the symbol is that it should be unique across the entire program. We don't need to check the type or any other attributes. Right now, the only place we expect a symbol is in a `// FUNCTION` or similar annotation, but we could add something like this:

```
// SYMBOL: TEST 0x1234
// ??_7PizzaMissionState@@6B@
// aka PizzaMissionState::`vftable`
```

It's probably most useful for this to run first, but `match_functions` doesn't depend on it. `match_static_variables` depends on having an orig entity (the function with the variable) with a symbol. You could match using either this or `match_functions`, or not at all if your annotation on the function uses the symbol.

reccmp/isledecomp/compare/match_msvc.py
disinvite: There are parens on that line around the check for `type is null or type = X`. Or do you mean somewhere else?

reccmp/isledecomp/compare/match_msvc.py
madebr: I was also thinking about making it configurable.
Or, and this I don't know for sure: perhaps it only truncates in the old-style PDB file format.
So we can make it dependent on the file format.

reccmp/isledecomp/compare/db.py
disinvite: Those views are just for convenience so we can exclude this boilerplate from all the queries:
```sql
where defined_addr is not null
and other_addr is null
order by defined_addr
```

We still have some abstraction: queries are used only in `db.py` and each of the matching functions. The DB _was_ self contained until now, but only because the matching functions were there and not in their own module. I think it makes sense to split those up so we can match new entity types and things specific to other compilers and not have the DB class get unmanageable.

Using queries directly in the match functions was just a way to get this new idea moving. I don't think single use-case functions like `get_unmatched_orig_function_names` are obviously better than a SQL query. Less-specific utility functions are slower because you're doing less work at the sqlite level. The tradeoff is that more queries flying around locks us into this schema, and if we ever move off of the JSON string then it's just more stuff to rewrite. Something to revisit later.

reccmp/isledecomp/compare/core.py
```diff
@@ -262,51 +273,82 @@ def orig_bin_checker(addr: int) -> bool:
...
+                recomp_addr = self._lines_db.search_line(
+                    fun.filename, fun.line_number, fun.end_line
+                )
+                if recomp_addr is not None:
+                    batch.match(fun.offset, recomp_addr)
+                    batch.set_recomp(
+                        recomp_addr, type=EntityType.FUNCTION, stub=fun.should_skip()
+                    )
+
+        with self._db.batch() as batch:
```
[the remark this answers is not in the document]
disinvite: Yes, I had these in different batches at one point and using `insert_orig` to match the previous code better. I'll switch this around tomorrow.

The reason to use separate batches is so that staging data with `insert_orig` would succeed once (in the first batch) and then not change the data in subsequent batches. If you keep calling `insert_orig` on the same address in the same batch, we modify the pending changes. This is by design so you can add attributes in stages (or only if certain conditions are met) as we do here.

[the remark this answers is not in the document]
disinvite: We _could_ do it all in one batch _if_ the annotations read from the `DecompCodebase` were guaranteed to have a unique orig address. I don't remember if that's true or not. We detect if you repeat the same addr in two different annotations (in the linter) but I don't think we remove the dupes here.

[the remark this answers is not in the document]
jonschz: That is a valid reason to dp multiple batches. I'd just add that to the code as a comment since the pattern was not obvious to me.

reccmp/isledecomp/compare/match_msvc.py
jonschz: Sorry, I've completely missed that. My bad

reccmp/isledecomp/compare/match_msvc.py
jonschz: That's what I expected. I'd add `"Depends on match_symbols to run first" to the docstrings of all the other functions (where applicable).

reccmp/isledecomp/compare/core.py
```diff
@@ -262,51 +273,82 @@ def orig_bin_checker(addr: int) -> bool:
...
+            for fun in codebase.iter_line_functions():
+                assert fun.filename is not None
+                recomp_addr = self._lines_db.search_line(
+                    fun.filename, fun.line_number, fun.end_line
+                )
+                if recomp_addr is not None:
+                    batch.match(fun.offset, recomp_addr)
+                    batch.set_recomp(
+                        recomp_addr, type=EntityType.FUNCTION, stub=fun.should_skip()
+                    )
```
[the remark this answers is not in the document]
disinvite: It would probably be better to have `set_orig` here, as with all the other imports from annotations that follow. I don't remember why I put `set_recomp` here. The final result to the database will be the same unless this match fails.

disinvite:
Today's changes:

- Added `truncate` param for symbol and function name match. This is just a bool for now. I don't know if there would be a need to truncate to a specific character length depending on the compiler. Defaults to `True` but we can control this later with a config option or by autodetecting when it is needed (PDB version).

- Small change to `DecompCodebase` to eliminate annotations that reuse an address in the current module. This combined with the new batch code gets the same behavior that we had before: only the first occurrence of the address is used/matched, and any duplicates are dropped. New error message to alert when this happens.

- With the above change: we can now load all annotations in a single batch because none of them will step on each other. I kept `set_orig` here because we want to override any existing information with the annotation data. Right now there isn't any data to replace (on the orig side) but that might change later.

- I moved the match functions out of `load_markers` and into the `__init__` function. We can move this to a helper function if that makes more sense, but the point is to separate loading from matching.

- Used `report` call in `load_markers` for the three error messages.

- Add comment for `match_static_variables`. It is most useful to call it after matching functions.

I think that covers all the outstanding items so please give it another look and let me know. Thanks!

madebr (approved):

jonschz (approved):
Looks good! We can resolve the few regressions afterwards. I really like the improvements away from `OFFSET1` as well.

Issue #85, opened 2025-02-03
Handle additional information in newer MSVC versions correctly


jonschz:
> Running reccmp-reccmp I get the following:
> ```
> AssertionError: Unhandled in union:     # members = 2,  field list type 0x22e2, SEALED, Size = 4        ,class name = <unnamed-tag>, unique name = .?AT<unnamed-tag>@@
> ```
> If I look in the console output, the first error I get is:
> ```
> [ERROR] Unmatched line in class:        Size = (LF_ULONG) 320996, class name = ViewportRenderState, unique name = .?AVViewportRenderState@@, UDT(0x00001056
> ```
> 
> I am running reccmp-reccmp on: the original binary, the recompiled binary (Release build), the Release build pdb file.
> 
> What should I look into to get this issue resolved?
> 
> I annotated my class function like this: 
> The game uses a single static instance of this class, so my recompilation does too.
> ```cpp
> class ViewportRenderState : ViewportRenderStateStruct {
>     public:
> 
>     //FUNCTION: MYEXE 0x401000
>     BOOL xyAreValid( uint x, uint y) {
>         ....
>     }
> }
> ``` 

 _Originally posted by @gynt in [#82](https://github.com/isledecomp/reccmp/issues/82#issuecomment-2630248084)_

jonschz:
This looks like additional information in the debug symbols that the code does not handle yet. Please run cvdump on your pdb file and upload it here (probably something like `.\reccmp\bin\cvdump.exe myfile.pdb > dump.txt`).

gynt:
Hi! Here is a simple class implementation in latest MSVC using cmake: https://github.com/gynt/reccmp-tests
It is 30 MB (: https://github.com/gynt/reccmp-tests/blob/ef062d76c33efc1564bd7d881bf4b50c0bac5bf7/dump.txt

jonschz:
@gynt Feel free to try the latest version on `master` and see whether things have improved for you. As mentioned on #86, `datacmp` is likely still broken.

gynt:
Works, using `--verbose` I now get a neat display of everything that doesn't match up (:

jonschz:
Some preliminary insights from poking around in the dump @gynt provided:
- The unknown types we have at the moment are:
  - `???(007C)`
  - `???(047C)`
  - `???(0x009b)` (possibly not a type)

### Hypothesis for `007C`

```
0x2020 : Length = 506, Leaf = 0x1203 LF_FIELDLIST
[...]

PR #86, opened 2025-02-05
Solve some of the `TYPES` parsing issues

disinvite:
As reported in #85. Not sure if this resolves it per se, because we now have the problem of unknown types. For example:

```
0x11f5 : Length = 22, Leaf = 0x1201 LF_ARGLIST argument count = 4
    list[0] = 0x11EA
    list[1] = ???(047C)
    list[2] = ???(047C)
    list[3] = 0x11F2
```

`0x047c` and `0x007c` are not listed in [cvinfo.h](https://github.com/microsoft/microsoft-pdb/blob/master/include/cvinfo.h) so that's probably why cvdump can't provide a type name like `T_RCHAR`.

It would help to know what these types are in the real program. Looks like they're related to `std` data structures.

We probably can't use `datacmp` effectively until we sort out the phantom types, but we can at least parse the `TYPES` section from #85 without an error.

There are still alerts for `LF_CLASS` attributes "CASTING" and "OVERLOADED ASSIGNMENT". I think we should parse these sections using a different regex that covers all the comma-delimited values. I have that working but it will go in a separate PR. It may bite us in the long run so we should discuss whether it's worth it to change our approach.

Other things addressed here:

- `LF_UNION` that doesn't end with a UDT attribute.
- Alert for "Return type = ???(047C)" in `LF_MFUNCTION`
- Alert for "Func attr" values other than "none"
- Alert for `LF_POINTER` type "R-value Reference"
- In `LF_CLASS`, make sure we can parse the UDT if the "unique name" field is present

### DIFF
--- a/reccmp/isledecomp/cvdump/types.py
+++ b/reccmp/isledecomp/cvdump/types.py
@@ -243,7 +243,7 @@ class CvdumpTypesParser:
     )
     LF_ENUM_UDT = re.compile(r"^\s*UDT\((?P<udt>0x\w+)\)$")
     LF_UNION_LINE = re.compile(
-        r"^.*field list type (?P<field_type>0x\w+),.*Size = (?P<size>\d+)\s*,class name = (?P<name>(?:[^,]|,\S)+)(?:,\s.*UDT\((?P<udt>0x\w+)\))?"
+        r"^.*field list type (?P<field_type>0x\w+),.*Size = (?P<size>\d+)\s*,class name = (?P<name>(?:[^,]|,\S)+)(?:, unique name = [^,]+)?(?:,\s.*UDT\((?P<udt>0x\w+)\))?$"
     )
 
     MODES_OF_INTEREST = {
--- a/reccmp/isledecomp/cvdump/types.py
+++ b/reccmp/isledecomp/cvdump/types.py
@@ -201,7 +201,7 @@ class CvdumpTypesParser:
 
     # LF_CLASS/LF_STRUCTURE name and other info
     CLASS_NAME_RE = re.compile(
-        r"^\s+Size = (?P<size>\d+), class name = (?P<name>(?:[^,]|,\S)+)(?:, UDT\((?P<udt>0x\w+)\))?"
+        r"^\s+Size = (?P<size>\d+), class name = (?P<name>(?:[^,]|,\S)+)(?:, unique name = [^,]+)?(?:, UDT\((?P<udt>0x\w+)\))?"
     )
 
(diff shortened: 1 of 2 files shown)


reccmp/isledecomp/cvdump/types.py
```diff
@@ -201,7 +201,7 @@ class CvdumpTypesParser:
 
     # LF_CLASS/LF_STRUCTURE name and other info
     CLASS_NAME_RE = re.compile(
-        r"^\s+Size = (?P<size>\d+), class name = (?P<name>(?:[^,]|,\S)+)(?:, UDT\((?P<udt>0x\w+)\))?"
+        r"^\s+Size = (?P<size>\d+), class name = (?P<name>(?:[^,]|,\S)+)(?:, unique name = [^,]+)?(?:, UDT\((?P<udt>0x\w+)\))?"
```
madebr, line 204: I reversed this in my microsoft-pdb fork:
https://github.com/madebr/microsoft-pdb/blob/1545c8af664f3a898356427eae86c610498d55e3/cvdump/type7.cpp#L672

jonschz (approved):
Thanks! This will be good enough as a hotfix. In the long run I'd like to reintroduce warnings for information the parser does not unterstand. "Ignore and don't crash" is fine for now.

reccmp/isledecomp/cvdump/types.py
```diff
@@ -245,7 +243,7 @@ class CvdumpTypesParser:
     )
     LF_ENUM_UDT = re.compile(r"^\s*UDT\((?P<udt>0x\w+)\)$")
     LF_UNION_LINE = re.compile(
-        r"^.*field list type (?P<field_type>0x\w+),.*Size = (?P<size>\d+)\s*,class name = (?P<name>(?:[^,]|,\S)+)(?:,\s.*UDT\((?P<udt>0x\w+)\))?$"
+        r"^.*field list type (?P<field_type>0x\w+),.*Size = (?P<size>\d+)\s*,class name = (?P<name>(?:[^,]|,\S)+)(?:,\s.*UDT\((?P<udt>0x\w+)\))?"
```
jonschz, line 246: Not quite sure about dropping the end match here. Maybe we want to capture everything until the end into an `unhandled` group?

disinvite: I put the `$` back in. We can just match `unique name` because that's the attribute that got us here.

jonschz:
@disinvite you're good to merge as far as I am concerned

PR #91, opened 2025-02-25
JSON report aggregate tool

### LINKED ISSUE #88: Support for diffing with multiple files
Off and on the team in isledecomp have discussed ways to cut through compiler noise and pinpoint regressions for a particular pull request. One approach is to add unnecessary (but valid) code to a random compilation unit and see how the accuracy percentages change. We could scale this up to X number of parallel builds, but then what do you do with all the data?

We have the `--json` and `--diff` options in `reccmp` to create a serialized accuracy report  and compare it against the current analysis. The typical use case is to create a report file, make some changes, then run `reccmp-reccmp` with `--diff` to see how you did. If we call these states A and B, `--json` creates a file for A and `--diff` will compare it against the in-memory data for B.

If we use `--json` to create a file for B, it should be simple to add a feature that compares the two files instead of doing the analysis to populate B.

We can then extend that to support as many files for B as we have, and either use the best percentage for each function or try to calculate some new average.

disinvite:
The main `reccmp-reccmp` script will save the report of the comparison run to a JSON file if you use the `--json` command line option. The report contains data for each orig. address with the then-current entity name, accuracy score, and flags to mark a function stub and effective match. The HTML display has the same data (also JSON format), but with unified diff output for each address included.

This update adds the `reccmp-denoise` script to combine multiple JSON reports into a single "noise reduced" report. (Resolves #88.) Provide the list of report files you want to aggregate in the `--samples` option, and the output file with `--output`.

Our strategy to create the aggregate is this: for each orig. address in any of the sample reports, use the highest accuracy score from any report. The hierarchy is:

- 100% match
- Any effective match
- Any accuracy score that is not an effective match
- Any stub

To clarify, what we mean by "noise" is: variations in the generated assembly in in functions not directly impacted by the code change under review.

This update also adds some rigidity for importing and exporting JSON data. This has been working fine for a while, but without any set schema. The only file format in the field is version 1, and this is marked in the JSON file to protect against possible future versions.

We now have the internal `ReccmpStatusReport` class and some `pydantic` schemas to serialize and deserialize the JSON report files. We could now restructure the data into a new schema version and this framework should make the conversion stable.

`reccmp-denoise` also allows you to diff two different report files. This is not possible with `reccmp-reccmp` because you can only diff the saved report against the current comparison run. If you specify `--diff` and `--samples` in the same run, we will diff the first file in the `--diff` list against the newly generated aggregate report.


foxtacles (approved):

jonschz:
Thank you for working on this. I probably won't be able to take a detailed look for a few days. Based on the description I'd say you're good to merge. Leaving it up to you if you want to wait for a full review or if you want to merge early.

disinvite:
Before merging should we rename this script from `reccmp-denoise` to `reccmp-report` or something along those lines? There are more things we can do with these files than just aggregating them. For example: if the JSON export has diffs included, we could generate the HTML report directly from that.

jonschz:
Maybe `reccmp-aggregate`? `-report` is also fine with me.

reccmp/isledecomp/compare/report.py
```diff
@@ -0,0 +1,185 @@
...
+
+
+def combine_reports(samples: list[ReccmpStatusReport]) -> ReccmpStatusReport:
+    """Combines the sample reports into a single report.
+    The current strategy is to use the entity with the highest
+    accuracy score from any report."""
+    assert len(samples) > 0
+
+    # TODO: hack
+    assert all(samples[0].filename == s.filename for s in samples)
```
jonschz, line 80: Still planning on addressing this before merging? If it's good enough, you can probably replace the comment by a non-TODO

reccmp/tools/denoise.py
```diff
@@ -0,0 +1,119 @@
...
+        saved_data = load_report_file(args.diff[0])
+
+        if agg_report is None:
+            if len(args.diff) > 1:
+                agg_report = load_report_file(args.diff[1])
+            else:
+                logger.error("Not enough files to diff!")
+                return 1
+
+        diff_json(saved_data, agg_report, show_both_addrs=False, is_plain=args.no_color)
```
madebr, line 113: Perhaps we should warn/fail when there are unused `--diff` arguments?

reccmp/tools/denoise.py
```diff
@@ -0,0 +1,119 @@
...
+        metavar="<file>",
+        help="Where to save the aggregate file.",
+    )
+    parser.add_argument(
+        "--samples",
+        type=Path,
+        metavar="<files>",
+        nargs="+",
+        help="Report files to aggregate.",
+    )
```
madebr, line 72: If neither of `--diff` and `--samples` is passed, then the program will exit silently.
I think this is an error condition.

disinvite: Thanks! See what you think now.

madebr (approved):

reccmp/tools/denoise.py
madebr: Looks good! I'd just have checked the `args` namespace.

PR #94, opened 2025-03-06
Store diffs in JSON report, HTML export for `reccmp-aggregate`

disinvite:
Two new features to help the entropy runs over at isledecomp:

The `--json` option for `reccmp-reccmp` now inserts diff information in the JSON report by default. This is the same data that has been included in the HTML reports. This doesn't seem to cause much of an impact other than increasing the report file size, which is why I enabled it by default. Deserializing the report with `--diff` is a tad slower (~100ms for me) so if you want the old behavior, add the parameter `--json-diet`. Adding diffs to the JSON is backward compatible with our current schema, so the version stays at 1.

The other component is the `--html` option for `reccmp-aggregate`. Now that we have the diffs in the JSON report, we can also generate an HTML report with the diff of the best match from each of the samples. In other words, for functions that still don't match, we can now see why and decide for ourselves if it's entropy not addressed by the effective match code.

Lastly, we had previously dropped the recomp addr when aggregating reports. In the rare occasion that it is the same in _all_ reports, we now keep it. Otherwise, we use the string "various".

### DIFF
--- a/tests/test_report.py
+++ b/tests/test_report.py
@@ -116,3 +116,22 @@ def test_aggregate_different_files():
 
     with pytest.raises(ReccmpReportSameSourceError):
         combine_reports([x, y])
+
+
+def test_aggregate_recomp_addr():
+    """We combine the entity data based on the orig addr because this will not change.
+    The recomp addr may vary a lot. If it is the same in all samples, use the value.
+    Otherwise use a placeholder value."""
+    x = create_report([("100", 0.8), ("200", 0.2)])
+    y = create_report([("100", 0.2), ("200", 0.8)])
+    # These recomp addrs match:
+    x.entities["100"].recomp_addr = "500"
+    y.entities["100"].recomp_addr = "500"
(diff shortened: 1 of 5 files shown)


tests/test_report.py
```diff
@@ -116,3 +116,22 @@ def test_aggregate_different_files():
...
+    # These recomp addrs match:
+    x.entities["100"].recomp_addr = "500"
+    y.entities["100"].recomp_addr = "500"
+    # Y report has no addr for this
+    x.entities["200"].recomp_addr = "600"
+
+    combined = combine_reports([x, y])
+    assert combined.entities["100"].recomp_addr == "500"
+    assert combined.entities["200"].recomp_addr != "600"
+    # TODO: string subject to change? better to leave as none?
```
jonschz, line 136: I think "various" is fine for now - if this ever causes an issue, we can always revert it.

jonschz (approved):
Looks good!

PR #97, opened 2025-03-24
reccmp-aggregate: support fromfile arguments

### LINKED ISSUE #96: `reccmp-aggregate` should support wildcard or glob
Got this error when trying to merge 1024 samples in github CI: `/d/a/isle/isle/.venv/Scripts/reccmp-aggregate: Argument list too long`

So we need a better way to load the samples if there are a lot of them. A glob expression would probably work best with the way the isle CI is set up. e.g. `reccmp-aggregate --samples build-entropy/**/CONFIG*.json`

disinvite:
Resolves #96. One-line fix thanks to @madebr.

We now also have some rudimentary tests on command-line parsing. We should add this for all the tools eventually. I thought I would start here because we have only a few arguments to deal with. Suggestions welcome on alternative ways to set this up.

### DIFF
--- a/reccmp/tools/aggregate.py
+++ b/reccmp/tools/aggregate.py
@@ -1,5 +1,6 @@
 #!/usr/bin/env python3
 
+import sys
 import argparse
 import logging
 from typing import Sequence
@@ -77,10 +78,11 @@ class TwoOrFewerArgsAction(argparse.Action):
         setattr(namespace, self.dest, values)
 
 
-def main():
+def parse_args(argv: list[str]) -> argparse.Namespace:
     parser = argparse.ArgumentParser(
         allow_abbrev=False,
         description="Aggregate saved accuracy reports.",
+        fromfile_prefix_chars="@",
     )
     parser.add_argument(
         "--diff",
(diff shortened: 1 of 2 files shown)


reccmp/tools/aggregate.py
```diff
@@ -127,6 +129,15 @@ def main():
...
         )
 
+    if args.diff and len(args.diff) == 1 and not args.samples:
+        parser.error("expected two report files to diff")
+
+    return args
+
+
+def main():
+    args = parse_args(sys.argv[1:])
```
madebr, line 139: I'd still pass all arguments, and split inside `parse_args`.
The [`prog` argument](https://docs.python.org/3/library/argparse.html#argparse.ArgumentParser) can then get the program name, which is used for `--help`.

reccmp/tools/aggregate.py
```diff
@@ -127,6 +129,15 @@ def main():
             "expected arguments for --output, --html, or --diff. (No output action specified)"
         )
 
+    if args.diff and len(args.diff) == 1 and not args.samples:
+        parser.error("expected two report files to diff")
```
madebr, line 133: I'd make the error more actionable
```suggestion
        parser.error("--diff expects two report files to diff")
```

disinvite:
Hopefully that's what you meant @madebr

madebr (approved):

PR #101, opened 2025-03-29
New PDB path conversion method

### LINKED ISSUE #95: Match filenames for a PDB built on another machine
Companion to #53. Solving this will result in the refactor requested in the other issue.

### How this works now

The PDB's `LINES` section associates modules and source filenames with pairs of line numbers and addresses. We use that to connect position-based `// FUNCTION` annotations with an address.

If you are running `reccmp` on Windows, we assume the PDB came from a build on your machine and all the (absolute) paths actually exist. Therefore it's easy to match them up.

If you are not on Windows, we still assume the PDB came from a build on your machine, but using Wine. We depend on `winepath` to translate the Windows paths to their POSIX equivalent.

To avoid calling `winepath` too often, we start by doing a POSIX-to-Windows conversion of the input source code path. This is either the path provided by the user at the command line or it is from the `reccmp-project.yml` file. Using that path fragment, we check each Windows path from the PDB to see if it is relative to the translated source path. If it is, we cut out the relative parts, add them to the original POSIX source path, then test to see if the file exists.

### The enhancement

Apply this same principle of using a path fragment mapping but without using `winepath` or assuming that the path must exist (on Windows). This would allow you to download the recomp binary and PDB from another location (i.e. the CI artifacts) and run the analysis on your own machine.

disinvite:
Resolves #95. It may also obviate #53 although that code has not changed.

The `LINES` section from the PDB connects code line numbers with addresses. The output from `cvdump` looks like this:

```
C:\work\lego-island\isle\LEGO1\mxdirectx\mxdirectdraw.cpp (None), 0001:0009BDB0-0009BE28, line/addr pairs = 7

 26 0009BDB0     27 0009BDB8     29 0009BDBE     31 0009BDC4
 33 0009BDCA     48 0009BE1F     49 0009BE25
```

We assume that the PDB came from a build on your machine and that this path exists. If you are not on Windows, we use `winepath` to resolve it.

The main limitation is that you cannot use a binary and PDB built somewhere else. We follow every path from the PDB, even if it would cause us to escape outside the directories provided to `reccmp`. There is a performance penalty on non-Windows platforms: if the PDB path is relative but cannot be found in the provided code directory, we need to call `winepath` to resolve it.[^1]

We now match the paths by comparing each path segment as a string. We use the provided code directory to create a list of candidate file paths. We convert everything to lower case to allow for differences in how the path appears in the PDB and on your host platform.

For example, if the path `/projects/isle/LEGO1/mxdirectx/mxdirectdraw.cpp` is the only path to a file called `mxdirectdraw.cpp`, we can match it easily. If we add `/projects/test/mxdirectdraw.cpp` to the list, we would still choose the earlier path because more of its segments match. If the PDB paths contains `.` or `..` then we stop comparing segments. We cannot resolve the path with 100% certainty because the user may have symlinks.

This isn't radically different to what we've already been doing. I haven't seen any issues with LEGO1 or BETA10 on Windows or Mac. For other projects, the success would depend on the uniqueness of the path segments. To protect against this, if we find two matches that are equally likely, the function returns `None`.

I could successfully run `reccmp` on my Mac against `LEGO1.DLL` and `LEGO1.PDB` built on my Windows box. It should be possible to run it this way in the Github CI on an Ubuntu runner, but I didn't get it working before submitting this. We still need Wine to run `cvdump` and I had issues with installing it.

I removed the sqlite db from the `lines.py` module. `Path` is a hashable type so we can use it as a dict key and it just works better. I want to eventually add _all_ code lines to this database, not just the addresses that start each function. We could then display the code lines with the assembly diff in the command line `reccmp-reccmp`.

[^1]:For example: code lines for templates from the MSVC include directory.

### DIFF
--- a/reccmp/isledecomp/compare/lines.py
+++ b/reccmp/isledecomp/compare/lines.py
@@ -1,52 +1,50 @@
 """Database used to match (filename, line_number) pairs
 between FUNCTION markers and PDB analysis."""
 
-import sqlite3
 import logging
 from functools import cache
-from pathlib import Path
-from reccmp.isledecomp.dir import PathResolver
-
-
-_SETUP_SQL = """
-    CREATE TABLE lineref (
-        path text not null,
-        filename text not null,
-        line int not null,
-        addr int not null
-    );
-    CREATE INDEX file_line ON lineref (filename, line);
-"""
+from pathlib import Path, PureWindowsPath
+from reccmp.isledecomp.dir import convert_foreign_path, walk_source_dir
 
 
 logger = logging.getLogger(__name__)
 
 
-@cache
-def my_samefile(path: str, source_path: str) -> bool:
-    return Path(path).samefile(source_path)
-
-
-@cache
-def my_basename_lower(path: str) -> str:
-    return Path(path).name.lower()
-
-
 class LinesDb:
     def __init__(self, code_dir) -> None:
-        self._db = sqlite3.connect(":memory:")
-        self._db.executescript(_SETUP_SQL)
-        self._path_resolver = PathResolver(code_dir)
+        # Code files of interest
+        # TODO: It may be better to provide the list of files instead of the directory
+        self._code_files = tuple(Path(p) for p in walk_source_dir(code_dir))
+        self._path_resolver = cache(convert_foreign_path)
+
+        # Set up memoized map of filenames to their paths
+        self._filenames: dict[str, list[Path]] = {}
+        for path in self._code_files:
+            self._filenames.setdefault(path.name.lower(), []).append(path)
+
+        # Local filename to list of (line_no, address) pairs
+        # This has to be a list instead of a dict because line numbers may be used twice.
+        # e.g. for the start and end of a loop.
+        self._map: dict[Path, list[tuple[int, int]]] = {}
 
     def add_line(self, path: str, line_no: int, addr: int):
-        """To be added from the LINES section of cvdump."""
-        sourcepath = self._path_resolver.resolve_cvdump(path)
-        filename = my_basename_lower(sourcepath)
+        """Connect the remote path to a line number and address pair."""
+        pdb_path = PureWindowsPath(path)
+        filename = pdb_path.name.lower()
 
-        self._db.execute(
-            "INSERT INTO lineref (path, filename, line, addr) VALUES (?,?,?,?)",
-            (sourcepath, filename, line_no, addr),
-        )
+        candidates = self._filenames.get(filename)
+        if candidates is None:
+            return
+
+        # Must convert to tuple (hashable type) so we can use functools.cache
+        sourcepath = self._path_resolver(pdb_path, tuple(candidates))
+        if sourcepath is None:
+            return
+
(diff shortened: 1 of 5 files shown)


madebr (approved):

reccmp/isledecomp/dir.py
```diff
@@ -12,6 +13,53 @@ def winepath_unix_to_win(path: str) -> str:
...
+    For best performance, you should narrow down the starting list to paths
+    that match the base filename."""
+    scored = [_count_matching_path_parts(remote_path, p) for p in local_paths]
+    scored.sort(reverse=True)
+
+    if len(scored) >= 2:
+        [(top_score, top_path), (next_score, _)] = scored[:2]
+        # Return if this is the best match above all others
+        if top_score > next_score:
+            return top_path
```
jonschz, line 52: Minor nitpick: One could add something like
```py
else:
    # The two best matches have the same score. We return None because we can't know which one is correct
    return None
```
to make the behaviour a bit more explicit

reccmp/isledecomp/compare/lines.py
```diff
@@ -1,52 +1,50 @@
...
 logger = logging.getLogger(__name__)
 
 
-@cache
-def my_samefile(path: str, source_path: str) -> bool:
-    return Path(path).samefile(source_path)
+class LinesDb:
+    def __init__(self, code_dir) -> None:
+        # Code files of interest
+        # TODO: It may be better to provide the list of files instead of the directory
```
jonschz, line 16: Could you elaborate on this TODO? Do you have any future plans / follow-ups / potential issues in mind?

reccmp/isledecomp/compare/lines.py
```diff
@@ -1,52 +1,50 @@
...
+        sourcepath = self._path_resolver(pdb_path, tuple(candidates))
+        if sourcepath is None:
+            return
 
-        self._db.execute(
-            "INSERT INTO lineref (path, filename, line, addr) VALUES (?,?,?,?)",
-            (sourcepath, filename, line_no, addr),
-        )
+        # mypy coersion. The function returns a PurePath because it is a superclass of Path.
+        assert isinstance(sourcepath, Path)
```
jonschz, line 45: Why can't you return `Path` from `convert_foreign_path`?

jonschz (approved):
Looks good overall! Haven't tried it yet, though

reccmp/isledecomp/compare/lines.py
disinvite: I could do it now. It's really just moving the call to `walk_source_dir` back up to the caller. The old path resolver used the root directory to avoid calls to `winepath` where possible, but we are now considering the entire path of each filename so the parent dir is less important. It would also be easier to run tests without depending on an actual directory.

reccmp/isledecomp/compare/lines.py
disinvite: I did that because we don't call `resolve()` or any of the methods that would rely on the filesystem inside of `convert_foreign_path`. As I look at this again, we don't call those functions in `LinesDb` either, so those could be `PurePaths` too. I'll try that out.

PR #103, opened 2025-05-03
Adjust first usage docs

halamix2:
- change `pip install https://github.com/isledecomp/reccmp` to `pip install git+https://github.com/isledecomp/reccmp`, otherwise the dependency won't install
- Adjust places referencing `reccmp-project create` `--originals` flag, I feel it wasn't obvious for the first time users that you can provide multiple space-separated binaries


jonschz:
Looks good, thanks!

Issue #104, opened 2025-05-03
JSON-schema for project .yaml files


halamix2:
Tracking progress:
We could create json-schema files to check if `reccmp-project/user/build.yaml` files are correct; useful for manual changes

Then either add the files to the https://www.schemastore.org, or add a bit to the documentation on how to set it up in e.g Visual Studio code settings:
```json
"yaml.schemas": {
  "./.reccmp-project-schema.json": "/reccmp-project.yml",
  "./.reccmp-user-schema.json": "/reccmp-user.yml",
  "./.reccmp-build-schema.json": "/reccmp-build.yml"
}
```

disinvite:
I think a prerequisite for this is to have all three files use the same structure. We touched on this (and other related things) in #7 and #8.

PR #105, opened 2025-05-03
add draft json schemas

halamix2:
- #104


jonschz:
Just a suggestion: Have you looked into generating the JSON schemas from pydantic (https://docs.pydantic.dev/latest/concepts/json_schema/#JSON%20Schema)? If that works well enough, it would be preferable since then you'd have a single source of truth and wouldn't have to maintain the schema in two places.

halamix2:
I haven't, I'll look into that

halamix2:
I've added required changes to the `reccmp/project/config.py` file, and for now I've added `generate_schemas.py` and the generated files.
Now the questions are
- where to put the generated files
- and where to put the generator, should it run in `pip install`, should it run manually, or even as a push Github Action?
  - if run manualy should it be connected to a postcommit githook?

PR #107, opened 2025-05-09
Add annotations for line matching

jonschz:
As briefly discussed on Matrix, here is my new feature for matching code lines. This is very useful when a function does not yet match well and the diff misjudges which parts belong together. It also makes it much easier to catch cases where the recomp is out of order with respect to the original.

Using this new code, I made significant progress on https://github.com/isledecomp/isle/issues/1427. I'll give it another go to get that one done completely before opening a PR.

This PR also contains a bunch of refactors. If you prefer, I will pull out the refactors into a separate PR so the change for the new feature is smaller.

Some open tasks:
- [x] Update documentation (for code annotations)

Code annotations look like 
```cpp
	short token = 0;
	// LINE: BETA10 0x1013e643
	short xmax = xofs + width - 1;
```

Here is a preview (also contained in the test), stock behaviour:
```diff
+++
@@ -0x200,19 +0x400,22 @@
0x200 : sub ecx, eax
0x202 : dec ecx
-0x203 : -mov word ptr [ebp - 0x28], cx
-0x207 : -jmp 0x1d1
-0x20c : -jmp 0xe
-0x211 : -movsx eax, word ptr [ebp - 0x18]
-0x215 : -movsx ecx, word ptr [ebp - 0x28]
+      : +mov word ptr [ebp - 4], cx
+      : +mov eax, dword ptr [ebp - 0xc]
+      : +mov ax, word ptr [eax]
+      : +mov word ptr [ebp - 0x10], ax
+      : +add dword ptr [ebp - 0xc], 2
+      : +movsx eax, word ptr [ebp - 0x10]
+      : +test eax, eax
+      : +jge 0x7e
+      : +movsx eax, word ptr [ebp - 0x10]
+      : +test ah, 0x40
+      : +je 0x13
+      : +movsx eax, word ptr [ebp - 4]
+      : +movsx ecx, word ptr [ebp - 0x10]
0x219 : add eax, ecx
-0x21b : -mov word ptr [ebp - 0x28], ax
-0x21f : -mov eax, dword ptr [ebp - 0x14]
-0x222 : -mov ax, word ptr [eax]
-0x225 : -mov word ptr [ebp - 0x18], ax
-0x229 : -add dword ptr [ebp - 0x14], 2
-0x22d : -movsx eax, word ptr [ebp - 0x18]
-0x231 : -test eax, eax
-0x233 : -jl 0xa
-0x239 : -jmp 0x19a
-0x23e : -jmp 0x68
-0x243 : -test byte ptr [ebp - 0x17], 0x40
+      : +mov word ptr [ebp - 4], ax
+      : +jmp 0x161
+      : +mov eax, dword ptr [ebp - 0x10]
+      : +push eax
+      : +mov eax, dword ptr [ebp - 4]
+      : +push eax
```
and with an annotation (note the improvement from orig `0x21b` to `0x22d`):
```diff
+++
@@ -0x200,9 +0x400,3 @@
0x200 : sub ecx, eax
0x202 : dec ecx
-0x203 : -mov word ptr [ebp - 0x28], cx
-0x207 : -jmp 0x1d1
-0x20c : -jmp cppfile.cpp:384 (LINE)
-0x211 : -movsx eax, word ptr [ebp - 0x18]
-0x215 : -movsx ecx, word ptr [ebp - 0x28]
-0x219 : -add eax, ecx
-0x21b : -mov word ptr [ebp - 0x28], ax
+      : +mov word ptr [ebp - 4], cx

---
+++
@@ -0x21f,1 +0x407,1 @@
-0x21f : -mov eax, dword ptr [ebp - 0x14]
+      : +mov eax, dword ptr [ebp - 0xc]

---
+++
@@ -0x222,9 +0x40a,18 @@
0x222 : mov ax, word ptr [eax]
-0x225 : -mov word ptr [ebp - 0x18], ax
-0x229 : -add dword ptr [ebp - 0x14], 2
-0x22d : -movsx eax, word ptr [ebp - 0x18]
+      : +mov word ptr [ebp - 0x10], ax
+      : +add dword ptr [ebp - 0xc], 2
+      : +movsx eax, word ptr [ebp - 0x10]
0x231 : test eax, eax
-0x233 : -jl 0xa
-0x239 : -jmp 0x19a
-0x23e : -jmp 0x68
-0x243 : -test byte ptr [ebp - 0x17], 0x40
+      : +jge 0x7e
+      : +movsx eax, word ptr [ebp - 0x10]
+      : +test ah, 0x40
+      : +je 0x13
+      : +movsx eax, word ptr [ebp - 4]
```


reccmp/isledecomp/compare/functions.py
```diff
@@ -0,0 +1,297 @@
...
+            is_effective,
+            local_ratio * (len(orig) + len(recomp)),
+        )
+
+    def _collect_line_annotations(self, recomp: AsmExcerpt) -> list[ReccmpMatch]:
+        """
+        Finds all `// LINE:` annotations within the given function
+        and drops any whose order is not consistent between original and recomp.
+        """
+        recomp_start_addr = recomp[0][0]
```
disinvite, line 201: It blows up here if the `recomp_combined` list is empty. That happens (for me) on `LEGO1` because of this annotation:
```
// LIBRARY: LEGO1 0x1008b340
// __FPinit
```
Ghidra identifies this as `__fpmath` instead. The problem is that `capstone` can't decode the bytes from the `__FPinit` variable into proper instructions, so the lists for `orig_combined` and `recomp_combined` are both empty.

jonschz: Good catch! In that case the check won't be needed anyway

reccmp/isledecomp/compare/match_msvc.py
```diff
@@ -328,3 +330,62 @@ def match_strings(db: EntityDb, report: ReccmpReportProtocol = reccmp_report_nop
...
+            # ```
+            # but it is significantly more effort to detect these false positives.
+            #
+
+            line_key = next(
+                (
+                    key
+                    for key, value in cv.lines.items()
+                    # We match `line + 1` since `line` is the comment itself
+                    if value.filename == filename and value.line_number == line + 1
```
disinvite, line 376: For non-Windows hosts, we need to do path conversion here or the Wine-like PDB paths won't match anything. We don't necessarily need to hold this PR up to solve that, though.

disinvite (approved):
Very impressive! Great idea to pull out all the asm-generation and comparison stuff into its own module. I think we just need the safeguard on an empty `recomp_combined` array and this is good to merge.

I think it's fine to solve the path conversion item in a follow-up because this is a new feature.

jonschz:
@disinvite feel free to try again, the bug should be fixed (I was able to reproduce it in a unit test)

disinvite:
Looks good to me!

madebr (approved):

PR #111, opened 2025-05-19
Find vtordisp functions using regex

### LINKED ISSUE #110: "Recomp vtable is larger" report skips most vtables
The `_find_vtordisp` function checks for vtables where recomp is larger than orig (true for many in `BETA10`) but this check is only done for the multiple inheritance vtables. We can easily pull this out into its own function and check all vtables, but for `BETA10` the list increases from 12 to 78 vtables.

Should this warning be downgraded to a debug alert instead?

disinvite:
The vtordisp functions (thunks used in vtables for multiple inheritance) take one of three forms:

```asm
; vtordisp{byte, 0}
sub ecx, dword ptr [ecx - byte]
jmp function

; vtordisp{byte, dword}
sub ecx, dword ptr [ecx - byte]
add ecx, dword
jmp function

; vtordisp{byte, byte}
sub ecx, dword ptr [ecx - byte]
sub ecx, byte
jmp function
```

If there are others, we haven't seen them yet in `LEGO1`/`BETA10` and we should add them.

Our current method of identifying and matching these is to scan each multiple inheritance vtable and check the bytes of the function in each position. We can tell which vtables have multiple inheritance by their name in the PDB. e.g. ``Act2Actor::`vftable'{for `LegoPathActor'}``

This change now identifies these functions via regex search without having to mark vtables first. The functions are very distinctive so I would not expect false positives unless the binary has handwritten assembly.

We can now drop the `is_vtordisp` function from the database. This relied upon the demangler module to do the check. We can't properly name these functions without having a name for the thunked function, but that's nothing new.

This marks a lot of new vtordisp functions in `BETA10`: the reason we could not do this before is that there are many layers of indirection and not all of these functions were matched. The vtables point to a thunk on the vtordisp, which itself points to a thunk on the real function.

Also fixes #110.

### DIFF
--- a/reccmp/isledecomp/compare/core.py
+++ b/reccmp/isledecomp/compare/core.py
@@ -21,7 +21,7 @@ from reccmp.isledecomp.compare.event import (
     reccmp_report_nop,
     create_logging_wrapper,
 )
-from reccmp.isledecomp.analysis import find_float_consts
+from reccmp.isledecomp.analysis import find_float_consts, find_vtordisp
 from .match_msvc import (
     match_lines,
     match_symbols,
@@ -95,7 +95,8 @@ class Compare:
         self._match_imports()
         self._match_exports()
         self._match_thunks()
-        self._find_vtordisp()
+        self._match_vtordisp()
+        self._check_vtables()
         self._unique_names_for_overloaded_functions()
 
         self.function_comparator = FunctionComparator(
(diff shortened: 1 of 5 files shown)


madebr (approved):

jonschz:
I may be able to take a more detailed look later. This looks quite promising. I have a few questions regarding future proofing / scaling behavior:
- Let's say a different version of MSVC produces vtordisp's that don't match this pattern. Will an error message be raised?
- What is the exact reason that we now see more vtordisps in BETA10? Did they come from unmatched vtables?
- In case we ever support 64 bit, do you think this pattern can be extended easily?

jonschz:
Thanks, looks good!

jonschz (approved):

reccmp/isledecomp/compare/core.py
```diff
@@ -593,70 +594,81 @@ def _match_exports(self):
...
-                # This could happen in debug builds due to code changes between BETA10 and LEGO1,
-                # but we have not seen it yet as of 2024-08-28.
+
+            orig_size_upper_limit = next_orig - match.orig_addr
+            if orig_size_upper_limit < match.size:
                 logger.warning(
                     "Recomp vtable is larger than orig vtable for %s",
                     match.name,
                 )
+                continue
```
jonschz, line 658: Do we really want to skip here? I think handling the sub-vtable until the minimum is fine

disinvite: That function is only checking the size of each vtable and reporting cases where we can prove recomp has more virtual methods. The `continue` here just skips the second check for null pointers (padding).

jonschz (approved):

PR #115, opened 2025-05-23
Fix union type crash (#113)

### LINKED ISSUE #113: Anonymous union with variable crash
When using `reccmp` against a test executable, if I define a union like this in a header:

```c
union {
    unsigned long x;
} myUnionField;
```

I get the following error:

```python
  File "<frozen runpy>", line 198, in _run_module_as_main
  File "<frozen runpy>", line 88, in _run_code
  File "C:\test\venv\scripts\reccmp-reccmp.exe\__main__.py", line 7, in <module>
  File "C:\test\venv\Lib\site-packages\reccmp\tools\asmcmp.py", line 217, in main
    isle_compare = IsleCompare(
                   ^^^^^^^^^^^^
  File "C:\test\venv\Lib\site-packages\reccmp\isledecomp\compare\core.py", line 79, in __init__
    self._load_cvdump()
  File "C:\test\venv\Lib\site-packages\reccmp\isledecomp\compare\core.py", line 117, in _load_cvdump
    self.cvdump_analysis = CvdumpAnalysis(self.cv)
                           ^^^^^^^^^^^^^^^^^^^^^^^
  File "C:\test\venv\Lib\site-packages\reccmp\isledecomp\cvdump\analysis.py", line 130, in __init__
    g_info = parser.types.get(glo.type)
             ^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "C:\test\venv\Lib\site-packages\reccmp\isledecomp\cvdump\types.py", line 395, in get
    members = self._get_field_list(obj)
              ^^^^^^^^^^^^^^^^^^^^^^^^^
  File "C:\test\venv\Lib\site-packages\reccmp\isledecomp\cvdump\types.py", line 302, in _get_field_list
    field_list_type = type_obj["field_list_type"]
                      ~~~~~~~~^^^^^^^^^^^^^^^^^^^
KeyError: 'field_list_type'
```

The following is fine though (although I understand they're not equivalent):

```c
union myUnion {
    unsigned long x;
};

struct {
    unsigned long x;
} myStructField;

struct myStruct {
    unsigned long x;
};
```

### LINKED ISSUE #114: C symbols not recognized in annotations
Using the function `strcat` as an example: it appears in the cvdump PDB output in these places:

```
*** PUBLICS
S_PUB32: [0001:001687F0], Flags: 00000000, _strcat


*** SYMBOLS
(000060) S_GPROC32: [0001:001687F0], Cb: 000000E0, Type:             0x18AA, strcat
         Parent: 00000000, End: 000001D8, Next: 00000000
         Debug start: 00000004, Debug end: 000000E0


*** GLOBALS
S_PROCREF: 0x00000000: ( 430, 00000060) strcat
```

From that we recognize that it is a function called `strcat` with the symbol/linker name of `_strcat`.

If you add this annotation expecting to match on the symbol:
```
// LIBRARY: TEST 0x1234
// _strcat
```
it won't work because we don't recognize the leading underscore as indicating a symbol. We look for a function with the _name_ `_strcat` and don't find one.

The workaround is to use just `strcat` in your annotation. The potential pitfall here is that nothing stops you from having a function with a leading underscore. STL uses underscores frequently for its internal data structures. For example: `_Tree<MxCore *,MxCore *,set<MxCore *,CoreSetCompare,allocator<MxCore *> >::_Kfn,CoreSetCompare,allocator<MxCore *> >::iterator::_Dec`

Another example: `__strlwr` and `_strlwr` are two different symbols (though they happen to point to the same function).

jonschz:
Fixes #113, though I am not entirely sure when and how the error occurs. There is a chance that the parsing of union types has some further issues. I never looked too deeply into it since we don't support them in Ghidra at the moment.

### DIFF
--- a/reccmp/isledecomp/cvdump/types.py
+++ b/reccmp/isledecomp/cvdump/types.py
@@ -736,6 +736,7 @@ class CvdumpTypesParser:
         if match.group("field_type") == "0x0000":
             self._set("is_forward_ref", True)
 
+        self._set("field_list_type", match.group("field_type"))
         self._set("size", int(match.group("size")))
         if match.group("udt") is not None:
             self._set("udt", normalize_type_id(match.group("udt")))
--- a/tests/test_cvdump_types.py
+++ b/tests/test_cvdump_types.py
@@ -665,6 +665,7 @@ def test_union(parser: CvdumpTypesParser):
     assert union == {
         "type": "LF_UNION",
         "name": "FlagBitfield",
+        "field_list_type": "0x2e84",
         "size": 1,
         "udt": "0x2e85",
     }
--- a/tests/test_cvdump_types.py
+++ b/tests/test_cvdump_types.py
@@ -703,6 +703,7 @@ def test_unnamed_union():
     # Make sure we can parse the members line
     union = parser.keys["0x369e"]
     assert union["size"] == 4
+    assert union["field_list_type"] == "0x369d"
 
 
 ARGLIST_UNKNOWN_TYPE = """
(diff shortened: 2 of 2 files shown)


disinvite (approved):

jonschz:
> Not sure how well Ghidra supports those.

Pretty sure it doesn't, but I think you can add them manually and it won't cause any issues. There weren't that many unions and bitfields in ISLE, so I imported them by hand back then.

PR #116, opened 2025-05-23
Import all code lines into the lines database

### LINKED ISSUE #109: Implement path conversion for `LINE` annotations
For non-Windows hosts, we need to do path conversion here or the Wine-like PDB paths won't match anything. We don't necessarily need to hold this PR up to solve that, though.

_Originally posted by @disinvite in https://github.com/isledecomp/reccmp/pull/107#discussion_r2085825545_

disinvite:
The lines database (lines.py) module converts the Windows-like paths from cvdump into the ones on your local filesystem. e.g. `Z:\Users\test\isle\ISLE\isleapp.cpp` into `/Users/test/isle/ISLE/isleapp.cpp`.

Until recently we have only been concerned with lines that indicate the start of a function so we can combine those with `// FUNCTION` annotations. We now want to use all the line-addr pairs.

This required the following changes, mostly to maintain performance:

1. In the cvdump `LINES` parser, we now group line-addr pairs by the code filename. This is always going to be a Windows-like path so we use `PureWindowsPath` for the type.
2. In the lines database, insert all lines for each filename all at once. This is to reduce the total number of calls.
3. Also in lines.py, the sqlite database is back after I removed it in #101. The difference now is that we use the hash on the `pathlib` object instead of storing the filename as a string. This is the same thing `pathlib` uses internally for equality checks, so it should work even when the host's paths are not case-sensitive.
4. We now mark which addresses from the lines section intersect with addresses from other sections in the PDB. This gives us the addresses that begin each function, so we can search using `find_function()`. We can search any range of code lines with `search_line()`.

These changes make the process faster if we import just the function starts, but importing everything means we break even on performance. (Great!)

We can now use the lines db in the function that matches `// LINE` annotations and the PDB paths will get converted correctly. The diff for the `DecodeSS2` function looks the same on my Mac as it does on Windows. Fixes #109.

### DIFF
--- a/docs/annotations.md
+++ b/docs/annotations.md
@@ -55,6 +55,13 @@ There are a few cases where two functions of the same name need to be annotated 
 // ?_Construct@@YAXPAPAVROI@@ABQAV1@@Z
 ```
 
+Names that begin with `?` are assumed to be [MSVC-like symbols](https://en.wikiversity.org/wiki/Visual_C%2B%2B_name_mangling). In all other cases, you should specify that the name is a symbol by adding `SYMBOL` after the address. For example:
+
+```c++
+// LIBRARY: TEST 0x10002000 SYMBOL
+// __strlwr
+```
+
 ### Annotation types
 
 #### `FUNCTION`
--- a/reccmp/isledecomp/analysis/__init__.py
+++ b/reccmp/isledecomp/analysis/__init__.py
@@ -1,1 +1,2 @@
 from .float_const import find_float_consts
+from .vtordisp import find_vtordisp
--- a/reccmp/isledecomp/analysis/vtordisp.py
+++ b/reccmp/isledecomp/analysis/vtordisp.py
@@ -1,0 +1,77 @@
+"""Find multiple-inheritance thunk (vtordisp) functions."""
+
+import re
+import struct
+from typing import Iterator, NamedTuple
+from reccmp.isledecomp.formats.pe import PEImage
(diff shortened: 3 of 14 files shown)


disinvite:
Yes, I tested with `flic.cpp` from your PR.

I don't think we need to worry about collisions. We were already using `PurePath` as a dict key in #101. This is just translating that back into SQL where we can't store the `PurePath` directly (without converting to string).

The reason to return to SQL at all is for the second (unfinished) feature: returning the code lines for a given range of addresses. It would be awkward to maintain two data structures for this, so we'll use the database to have less code.

jonschz:
> I don't think we need to worry about collisions. We were already using `PurePath` as a dict key in #101. This is just translating that back into SQL where we can't store the `PurePath` directly (without converting to string).


Not sure I fully get your point. Using the `PurePath` as a dict key is completely fine with me. Intuitively, I would have put its string representation into the SQL database instead of a hash of it. Anyway, if there's a good reason not to do that and if you consider the collision risk to be minor, feel free to merge.

disinvite:
We can just kick the can for now and keep using `dict`. Going back to SQL had more to do with performance for the addr search, and I'm still working on that feature.

Looks like there's [more going on](https://stackoverflow.com/a/9022664) behind the scenes to prevent hash collisions on dicts, so translating that to SQL will have to incorporate the original string value.

jonschz (approved):

PR #118, opened 2025-05-24
Use our PE module for verexp

disinvite:
When we started `LEGO1` we didn't have our own code to unpack the PE header, so `reccmp-verexp` used the text output from `DUMPBIN.EXE` to verify the exports.

The script now uses our `PEImage` module and we can drop `DUMPBIN` altogether.

You have to test this manually but I can see the expected diff if I remove an exported function or add an extra one in LegoOmni.def.


madebr (approved):
LGTM (when using the symbol names)
```py
pe = reccmp.isledecomp.formats.detect.detect_image("/path/to/LEGO1.DLL")
for i, e in enumerate(pe.exports):
  print(i, hex(e[0]), e[1])

print(f"{len(pe.exports)} symbols")
```
gives the same names as running `DUMPBIN.EXE /EXPORTS LEGO1.DLL`

It looks like our export addresses have the ImageBase offset applied to the symbols, whereas `DUMBPIN.EXE` does not.

PR #119, opened 2025-05-24
Symbol option for nameref annotations

### LINKED ISSUE #114: C symbols not recognized in annotations

disinvite:
Adds support for annotations like this:
```
// LIBRARY: TEST 0x1234 SYMBOL
// _exit
```
If `SYMBOL` is there (in any case) on a function annotation, we create the orig entity using `_exit` as the linker name (symbol) instead of the regular name. This will get matched to a recomp entity by its unique linker name.

This wasn't difficult to add because we already parse the (so called) extra field after the address for vtable multiple inheritance.

Note: for these annotations we only set the linker name and skip the regular name. If it doesn't get matched, you will see `<OFFSETX>` in the asm output because it has no name. (We could assume `name = value[1:]` and remove the leading underscore pretty easily.)

Thoughts on this approach?

Fixes #114 but only for functions.

### DIFF
--- a/reccmp/isledecomp/compare/core.py
+++ b/reccmp/isledecomp/compare/core.py
@@ -266,7 +266,7 @@ class Compare:
                     fun.offset, type=EntityType.FUNCTION, stub=fun.should_skip()
                 )
 
-                if fun.name.startswith("?"):
+                if fun.name.startswith("?") or fun.use_linker:
                     batch.set_orig(fun.offset, symbol=fun.name)
                 else:
                     batch.set_orig(fun.offset, name=fun.name)
--- a/reccmp/isledecomp/compare/core.py
+++ b/reccmp/isledecomp/compare/core.py
@@ -266,7 +266,7 @@ class Compare:
                     fun.offset, type=EntityType.FUNCTION, stub=fun.should_skip()
                 )
 
-                if fun.name.startswith("?") or fun.use_linker:
+                if fun.name.startswith("?") or fun.name_is_symbol:
                     batch.set_orig(fun.offset, symbol=fun.name)
                 else:
                     batch.set_orig(fun.offset, name=fun.name)
--- a/reccmp/isledecomp/parser/node.py
+++ b/reccmp/isledecomp/parser/node.py
@@ -36,7 +36,8 @@ class ParserFunction(ParserSymbol):
     # referenced by name, but only if this flag is true.
     lookup_by_name: bool = False
 
-    use_linker: bool = False
+    # True if the annotation name is the linker name (symbol) for this entity.
+    name_is_symbol: bool = False
 
     def should_skip(self) -> bool:
         return self.type == MarkerType.STUB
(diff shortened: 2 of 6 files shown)


reccmp/isledecomp/compare/core.py
```diff
@@ -266,7 +266,7 @@ def orig_bin_checker(addr: int) -> bool:
                     fun.offset, type=EntityType.FUNCTION, stub=fun.should_skip()
                 )
 
-                if fun.name.startswith("?"):
+                if fun.name.startswith("?") or fun.use_linker:
```
jonschz, line 269: Doing it this way reduces boilerplate, but there is some implicit logic here that might not be intuitive. I'm a bit torn on this, but I do see the benefit of being able to skip the explicit `SYMBOL` in most cases.

reccmp/isledecomp/parser/node.py
```diff
@@ -36,6 +36,8 @@ class ParserFunction(ParserSymbol):
     # referenced by name, but only if this flag is true.
     lookup_by_name: bool = False
 
+    use_linker: bool = False
```
jonschz, line 39: How about `lookup_by_linker_symbol`, `is_linker_symbol`, or something along these lines?

tests/test_parser.py
jonschz, line 1: very nice test cases!

jonschz (approved):
This looks very good to me.

The only thing I miss is an update to the markdown documentation, including details on when exactly the match by symbol will be triggered.

reccmp/isledecomp/compare/core.py
disinvite: I agree, but we would have to fix all the existing annotations for `LEGO1`. (Not impossible). The better way long-term is something like #32 so you don't need these annotations in code files at all.

jonschz:
LGTM

PR #121, opened 2025-05-25
Run reccmp on C-like languages (C/C++)

madebr:


madebr:
@dethrace-labs Will this allow it to run on dethrace?

reccmp/isledecomp/dir.py
```diff
@@ -133,8 +133,18 @@ def resolve_cvdump(self, path_str: str) -> str:
...
+    return Path(filename).suffix.lower() in (
+        ".c",
+        ".h",
+        ".cc",
+        ".hh",
+        ".cxx",
+        ".hxx",
+        ".cpp",
+        ".hpp",
+        ".C",
```
disinvite, line 146: Can drop the upper case C :)

dethrace-labs:
> @dethrace-labs Will this allow it to run on dethrace?

Yes, well, it removed my blocker, the couple of functions I've tried so far I've got to 100% now. Will try a few more complex cases to try and ensure I've got the right compiler flags, then will work on introducing this to our dethrace workflow 🎉

Issue #106, opened 2025-05-07
Universal type database


disinvite:
We currently read type information by parsing the text output of `cvdump.exe` used on the pdb file. Parsing and type management (and resolving forward references) are all combined in one module. The preliminary step here would be to decouple these two actions.

`datacmp` and the Ghidra scripts are the main consumers for type information. (`reccmp` also uses it to match up array offsets.) If we wanted to add more structure to the type database (beyond just dict keys) it may be worth looking at how Ghidra manages this stuff internally.

The background is that I was recently looking into another game: [Pirates: Captain's Quest](https://archive.org/details/pirates-captains-quest-1996). I haven't run this to confirm, but it apparently shipped with a debug build in the form of `pir32db.exe`.

This is a Borland compiled program, so you need to use `tdump.exe` from [Borland C++ 5.0](https://winworldpc.com/product/borland-c/5x) to read the debug data. The presentation of types is remarkably similar. Excerpt:

```
20956A  Type: 1000  Len: 001C CLASS 
  Fields: 000A  FieldIdx: 1D87  Name: BCRestoreGameScene [277]
  Containing Class: 0000  Length: 0094
  Packed: No   Constructors: Yes  Destructors: Yes  Overloaded operators: No 
  Is nested: No   Nested classes: No   Overload assigments: Yes
  Casting methods: No   Forward Ref: No 

20958A  Type: 1001  Len: 0008 POINTER 
  near32 *  (PtrType: 0A  PtrMode: 0)
  Points to: 1000

209596  Type: 1002  Len: 0008 POINTER 
  near32 *  (PtrType: 0A  PtrMode: 0)
const   Points to: 1000

2095A2  Type: 1003  Len: 0014 ENUM 
  Count: 005F  Type: rint32  Fields: 1647  Class: 0000  Name: BCSceneEnum [1E3]
```

disinvite:
This topic came up again recently with regard to performance. To prepare to migrate off of cvdump, we need an API to get only what we want out of the database.

My original idea (when opening this) was to come up with an alternate data structure not (entirely) based on the codeview leaf structures. It would be better to just return simple data types through various functions depending on what kind of information you want. For example:
- `get_size_for_type(key)` for sizing entities that have a data type
- `get_offset_name(key, offset)` already exists, used during asm sanitize
- `get_format_string(key)` already exists, used by `datacmp` for `struct.unpack`

[...]

PR #130, opened 2025-06-06
Refactor cvdump `TYPES` parsing

disinvite:
The output from cvdump looks like this:

```
*** MODULES

0001 "RelWithDebInfo\LEGO1.exp"
0002 "CMakeFiles\lego1.dir\RelWithDebInfo\LEGO1\define.cpp.obj"
(...)

*** PUBLICS

S_PUB32: [0001:000AA3F0], Flags: 00000000, ??1MxFrequencyMeter@@QAE@XZ
S_PUB32: [0001:00049DA0], Flags: 00000000, ??_ELegoEndAnimNotificationParam@@UAEPAXI@Z
(...)
```

Each section begins on its own line with three asterisks and the section name.[^1]

Until now we've parsed this entire text one line at a time, which requires a state machine to keep track of where we are. The `TYPES` section has its own state machine to parse each type leaf. For example:

```
0x10b4 : Length = 34, Leaf = 0x1504 LF_CLASS
    # members = 55,  field list type 0x10b2, CONSTRUCTOR, OVERLOAD, 
    Derivation list type 0x0000, VT shape type 0x10b3
    Size = 148, class name = MxDSAction, UDT(0x00006529)

0x10b5 : Length = 30, Leaf = 0x1504 LF_CLASS
    # members = 0,  field list type 0x0000, FORWARD REF, 
    Derivation list type 0x0000, VT shape type 0x0000
    Size = 0, class name = MxParam, UDT(0x00002bed)
```

To eliminate most of the state variables in `CvdumpTypesParser`, I changed the code in `runner.py` to load an entire cvdump section at a time. This required changing most of the regular expressions so we could match in the middle of the text (with `re.search()`) instead of clamping to the start of each line. In cases where we had to match one line first, then a second, I combined the two regular expression to match both items.

Microsoft has made some[^2] of the code for cvdump available. If we look at [cvinfo.h](https://github.com/microsoft/microsoft-pdb/blob/master/include/cvinfo.h) we can see what data is there for each kind of type leaf. For example:

```cpp
typedef struct lfEnum_16t {
    unsigned short  leaf;           // LF_ENUM_16t
    unsigned short  count;          // count of number of elements in class
    CV_typ16_t      utype;          // underlying type of the enum
    CV_typ16_t      field;          // type index of LF_FIELD descriptor list
    CV_prop_t       property;       // property attribute field
    unsigned char   Name[1];        // length prefixed name of enum
} lfEnum_16t;
```

Not that aren't still surprises to be found, but I think we can be a bit more confident in how we parse the leaves by anticipating which items _could_ appear. For example: the similar layouts for `LF_MFUNCTION` and `LF_PROCEDURE`.

If one of our regular expressions doesn't match, we now raise an assert error. This is caught by the `except` trap so we continue parsing but dump the leaf contents using the `logger`. This should make it easier to react to new parsing issues as they come up.

As for performance: it is better if you read the input from a file -- i.e. eliminating the `subprocess` overhead. On non-Windows platforms where we need to use `wine`, it is about 20%[^3] slower. I'm not really sure why: None of the profiling I've done can point to the bottleneck. Waiting for a buffer to get filled maybe?

The benefit for doing this is that the code is a lot simpler to reason about, and we are now separating type parsing from managing the type database (#106). I dumped the output of `cvdump.types.keys` and we get the same data before and after for both `LEGO1` and `BETA10`. We are still using the original unit tests for the `TYPES` section but we could start to break this out into leaf-specific items in later updates.

Code to dump the finished type database for testing/comparison:
```python
def dump(pdbfile):
    cv = (
        Cvdump(pdbfile)
        .lines()
        .globals()
        .publics()
        .symbols()
        .section_contributions()
        .types()
        .run()
    )

    for key, value in cv.types.keys.items():
        print(key, value)

```

[^1]: We must be careful when parsing `TYPES` to not start a new section on the line `*** Converting 16-bit types to 32-bit equivalents`.
[^2]: The file `type7.cpp` is blank.
[^3]: This sounds worse than it is. On my M3 Macbook it is 200ms slower.

### DIFF
--- a/reccmp/isledecomp/cvdump/parser.py
+++ b/reccmp/isledecomp/cvdump/parser.py
@@ -1,11 +1,9 @@
 import re
 from pathlib import PureWindowsPath
-from typing import Iterable, NamedTuple
+from typing import NamedTuple
 from .types import CvdumpTypesParser
 from .symbols import CvdumpSymbolsParser
 
-# e.g. `*** PUBLICS`
-_section_change_regex = re.compile(r"\*\*\* (?P<section>[A-Z/ ]{2,})")
 
 # e.g. `     27 00034EC0     28 00034EE2     29 00034EE7     30 00034EF4`
 _line_addr_pairs_findall = re.compile(r"\s+(?P<line_no>\d+) (?P<addr>[A-F0-9]{8})")
@@ -107,7 +105,6 @@ class LinesFunction(NamedTuple):
 class CvdumpParser:
     # pylint: disable=too-many-instance-attributes
     def __init__(self) -> None:
-        self._section: str = ""
         self._lines_function = LinesFunction(PureWindowsPath(), 0)
 
         self.lines: dict[PureWindowsPath, list[LineValue]] = {}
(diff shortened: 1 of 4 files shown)


madebr:
> 2\. The file `type7.cpp` is blank. [↩](#user-content-fnref-2-49b82b835972f9a0ad4f6468615eadf6)

https://github.com/madebr/microsoft-pdb/blob/wip/cvdump/type7.cpp (untested code)

madebr (approved):

madebr:
> I had been thinking about that change for a while myself! I'll take a closer look, but if @madebr already reviewed it, there's no need to wait as far as I am concerned.

The changes looked like a straightforward refactoring to me. But please feel free to do another (more thorough) review.

jonschz:
Yeah, if the tests are happy and there's no regression on ISLE, we're probably good to go. @disinvite I assume you checked on ISLE?

disinvite:
That one is fixed. We were failing to parse it correctly before, but it failed on this line instead:
```
[ERROR] Unrecognized pointer attribute:         Type of pointer to member = Not specifie
```

Excluding the change introduced by that fix, we have the same types database for all the `ISLE` binaries before and after this PR.

I forgot that @madebr reversed some of the missing pieces from cvdump. That code shows that the "pointer type" is more complex than just our list of possible values. In this case, though, it worked to alert us to the missed match.

Issue #132, opened 2025-06-07
Make tests `mypy` compliant


jonschz:
As the title says, there are a few typing errors in the tests.

- [ ] Fix the issues in the current tests
- [ ] Update the documentation (if needed)
- [ ] Add `mypy ./tests` to CI

PR #133, opened 2025-06-07
Improve `pathlib` usage in `test_lines.py`

disinvite:
The existing tests have two problems:

1. They use `PurePath` for the sample value.
2. They use that value with two different kinds of functions: those that expect a local path, and `add_line()` that expects the PDB path.

The value `test.cpp` is simple enough that this didn't matter (i.e. it worked through implicit conversion). Using `PurePath` means that you get either `PureWindowsPath` or `PurePosixPath` depending on where you run the test. Since we test both on Windows and Ubuntu hosts, we do test everything.

It's better to be explicit. We now use better sample values that are not specific to the test-runner platform. I also improved the comments and variable names so it's more clear what's happening at each stage. We also pass the `mypy` check for this file. (#132)

### DIFF
--- a/tests/test_lines.py
+++ b/tests/test_lines.py
@@ -1,64 +1,83 @@
 """Testing lines database: mapping between (filename, line_number) to virtual address."""
 
-from pathlib import PurePath, PurePosixPath, PureWindowsPath
+from pathlib import PurePosixPath, PureWindowsPath
+import pytest
 from reccmp.isledecomp.compare.lines import LinesDb
 
-TEST_PATH = PurePath("test.cpp")
+# For tests that don't require path conversion, parametrize the
+# input local_path so that we test both Windows and Posix paths.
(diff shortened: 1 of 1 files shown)


tests/test_lines.py
```diff
@@ -82,40 +103,43 @@ def test_db_hash_posix():
     The goal here is to show that we are not taking liberties with PurePath
     and when it can be expected to match."""
 
-    path = PurePosixPath("test.cpp")
-    lines = LinesDb([path])
-    lines.add_line(path, 2, 0x1234)
+    local_path = PurePosixPath("test.cpp")
+    lines = LinesDb([local_path])
+
+    pdb_path = PureWindowsPath("test.cpp")
```
jonschz: Would it make sense to include (back)slashes here as well?

jonschz (approved):
Looks good!

PR #139, opened 2025-06-15
Fix displacement reads into data

jonschz:
While this doesn't appear to happen on ISLE, later MSVC versions like MSVC 7.00 use instructions like
```
mov dword ptr [eax*4 + 0xc815a8], edi
```
to look up values in arrays. So far, `reccmp` was built on the assumption that this type of pointer dereference only happens for vtables or stack variables (edit: or relocation table entries).

It is non-trivial to distinguish these cases. My approach was to check if the base address in the displacement is an address pointing to something we know.

Not sure if I found the most efficient solution, but it produced no regressions on ISLE (the reviewer is encouraged to double check that) and it solves the problem on MSVC 7.00.


tests/test_function_comparator.py
```diff
@@ -27,12 +26,19 @@ def compare_functions(
...
 
-    orig_bin = Mock(PEImage)
+    # Do not use `spec=PEImage`. It may have default implementations that don't do what you expect
+    # (like functions that always return `True`).
+    orig_bin = Mock(spec=[])
     orig_bin.read = Mock(return_value=orig)
     orig_bin.imagebase = 0
-    recomp_bin = Mock(PEImage)
+    orig_bin.is_relocated_addr = Mock(return_value=False)
+    orig_bin.is_debug = Mock(return_value=False)
```
jonschz, line 35: This fixes an issue with the test setup. Without it, `test_displacement_without_match` erroneously thinks that the `diffreport.ratio == 1` because both offsets are replaced by `<OFFSET1>` since `is_relocated_addr` always returns `True`.

reccmp/isledecomp/compare/functions.py
```diff
@@ -24,11 +25,30 @@ class FunctionPartCompareResult(NamedTuple):
...
 
 
-def create_reloc_lookup(bin_file: PEImage) -> Callable[[int], bool]:
-    """Function generator for relocation table lookup"""
+def create_valid_addr_lookup(
+    db_getter: Callable[[int, bool], ReccmpEntity | None],
+    is_recomp: bool,
+    bin_file: PEImage,
+) -> Callable[[int], bool]:
+    """Function generator for a lookup whether an address from a call is valid (either )"""
```
disinvite: Incomplete docstring?

jonschz:
The C++ code is something like: 
```
g_unk0xc815a8[i] = something
```
where `sizeof(g_unk0xc815a8[0]) == 4`. The old MSVC does not compile this line to displacement reads (I think it might use `lea` instead). So our assumption that all such displacements are relocations does not hold for newer MSVC versions.

disinvite:
I'm surprised that address is _not_ relocated. Could there be more relocation tables in this newer binary and we're missing them?

In any case: this looks good. I just tested with #141 and (as expected) I see no change to the asm generated for `LEGO1` and `BETA10`.

disinvite (approved):

jonschz:
It doesn't look relocated in Ghidra:
grafik
This is what you find at the address:
grafik

disinvite:
Maybe our problem is that we ignore the [relocation type](https://learn.microsoft.com/en-us/windows/win32/debug/pe-format#base-relocation-types).

disinvite:
No, let's merge this one.

PR #140, opened 2025-06-16
Unify size for string entities

disinvite:
A small change as part of a larger refactor involving string entities and how we detect and match them.

The reported size for a string entity now includes the null terminator. The ASCII string "Pizza" would now have an entity size of 6 instead of the string length of 5.

The goal here is to be consistent about adding string data from multiple sources. The size should represent the memory footprint for that entity, irrespective of string encoding. If this were a UTF-16 string, the size should be 12. (2-byte null terminator). If it were a pascal string, the size would still be 6, but the data would be `b"\x05Pizza"` instead of `b"Pizza\x00"`.

I used this code as a manual test:
```python
isle_compare = go(origfile, recompfile, pdbfile, codepath)
for (orig_addr, recomp_addr, size, name) in isle_compare._db.sql.execute(
    """SELECT orig_addr, recomp_addr,
    coalesce(json_extract(kvstore,'$.size'), 0),
    coalesce(json_extract(kvstore,'$.name'), '')
    FROM entities WHERE coalesce(json_extract(kvstore,'$.type'), 0) = ?
    """,
    (EntityType.STRING,)
):
    assert size is not None
    orig_raw = origfile.read(orig_addr, size) if orig_addr is not None else None
    recomp_raw = recompfile.read(recomp_addr, size) if recomp_addr is not None else None

    if name:
        assert name[-1] != "\x00"

    if orig_raw and recomp_raw:
        assert orig_raw == recomp_raw

    try:
        if orig_raw is not None:
            try:
                assert orig_raw[-1] == 0
                assert orig_raw[-2] != 0
            except IndexError:
                pass
            assert orig_raw.decode("latin1").rstrip("\x00") == name

        if recomp_raw is not None:
            try:
                assert recomp_raw[-1] == 0
                assert recomp_raw[-2] != 0
            except IndexError:
                pass
            assert recomp_raw.decode("latin1").rstrip("\x00") == name
    except AssertionError as ex:
        print(orig_raw, recomp_raw)
        #raise ex
```

### DIFF
--- a/reccmp/isledecomp/compare/core.py
+++ b/reccmp/isledecomp/compare/core.py
@@ -195,13 +195,21 @@ class Compare:
                     except UnicodeDecodeError:
                         pass
 
-                batch.set_recomp(
-                    addr,
-                    type=sym.node_type,
-                    name=sym.name(),
-                    symbol=sym.decorated_name,
-                    size=sym.size(),
-                )
+                    batch.set_recomp(
+                        addr,
+                        type=sym.node_type,
+                        name=sym.name(),
+                        symbol=sym.decorated_name,
+                        size=len(rstrip_string) + 1,
+                    )
+                else:
+                    batch.set_recomp(
+                        addr,
+                        type=sym.node_type,
+                        name=sym.name(),
+                        symbol=sym.decorated_name,
+                        size=sym.size(),
+                    )
 
         for filename, values in self.cvdump_analysis.lines.items():
             lines = [
@@ -312,7 +320,7 @@ class Compare:
                     string.offset,
                     name=string.name,
                     type=EntityType.STRING,
-                    size=len(string.name),
+                    size=len(string.name) + 1,
                 )
 
             for line in codebase.iter_line_symbols():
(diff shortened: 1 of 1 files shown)


jonschz:
> If it were a pascal string, the size would still be 6

My Pascal / Delphi times have been ages ago, but didn't they also have a null terminator so you wouldn't have to do any conversions when calling Windows API functions. Just off on a tangent.

reccmp/isledecomp/compare/core.py
```diff
@@ -195,13 +195,21 @@ def _load_cvdump(self):
...
+                        size=len(rstrip_string) + 1,
+                    )
+                else:
+                    batch.set_recomp(
+                        addr,
+                        type=sym.node_type,
+                        name=sym.name(),
+                        symbol=sym.decorated_name,
+                        size=sym.size(),
+                    )
```
jonschz, line 215: Maybe add a comment here what and why we are doing something differently

jonschz (approved):
Looks good! I didn't find any regressions.

disinvite:
> > If it were a pascal string, the size would still be 6
> 
> My Pascal / Delphi times have been ages ago, but didn't they also have a null terminator so you wouldn't have to do any conversions when calling Windows API functions. Just off on a tangent.

That sounds right. My only Pascal experience is a decomp on DOS Turbo Pascal 7 where they used the length byte prefix.

disinvite:
I forgot to mention the entire reason for doing this: the PDB gives most strings a symbol if string pooling is enabled. This is typically the case for non-debug builds. We had been using the string length [from the mangled name](https://en.wikiversity.org/wiki/Visual_C%2B%2B_name_mangling) as the entity size, but this is not consistent with respect to the null terminator. Sometimes it is considered part of the string length, but sometimes not.

PR #141, opened 2025-06-16
Fix asm debug (#136)

### LINKED ISSUE #136: Debug asm dump not working
Running `reccmp-reccmp` with `--debug` should create two files in the working directory: the complete asm output for every function we disassembled in both orig and recomp.

After #107, you can't enable this output because changes to the `debug` property in the `Compare` object are not reflected in the `FunctionComparator`.

Other items to fix: the address output is in decimal. This used to be a string, so we just need to convert back to hex. To keep the output files unique, I generated a random hash used by both the orig and recomp files. It would be more helpful to just use a timestamp instead. (same value for both files)

disinvite:
Fixes #136. We now use a better filename so it's easy to tell which files go together.

### DIFF
--- a/reccmp/isledecomp/compare/core.py
+++ b/reccmp/isledecomp/compare/core.py
@@ -3,7 +3,6 @@ import logging
 import difflib
 from pathlib import Path
 import struct
-import uuid
 from typing import Iterable, Iterator
 from reccmp.isledecomp.compare.functions import FunctionComparator
 from reccmp.isledecomp.formats.pe import PEImage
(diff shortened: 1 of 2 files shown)


disinvite:
> Looks good! I don't see a good way to test this, so I'm fine with not adding any tests here. An integration / e2e test feels appropriate to me, but we haven't set up anything like that. Good to merge as far as I am concerned.

This feature isn't really documented, but the test is just to run `reccmp-reccmp` with `--debug`. If debug is enabled then it should create the two files.

We should definitely look into e2e tests of some kind. That probably needs its own issue for coming up with ideas.

PR #143, opened 2025-06-19
Create `isle_compare` using reccmp target

disinvite:
Small refactor to consolidate some code. The test is to make sure you can still run asmcmp, datacmp, and stackcmp.


jonschz (approved):
Very nice!

Issue #27, opened 2024-12-01
Manage const data differently


disinvite:
We have been storing strings in our program database without considering the actual bytes. This has worked fine for most ASCII strings, but we should support unicode too.

The PDB for LEGO1.DLL points at a string `"\x00\x00"` in the CRT. This is functionally equivalent to the empty string, but these should not collide in the database if the linker has them as two different items.

Matching const data has not been strictly necessary: duplicated values appear the same when substituted in the assembly output. Tools like `roadmap` or `datacmp` need these to be matched, though. I think we can do this somewhat easily by using the relocation table, collecting pointers to floats or strings, then matching those according to the matched functions where they are used.

jonschz:
The following issue might or might not be related: In LEGO1, we have an annotation
```
// GLOBAL: LEGO1 0x100fd200
// ___decimal_point
```
while the debug symbols insist that this is a `STRING`. However, changing this annotation to a `STRING` causes mismatches in other functions, e.g. in `0x10090580`:
```diff
-0x100908dd : -cmp bl, byte ptr ['.' (STRING)]
+           : +cmp bl, byte ptr [___decimal_point (UNK)]
```

disinvite:
That's caused by our string matcher overstepping its bounds a little bit. We check the relocation table to get verified addresses, then look in each location for an ascii-like sequence with a null-terminator. This includes the `___decimal_point` variable from the CRT library, and we assume incorrectly that it is a string. I can't find anywhere in LEGO1 that we use the string `"."`.

These short strings (4 chars or less) are the ones that are critical to annotate because we (mostly) can't make the assumption that they are strings as opposed to variables or other data.

disinvite:
I closed this prematurely, although we have made a lot of progress. My intended goal was to use a SQL table specifically for const data. Once we have that, good to close.

PR #147, opened 2025-06-20
Redesign string search and matching

disinvite:
For detecting and matching string data, we do two things:

1. Identify "verified" strings using any information at our disposal. This includes `// STRING` annotations and symbols in the PDB that begin with `??_C@`.
2. Using the binary's relocation table, test each address from a non-code section to see if it resembles a string.

Until now, we only matched the verified strings. It's not vital to match a string entity for the sake of generating asm output, but it is helpful for analysis using `reccmp-roadmap`.

Method 2 was only applied for debug binaries (i.e. `BETA10`) because we assumed that binaries built with string pooling enabled would give each deduped string a symbol. This is generally true, but some strings are missed if we don't cover every address.

With this change, we apply method 2 for every binary. We take the following steps to not create invalid string entities:
1. Match strings in the Latin1[^1] character set, but only where the first character is regular ASCII.
2. Ignore addresses that are the site of a relocation. This helps to avoid vtable offsets referenced by an indirect call.
3. Run this step last and do not overwrite an existing entity. This helps to avoid situations where a variable entity is mistaken for a string.

The `match_string()` function is now run after this step so we can match the newly-created entities. To avoid a deluge of "failed to match" alerts (e.g. for code file paths used in `assert` calls) we only alert if a "verified" string in the original binary cannot be matched.

The biggest change is actually removing a lot of now unneeded code. The PE module identified possible strings in a roundabout way, and this new approach is better. The DB had a function specifically for matching strings, and it gave a specific error report if you tried to match a string entity with an existing variable. I think we avoid these problems with the above changes, but maybe I've overlooked something. (attn: @jonschz)

For testing, I have been diffing the output of `reccmp-roadmap --verbose` between master and this branch. I think we're in good shape and not dropping any strings that we should have. It isn't 100% perfect, though. For example, it creates an entity for this substring:

```diff
 0002:00000418   0002:0000042c         20  str  23      '<program name unknown>'
+0002:00000428                             str  7       'known>'
```

[^1]: I had forgotten why I chose Latin1 instead of just ASCII for the string encoding. It must have been done to support strings with the registered trademark symbol ® in `ISLE`.

### DIFF
--- a/reccmp/isledecomp/analysis/string_const.py
+++ b/reccmp/isledecomp/analysis/string_const.py
@@ -1,10 +1,10 @@
 import re
 
-r_likely_string = re.compile(r"([\t\r\n\x20-\x7f][\t\r\n\x20-\x7f\xa0-\xff]*)")
+r_likely_string = re.compile(r"[\t\r\n\x20-\x7f][\t\r\n\x20-\x7f\xa0-\xff]*")
 
 
 def is_likely_latin1(string) -> bool:
     """Heuristic to eliminate data streams that are not real strings.
-    We exclude bytes not in the Latin1 (ISO/IEC 8859-1) characte set
+    We exclude bytes not in the Latin1 (ISO/IEC 8859-1) character set
     and also assume the string begins with an ASCII character."""
     return r_likely_string.fullmatch(string) is not None
--- a/reccmp/isledecomp/analysis/string_const.py
+++ b/reccmp/isledecomp/analysis/string_const.py
@@ -1,0 +1,10 @@
+import re
+
+r_likely_string = re.compile(r"([\t\r\n\x20-\x7f][\t\r\n\x20-\x7f\xa0-\xff]*)")
+
+
+def is_likely_latin1(string) -> bool:
+    """Heuristic to eliminate data streams that are not real strings.
+    We exclude bytes not in the Latin1 (ISO/IEC 8859-1) characte set
+    and also assume the string begins with an ASCII character."""
+    return r_likely_string.fullmatch(string) is not None
--- a/reccmp/isledecomp/compare/core.py
+++ b/reccmp/isledecomp/compare/core.py
@@ -20,7 +20,11 @@ from reccmp.isledecomp.compare.event import (
     reccmp_report_nop,
     create_logging_wrapper,
 )
-from reccmp.isledecomp.analysis import find_float_consts, find_vtordisp
+from reccmp.isledecomp.analysis import (
+    find_float_consts,
+    find_vtordisp,
+    is_likely_latin1,
+)
 from .match_msvc import (
     match_lines,
     match_symbols,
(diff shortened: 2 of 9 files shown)


reccmp/isledecomp/analysis/string_const.py
```diff
@@ -0,0 +1,10 @@
+import re
+
+r_likely_string = re.compile(r"([\t\r\n\x20-\x7f][\t\r\n\x20-\x7f\xa0-\xff]*)")
+
+
+def is_likely_latin1(string) -> bool:
+    """Heuristic to eliminate data streams that are not real strings.
+    We exclude bytes not in the Latin1 (ISO/IEC 8859-1) characte set
```
jonschz:
```suggestion
    We exclude bytes not in the Latin1 (ISO/IEC 8859-1) character set
```

reccmp/isledecomp/analysis/string_const.py
```diff
@@ -0,0 +1,10 @@
+import re
+
+r_likely_string = re.compile(r"([\t\r\n\x20-\x7f][\t\r\n\x20-\x7f\xa0-\xff]*)")
```
jonschz: Do we need a match group around the string if we are matching the full regex anyway?

jonschz: Do we also want to match for the null terminator? Or do we have any / a significant amount of strings that are not null terminated?

reccmp/isledecomp/compare/core.py
```diff
@@ -447,60 +454,36 @@ def _add_match_in_array(
...
-                            type=EntityType.STRING,
-                            name=string,
-                            size=len(string) + 1,
-                        )
+                if is_likely_latin1(string):
+                    batch.insert_recomp(
+                        addr,
+                        type=EntityType.STRING,
+                        name=string,
+                        size=len(string) + 1,
```
jonschz: If this line stays that way: Add a comment that the +1 is to account for the null terminator

reccmp/isledecomp/compare/match_msvc.py
```diff
@@ -334,11 +335,12 @@ def match_strings(db: EntityDb, report: ReccmpReportProtocol = reccmp_report_nop
                 recomp_addr = string_index.pop(text)
                 batch.match(orig_addr, recomp_addr)
             else:
-                report(
-                    ReccmpEvent.NO_MATCH,
-                    orig_addr,
-                    msg=f"Failed to match string {repr(text)} at 0x{orig_addr:x}",
-                )
+                if verified:
```
jonschz: maybe use `elif`

reccmp/isledecomp/analysis/string_const.py
disinvite: Yes, but I would need to change `iter_string` and `read_string` from the PE module so that the terminator is retained and the string is still in `bytes` format. I'm going to address those in a future update (or updates) that should also cover #27.

reccmp/isledecomp/compare/core.py
disinvite: It will change, but we can add a comment for now.

PR #148, opened 2025-06-22
Convert the other `Compare` constructors to use `from_target()`.

disinvite:
I don't know why I skipped these in #143. This covers the remaining calls to the `Compare()` constructor in the Ghidra script, roadmap, and vtable tools.

My end goal here is to support situations where the code directory and PDB path are null. If we can handle those in one place, it will make it simpler to support projects with more or less data sources.

Remember to restart Ghidra before running the import script. It didn't detect the new `from_target()` method until I did that.

### DIFF
--- a/reccmp/ghidra_scripts/import_functions_and_types_from_pdb.py
+++ b/reccmp/ghidra_scripts/import_functions_and_types_from_pdb.py
@@ -317,23 +317,7 @@ def main():
         logging.getLogger("isledecomp.cvdump.symbols").setLevel(logging.WARNING)
 
     logger.info("Starting comparison")
-
-    origfile = detect_image(target.original_path)
-    if not isinstance(origfile, PEImage):
-        raise ValueError
-
-    recompfile = detect_image(target.recompiled_path)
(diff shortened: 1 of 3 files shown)


jonschz:
> Remember to restart Ghidra before running the import script. It didn't detect the new from_target() method until I did that.

Yes, that is a known issue which I discussed in the [Ghidrathon repo](https://github.com/mandiant/Ghidrathon/issues/103). I haven't gotten around to trying whether we still need Ghidrathon in the latest releases of Ghidra.

jonschz (approved):
Trusting you have tested all of these. The change looks sound. Thanks!

madebr (approved):

disinvite:
> > Remember to restart Ghidra before running the import script. It didn't detect the new from_target() method until I did that.
> 
> Yes, that is a known issue which I discussed in the [Ghidrathon repo](https://github.com/mandiant/Ghidrathon/issues/103). I haven't gotten around to trying whether we still need Ghidrathon in the latest releases of Ghidra.

I only point it out because _I_ forgot 😉

PR #153, opened 2025-06-26
Batch API: Skip updates where address is invalid

### LINKED ISSUE #152: `db.batch()` will create rows with no address
This test fails:
```python
with db.batch() as batch:
    batch.set_orig(None, name="Test")
    batch.set_recomp(None, name="Hello")
    batch.set_orig("Test", name="Test")
    batch.set_recomp("Hello", name="Hello")

assert db.count() == 0
```

In the case of entries with no address, an assert will fail when you call `get_all()`.

This manifested as a regression introduced by #151, but it is possible anywhere in the API. I think it was not caught by `mypy` because `recomp_addr` was returned by the result of another query.

Looks like we can easily check with `isinstance(addr, int)` before inserting. I assume `match()` is also affected.

disinvite:
Fixes #152. I also changed the query in `unique_names_for_overloaded_functions` so we can't get a null recomp address.

This closes the gap I used to trigger a SQL rollback, so I had to change two of the tests to cause a different exception.

### DIFF
--- a/reccmp/isledecomp/compare/db.py
+++ b/reccmp/isledecomp/compare/db.py
@@ -13,7 +13,6 @@ _SETUP_SQL = """
     CREATE TABLE entities (
         orig_addr int unique,
         recomp_addr int unique,
-        matched int as (orig_addr is not null and recomp_addr is not null),
         kvstore text default '{}'
     );
 
@@ -26,6 +25,14 @@ _SETUP_SQL = """
         SELECT recomp_addr, kvstore FROM entities
         WHERE recomp_addr is not null and orig_addr is null
         ORDER by recomp_addr;
+
+    -- ReccmpEntity
+    CREATE VIEW entity_factory (orig_addr, recomp_addr, kvstore) AS
+        SELECT orig_addr, recomp_addr, kvstore FROM entities;
+
+    -- ReccmpMatch
+    CREATE VIEW matched_entity_factory AS
+        SELECT * FROM entity_factory WHERE orig_addr IS NOT NULL AND recomp_addr IS NOT NULL;
 """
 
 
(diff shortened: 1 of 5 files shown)


reccmp/isledecomp/compare/db.py
```diff
@@ -277,12 +277,20 @@ def bulk_orig_insert(
                 """INSERT INTO entities (orig_addr, kvstore) values (?,?)
                 ON CONFLICT (orig_addr) DO UPDATE
                 SET kvstore = json_patch(kvstore, excluded.kvstore)""",
-                ((addr, json.dumps(values)) for addr, values in rows),
+                (
+                    (addr, json.dumps(values))
+                    for addr, values in rows
+                    if isinstance(addr, int)
+                ),
```
jonschz, line 291: Should we log a warning / error for all of these entries? With this change, they are dropped silently, which may not be what we want (applies to all instances)

jonschz (approved):
LGTM

reccmp/isledecomp/compare/db.py
disinvite: What if this were a fatal error instead? Not that those are popular with our users. 😀 But it would give the most important piece of information: the function that tried to set the bad data. We could do it easily by adding `check (orig_addr is not null and recomp_addr is not null)` to the `entities` table declaration.

We're a little stuck because the types coming back from `db.execute()` are all `Any`, so the `None` value can slip through as it did here. If we assert the datatype for each row column we're going to halt anyway, so same result as above.

tests/test_compare_db.py
```diff
@@ -322,9 +321,55 @@ def test_batch_sqlite_exception_insert_only(db):
...
+    with pytest.raises(OverflowError):
         batch.commit()
 
     assert db.get_by_orig(100) is None
+
+
+def test_batch_insert_with_invalid_address(db):
+    """Should not create (insert functions) an entity if the address parameter is invalid."""
+    with db.batch() as batch:
+        batch.insert_orig(None, name="Test")
```
jonschz, line 335: Now that I think about it again, I am genuinely a bit confused why `mypy` isn't complaining here

jonschz: Ah, here we go: It's the untyped `db` parameter, which is treated as an `Any` because our mypy config isn't exactly strict:
grafik

reccmp/isledecomp/compare/db.py
jonschz: Now that I saw that `mypy` can in fact catch this, I am wondering if this check is really worth it, or if we could just tighten our `mypy` setup instead. If you still think we should check the type, I suggest we throw in case the type is incorrect.

disinvite: Ideally we would get by with just `mypy`, but I don't see a way to get specific types for the tuples returned by `db.execute()` because the queries are not compiled until runtime. We should fix it for the tests, but we're still susceptible in the main code.

This relates to something you pointed out a while back that stuck with me. I think it was when I added the `match_msvc` module. It's better design to keep the SQL stuff confined to one module rather than having queries all over the place. I've been thinking about ways to move back towards the former while still doing what we need to do. If we had that, we could verify that you get good results (types) back from a limited set of functions.

jonschz: The way I'd probably approach this:
- Encapsulate all database calls (e.g. only call SQL commands from dedicated modules, both read and write)
- Implement sufficiently strict type signatures in those modules, especially for the writing functions
- Validate the types at runtime when reading from the DB (just to be sure)

PR #154, opened 2025-06-26
Use SQL view for `ReccmpEntity` and `ReccmpMatch` objects

disinvite:
Small refactor to allow for some experiments with these ORM objects.

I also eliminated the `_get_closest_orig` and `_get_closest_recomp` functions and integrated their queries into the `get_by_orig` and `get_by_recomp` functions.

### DIFF
--- a/reccmp/isledecomp/compare/db.py
+++ b/reccmp/isledecomp/compare/db.py
@@ -375,45 +375,23 @@ class EntityDb:
         cur.row_factory = matched_entity_factory
         return cur.fetchone()
 
-    def _get_closest_orig(self, addr: int) -> int | None:
-        for (value,) in self._sql.execute(
-            "SELECT orig_addr FROM entities WHERE ? >= orig_addr ORDER BY orig_addr desc LIMIT 1",
-            (addr,),
-        ):
-            return value
+    def get_by_orig(self, addr: int, exact: bool = True) -> ReccmpEntity | None:
+        if exact:
+            query = "SELECT * FROM entity_factory WHERE orig_addr = ?"
+        else:
+            query = "SELECT * FROM entity_factory WHERE ? >= orig_addr ORDER BY orig_addr desc LIMIT 1"
 
-        return None
-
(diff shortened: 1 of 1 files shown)


reccmp/isledecomp/compare/db.py
```diff
@@ -347,79 +355,50 @@ def get_unmatched_strings(self) -> list[str]:
...
-            return value
-
-        return None
-
-    def get_by_orig(self, orig: int, exact: bool = True) -> ReccmpEntity | None:
-        addr = self._get_closest_orig(orig)
-        if addr is None or exact and orig != addr:
-            return None
+    def get_by_orig(self, addr: int, exact: bool = True) -> ReccmpEntity | None:
+        if exact:
```
jonschz, line 378: Not related to your change, but maybe we want to add a docstring here to explain the semantics of `exact`? While it is intuitive from reading the implementation, the signature is not. Also, it may make sense to make `exact` a kw-only argument, since `get_by_orig(1234, exact=false)` makes it less likely that a reader needs to go into the documentation/implementation of `get_by_orig` to understand the arguments.

reccmp/isledecomp/compare/db.py
```diff
@@ -433,10 +412,9 @@ def get_lines_in_recomp_range(
         """Fetches all matched annotations of the form `// LINE: TARGET 0x1234` in the given recomp address range."""
 
         cur = self._sql.execute(
-            """SELECT orig_addr, recomp_addr, kvstore FROM entities
+            """SELECT * FROM matched_entity_factory
             WHERE json_extract(kvstore, '$.type') = ?
             AND recomp_addr >= ? AND recomp_addr <= ?
-            AND matched = 1
```
jonschz, line 439: I think we are no longer using the `matched` column? (line 16)

reccmp/isledecomp/compare/db.py
disinvite: Yes. Making those changes drags in more files, so we'll do it in its own PR. #155

PR #156, opened 2025-06-28
Improve documentation for 'exact' parameter in address lookups

### LINKED ISSUE #155: Better documentation for DB addr lookup functions
Not related to your change, but maybe we want to add a docstring here to explain the semantics of `exact`? While it is intuitive from reading the implementation, the signature is not. Also, it may make sense to make `exact` a kw-only argument, since `get_by_orig(1234, exact=false)` makes it less likely that a reader needs to go into the documentation/implementation of `get_by_orig` to understand the arguments.

_Originally posted by @jonschz in https://github.com/isledecomp/reccmp/pull/154#discussion_r2170805707_

disinvite:
Fixes #155. The code reads better with `exact` as a required keyword arg.

The behavior of `exact` was not covered thoroughly or directly by the existing tests, so I added two new ones.

### DIFF
--- a/tests/test_compare_db.py
+++ b/tests/test_compare_db.py
@@ -72,6 +72,63 @@ def test_db_all_order(db):
     ]
 
 
+def test_get_by_exact(db):
+    """get_by_orig and get_by_recomp have two parameters: the address and the 'exact' option.
+    If exact=False, we return the entity at the address OR one at the preceding address if it exists.
+    """
+
+    with db.batch() as batch:
+        batch.set_orig(100)
+        batch.set_recomp(100)
+
+    # If there is an exact addr match, return the entity.
+    assert db.get_by_orig(100) is not None
+    assert db.get_by_orig(100, exact=True) is not None
+    assert db.get_by_orig(100, exact=False) is not None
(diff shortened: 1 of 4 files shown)


tests/test_compare_db.py
```diff
@@ -72,6 +72,63 @@ def test_db_all_order(db):
...
+        db.get_by_orig(100, False)
+
+    with pytest.raises(TypeError):
+        db.get_by_orig(100, True)
+
+    with pytest.raises(TypeError):
+        db.get_by_recomp(100, True)
+
+    with pytest.raises(TypeError):
+        db.get_by_recomp(100, False)
```
jonschz: This should be caught by `mypy` if you provide a type for the `db` argument, but it doesn't hurt to have it here

jonschz (approved):
Nice!

tests/test_compare_db.py
disinvite: In case we forget why this is there if/when we add the `db` types later, what do you think about adding it now for this function only, along with `# type: ignore` for the 4 calls?

jonschz: Sounds good 👍

PR #157, opened 2025-06-29
Improve line pinning

jonschz:
Two new features:
- The `// LINE` annotation now has a more readable output. Pinned Lines are no longer in its own diff group, but are integrated into the existing ones. A hint that the line is pinned is added at the end. The code has also become cleaner with better separation of concerns.
- All lines we know from the decomp are now printed in the diff.

For example:
```diff
0x100721d4 : mov eax, dword ptr [esp + 0x18]    (act3.cpp:224)
+           : +mov edx, dword ptr [esi + 4]
+           : +mov ecx, dword ptr [eax + 4]
+           : +mov dword ptr [esp + 0x18], ecx
+           : +cmp dword ptr [edx], ecx
+           : +jne -0x6b
-0x100721d8 : -cmp byte ptr [eax + 8], 0
+           : +cmp byte ptr [edi + 8], 0         (act3.cpp:248, pinned)
0x100721dc : jne 0x1a
-0x100721de : -mov byte ptr [eax + 8], 1
+           : +mov byte ptr [edi + 8], 1         (act3.cpp:250, pinned)
0x100721e2 : push 0     (act3.cpp:252, pinned)
```

### DIFF
--- a/reccmp/isledecomp/compare/functions.py
+++ b/reccmp/isledecomp/compare/functions.py
@@ -5,7 +5,7 @@ import struct
 from itertools import pairwise
 from typing import Callable, Iterator, NamedTuple
 from reccmp.isledecomp.compare.lines import LinesDb
-from reccmp.isledecomp.compare.pinned_sequences import match_sequences_with_pins
+from reccmp.isledecomp.compare.pinned_sequences import SequenceMatcherWithPins
 from reccmp.isledecomp.compare.asm.fixes import assert_fixup, find_effective_match
 from reccmp.isledecomp.compare.asm.parse import AsmExcerpt, ParseAsm
 from reccmp.isledecomp.compare.asm.replacement import create_name_lookup
@@ -217,12 +217,12 @@ class FunctionComparator:
         orig_asm = [x[1] for x in orig]
         recomp_asm = [x[1] for x in recomp]
 
-        diff = match_sequences_with_pins(orig_asm, recomp_asm, split_points)
+        diff = SequenceMatcherWithPins(orig_asm, recomp_asm, split_points)
 
-        if diff.match_ratio != 1.0:
+        if diff.ratio() != 1.0:
             # Check whether we can resolve register swaps which are actually
             # perfect matches modulo compiler entropy.
-            is_effective = find_effective_match(diff.opcodes, orig_asm, recomp_asm)
+            is_effective = find_effective_match(diff.get_opcodes(), orig_asm, recomp_asm)
 
             # Convert the addresses to hex string for the diff output
             orig_for_printing = [
@@ -244,7 +244,7 @@ class FunctionComparator:
             ]
 
             unified_diff = combined_diff(
-                diff.opcode_groups,
+                diff.get_grouped_opcodes(),
                 orig_for_printing,
                 recomp_for_printing,
             )
@@ -255,7 +255,7 @@ class FunctionComparator:
         return FunctionCompareResult(
             unified_diff,
             is_effective,
-            diff.match_ratio,
+            diff.ratio(),
         )
 
     def _collect_line_annotations(self, recomp: AsmExcerpt) -> list[ReccmpMatch]:
--- a/reccmp/isledecomp/compare/functions.py
+++ b/reccmp/isledecomp/compare/functions.py
@@ -1,11 +1,10 @@
 from datetime import datetime
 from dataclasses import dataclass
-from difflib import SequenceMatcher
 from functools import cache
-import functools
 import struct
 from itertools import pairwise
 from typing import Callable, Iterator, NamedTuple
+from reccmp.isledecomp.compare.pinned_sequences import match_sequences_with_pins
 from reccmp.isledecomp.compare.asm.fixes import assert_fixup, find_effective_match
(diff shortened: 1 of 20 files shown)


reccmp/isledecomp/compare/pinned_sequences.py
```diff
@@ -0,0 +1,135 @@
...
+class _IntermediateSequenceMatch(NamedTuple):
+    opcodes: list[tuple[str, int, int, int, int]] = []
+    total_lines: int = 0
+    weighted_match_ratio: float = 0.0
+
+
+# TODO: Unit tests
+# - Compare with and without pins
+# - edge cases: first and/or last line are already pinned
+
```
jonschz: Address TODO / write tests

tests/test_function_comparator.py
```diff
@@ -230,24 +246,17 @@ def test_impact_of_line_annotation(db: EntityDb, report: ReccmpReportProtocol):
...
+                        ),
                     ],
-                }
-            ],
-        ),
-        (
-            "@@ -0x222,9 +0x40a,18 @@",
-            [
+                },
                 {
```
jonschz, line 260: Note the change both in the structure and in the added `(test.cpp:123, pinned)`

disinvite:
Very nice! Some first impressions after playing around for a bit:
- I think the code for `accumulator` would be easier to read (and maybe shorter) without doing it as a reducer. Could that just be its own function (or a static method of the class) that gives the results for the one chunk? You wouldn't need to reference `pinned_lines` because you could raise whatever exception in the method and then catch it in the caller. You could use the `_IntermediateSequenceMatch` tuple as the return type.
- I haven't looked back at `difflib.py` file to confirm, but I think their diff ratio is influenced by the number of characters in the lines rather than just `lines_matched / total_lines`. If that's right, do you think it's worth doing our own ratio calculation instead? That might help eliminate diff noise to some small degree. For example: if an address changes from `<OFFSET1>` to `some_longer_string (FUNCTION)` but it is still a diff because of the register used.
- Since we're using the type hint `tuple[str, int, int, int, int]` in more places, is it worth defining that as its own type? Actually... do we have it already? I remember seeing it but maybe that was one of my own test files.

disinvite:
If we _do_ change the ratio calc that should maybe be its own PR because it will create the appearance of a diff in the CI results.

jonschz:
> * I haven't looked back at `difflib.py` file to confirm, but I think their diff ratio is influenced by the number of characters in the lines rather than just `lines_matched / total_lines`. If that's right, do you think it's worth doing our own ratio calculation instead? That might help eliminate diff noise to some small degree. For example: if an address changes from `<OFFSET1>` to `some_longer_string (FUNCTION)` but it is still a diff because of the register used.

I haven't tested this thoroughly, but I think I saw that adding pins to a line that already matched didn't affect the match percentage, suggesting that the current way is at least close to what `difflib` does. I'll take a closer look, you make a good point regarding the noise.

jonschz:
> I think the code for accumulator would be easier to read (and maybe shorter) without doing it as a reducer.

I have changed the `accumulator` code.

> I haven't looked back at difflib.py file to confirm, but I think their diff ratio is influenced by the number of characters in the lines rather than just lines_matched / total_lines. 

It looks like `SequenceMatcher` only cares about whether lines match or not:
```py
>>> from difflib import SequenceMatcher
>>> a = ["0", "abcdefg", "1"]
>>> b = ["0", "abcdefh", "1"]
>>> c = ["0", "wxyz", "1"]
>>> SequenceMatcher(None, a, b).ratio()      
0.6666666666666666
>>> SequenceMatcher(None, a, c).ratio() 
0.6666666666666666
```
Do you want me to investigate more whether (correct) pinning affects the match ratio?

> Since we're using the type hint tuple[str, int, int, int, int] in more places, is it worth defining that as its own type? 

Good catch!

reccmp/isledecomp/compare/lines.py
```diff
@@ -112,3 +117,6 @@ def find_function(
             line_start,
         )
         return None
+
+    def find_line_of_recomp_address(self, address: int) -> tuple[PurePath, int] | None:
```
disinvite, line 121: We should probably have a couple of tests on this function to protect against changes to `LinesDb`

reccmp/isledecomp/compare/functions.py
```diff
@@ -154,8 +167,6 @@ def compare_function(self, match: ReccmpMatch) -> DiffReport:
         orig_combined = self.orig_sanitize.parse_asm(orig_raw, match.orig_addr)
         recomp_combined = self.recomp_sanitize.parse_asm(recomp_raw, match.recomp_addr)
 
-        total_lines = len(orig_combined) + len(recomp_combined)
-
         if self.debug:
             self._dump_asm(orig_combined, recomp_combined)
```
disinvite, line 171: Not something we need to address for this PR because this is just for debugging anyway, but the later modifications to the asm output (i.e. adding the file/line number from the pins) are not represented here.

disinvite (approved):

reccmp/isledecomp/compare/functions.py
jonschz: I thought about that too and didn't include it on purpose in case the user wants to diff the orig and recomp assembly with a different tool. That would create lots of false positives.

PR #158, opened 2025-06-29
Refactor project loading

disinvite:
Right now we have two dataclasses that represent a project:
- `RecCmpProject`, used only with the `reccmp-project` tool. Reads `reccmp-project.yml` and `reccmp-user.yml` from the same directory, but not `reccmp-build.yml`.
- `RecCmpBuiltProject`, used in the other tools. Starts by loading `reccmp-build.yml`, then loads the other two files from the directory specified in the `project` attribute.

Since these are very similar, I had the idea of combining them to one dataclass: `RecCmpProject`. The loading steps for the new `RecCmpProject.from_directory()` are:

1. Look for `reccmp-build.yml` in the starting directory or its parent directories.
2. If we found the build file, read the `project` directory and look there for `reccmp-project.yml`. If it exists, jump to step 4. If not, fail with an error.
3. If there is no build file, look for `reccmp-project.yml` in the starting directory or its parent directories. If we don't find it, fail with an error.
4. Look for `reccmp-user.yml` in the same directory as the one where we loaded `reccmp-project.yml`. If it is there, load it.

Only `reccmp-project.yml` is "required" because it defines the targets. In practice, though, we can't run any analysis without having all three files. None of the three schemas overlap, so each file provides only its own information and we need all of it.

Right now, all the tools assume that a target has complete information. To solve the type-checking concerns, we now have two dataclasses for targets. They have the same attributes but with different types. `RecCmpPartialTarget` is for building up information in steps, and some of the fields can be null. `RecCmpTarget` is intended for "completed" targets. You call the `get()` method on a project with the target name to get a `RecCmpTarget` back.

I added some tests for the scenarios where you load a project and only some of the files exist. There is some definite boilerplate code here for creating each of the files, so we could condense this if it doesn't hurt the readability. I might have missed some edge cases so any suggestions for new tests are welcome!

There is a new error message if you try to run with an incomplete target that tells which fields are missing:
```
Warning:  User config reccmp-user.yml is missing target ids: BETA10,ALPHA
Warning:  Build config reccmp-build.yml is missing target ids: BETA10,ALPHA
```

I dropped two `argparse` functions that were not used anywhere. I think these were intended for loading in the `reccmp-project` tool.

The `--paths` command line option now assumes the module name is the base filename in upper-case. (e.g. for `lego1.dll`, it is `LEGO1`.) We had previously set it to `None` and made the same assumption later in the `Compare` constructor. I left that code alone for now so that (my) test scripts that call this function directly will keep working.

### DIFF
--- a/reccmp/isledecomp/compare/asm/fixes.py
+++ b/reccmp/isledecomp/compare/asm/fixes.py
@@ -193,8 +193,11 @@ def _is_relocatable(instr: str) -> bool:
     Excludes certain instructions whose relocation will always change the logic
     to be considered for an effective match.
     """
-    # Do not relocate jump table entries (this most likely influences the behaviour)
     if instr.startswith("start +"):
+        # Do not relocate jump table entries (this most likely influences the behaviour)
+        return False
+    if instr.startswith("0x"):
+        # Do not relocate data table entries (this most likely influences the behaviour)
         return False
     return True
 
--- a/reccmp/isledecomp/compare/functions.py
+++ b/reccmp/isledecomp/compare/functions.py
@@ -252,7 +252,7 @@ class FunctionComparator:
             ]
 
             unified_diff = combined_diff(
-                diff.get_grouped_opcodes(),
+                diff.get_grouped_opcodes(n=10),
                 orig_for_printing,
                 recomp_for_printing,
             )
--- a/tests/test_function_comparator.py
+++ b/tests/test_function_comparator.py
@@ -427,10 +427,15 @@ def test_jump_table_wrong_order(
     Jump tables with the correct entries in the wrong order.
     In particular, this must not become an accidental effective match.
     """
-    orig = b"\xff\x24\x85\x07\x02\x00\x00\x33\x04\x00\x00\x43\x04\x00\x00"
-    recm = b"\xff\x24\x85\x07\x04\x00\x00\x43\x06\x00\x00\x33\x06\x00\x00"
+    orig = (
+        # jmp dword ptr [eax * 4 + $jump_table]
+        b"\xff\x24\x85\x07\x02\x00\x00"
+        # $jump_table:
+        + b"\x33\x04\x00\x00\x43\x04\x00\x00"
+    )
+    recm = b"\xff\x24\x85\x07\x04\x00\x00" + b"\x43\x06\x00\x00\x33\x06\x00\x00"
 
-    # Required to get an `<OFFSET1> into the jump instruction`
+    # Required to get an `<OFFSET1>` into the jump instruction
     is_relocated_addr = Mock(return_value=True)
     diffreport = compare_functions(db, lines_db, orig, recm, report, is_relocated_addr)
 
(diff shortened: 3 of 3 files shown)


jonschz:
Just to be sure: This changes nothing about the project file structure we discussed in #7, right? Will that make the potential future structure change easier, harder, or no impact at all?

reccmp/project/detect.py
```diff
@@ -71,213 +86,238 @@ class GhidraConfig:
 
 @dataclass
 class RecCmpTarget:
+    # pylint: disable=too-many-instance-attributes
     """Partial information for a target (binary file) in the decomp project
     This contains only the static information (same for all users).
     Saved to project.yml. (See ProjectFileTarget)"""
```
jonschz: Not your change, but this docstring is outdated. `project.yaml` doesn't exist, and even if it were `reccmp-project.yaml`, I think this class is also used for the other files.

reccmp/project/detect.py
```diff
@@ -71,213 +86,238 @@ class GhidraConfig:
 
 @dataclass
 class RecCmpTarget:
```
jonschz: Just a suggestion: How about renaming
RecCmpTarget -> PartialReccmpTarget
RecCmpBuiltTarget -> ReccmpTarget
?

reccmp/project/detect.py
```diff
@@ -71,213 +86,238 @@ class GhidraConfig:
...
+                "original_path",
+                "recompiled_path",
+                "recompiled_pdb",
+            ):
+                if getattr(target, attr) is None:
+                    missing.append(attr)
+
+            raise IncompleteReccmpTargetError(
+                f"Target {target_id} is missing data: {','.join(missing)}"
+            ) from ex
```
jonschz: Minor stylistic point: Maybe it's better to do this check first. Then you can be sure that the assertions don't fail, and you don't need to catch an assertion error. The KeyError might also be preventable, maybe by something like `self.targets.get(target_id, None); if target_id is None: [error handling]`

reccmp/project/detect.py
```diff
@@ -71,213 +86,238 @@ class GhidraConfig:
...
-            raise InvalidRecCmpProjectException(
-                f"{build_config}: .project is invalid ({project_config_path} does not exist)"
-            )
-        logger.debug("Using project config: %s", project_config_path)
+    def from_directory(cls, directory: Path) -> "RecCmpProject":
+        project = cls()
 
-        # Parse project.yml
-        project_data = ProjectFile.from_file(project_config_path)
+        # Searching for build.yml
```
jonschz:
```suggestion
        # Searching for reccmp-build.yml
```

reccmp/project/detect.py
```diff
@@ -71,213 +86,238 @@ class GhidraConfig:
...
 
-        # Searching for user.yml
-        user_config = project_directory / RECCMP_USER_CONFIG
-        if not user_config.is_file():
-            raise InvalidRecCmpProjectException(
-                f"Missing {RECCMP_USER_CONFIG}. First run 'reccmp-project detect'."
-            )
-        logger.debug("Using user config: %s", user_config)
+        if project_data is None:
+            raise InvalidRecCmpProjectException(f"No project file in path: {directory}")
```
jonschz:
```suggestion
            raise RecCmpProjectNotFoundException(f"No project file in path: {directory}")
```
I noticed this because it broke my local Ghidra setup. In case you don't want to accept this change, the Ghidra script needs a small change in order to work (in the `dev_config.json` code path)

jonschz (approved):
Looks good!

reccmp/project/detect.py
disinvite: Thanks! I think it reads better now.

reccmp/project/detect.py
disinvite: I thought about adding that here, but it drags in a lot of files. Doing it properly means removing "built" from the function names, so that's almost every file in the `tools/` path. We can do it now if you don't think it would clash with the other PRs in flight.

disinvite:
> Just to be sure: This changes nothing about the project file structure we discussed in #7, right? Will that make the potential future structure change easier, harder, or no impact at all?

No change to the file schemas yet. We can support the current format for a little while longer if we only add new attributes.

It should hopefully be easier to make changes with this new setup. The path that begins with loading the files and ends with running the analysis is more well defined.

reccmp/project/detect.py
jonschz: All of the current PRs are by one of us, and even if there are merge conflicts, they should be easy to resolve. So I'd go for it.

madebr:
> @madebr You came up with the original project design. Any thoughts on these changes before we merge it?

If I'm reading this PR correctly, the principles remain the same.
You formalized the search procedure a bit, which makes sense to me.

Issue #47, opened 2024-12-29
Revamp the HTML interface


disinvite:
We haven't done anything to the HTML output file from `reccmp` in a while. What I think would help is more modularization and the ability to run CI tests. In lieu of an actual JS framework, I used the [Web components API](https://developer.mozilla.org/en-US/docs/Web/API/Web_components) a while back but I didn't find it all that great to work with.

The obstacle with having a build process for the HTML/JS is where to put the source files. We definitely don't want to require users to install nodejs just to use `reccmp`, so I think that means building externally and loading just the artifact files into the package. What's the best way to do that? Build using some framework and then commit a non-minified file here to the main repo?

The other question is what framework to use, if any.

disinvite:
I would call this done after #238. We have a path forward: keep using web components for now but keep the door open to adding a framework, whatever that ends up being.

PR #160, opened 2025-07-02
Add some e2e tests for HTML display

disinvite:
The first step to improving our HTML reports (#47) is to validate what we already have. The web components aren't isolated enough (yet) to support unit tests, so I wrote some end-to-end tests with Playwright. This doesn't cover 100% of the functionality but it does test the majority of the user interactions, enough that I think we can start making changes.

To create the HTML report file, I used `reccmp-aggregate` with some dummy data in `testdata.json`. This file is aggregated against itself (because two sample files are required) but in effect this creates the HTML file using JSON as the data source.

I went with Biome for javascript linting and formatting, with some changes to the default config. This isn't enabled for the actual code in `reccmp.js` because I wanted to get the tests established first.

Thoughts on shoving all this javascript stuff alongside our python libray? Playwright does have a Python module but I thought it best to keep the worlds separate.

In the CI action, Playwright needs to download browser binaries to run the tests. This takes about 90-120 seconds. They caution against caching these files because they claim it isn't any faster (probably right). Should we configure this test so it only runs when JS files are changed?

### DIFF
--- a/.github/workflows/playwright.yml
+++ b/.github/workflows/playwright.yml
@@ -1,0 +1,48 @@
+name: Playwright Tests
+
+on: [push, pull_request]
+
+jobs:
+  test:
+    timeout-minutes: 60
+    runs-on: ubuntu-latest
+
+    steps:
+    - uses: actions/checkout@v4
+
+    - uses: actions/setup-python@v5
+      with:
+        python-version: '3.12'
+
+    - name: Install reccmp
+      shell: bash
+      run: |
+        pip install .
+
+    - name: Create HTML file
+      shell: bash
+      run: |
+        reccmp-aggregate --samples ./tests/js/testdata.json ./tests/js/testdata.json --html ./tests/js/index.html
+
+    - uses: actions/setup-node@v4
+      with:
+        node-version: lts/*
+
+    - name: Install dependencies
+      run: |
+        npm ci
+
+    - name: Install Playwright Browsers
+      run: |
+        npx playwright install --with-deps
+
+    - name: Run Playwright tests
+      run: |
+        npx playwright test
+
+    - uses: actions/upload-artifact@v4
+      if: ${{ !cancelled() }}
+      with:
+        name: playwright-report
+        path: playwright-report/
+        retention-days: 30
--- a/.github/workflows/format.yml
+++ b/.github/workflows/format.yml
@@ -30,3 +30,18 @@ jobs:
     - name: Run mypy
       shell: bash
       run: mypy ./reccmp ./tests
+
+  js-format:
+    name: 'Javascript Format'
+    runs-on: ubuntu-latest
+
+    steps:
+    - uses: actions/checkout@v4
+
+    - name: Install dependencies
+      run: |
+        npm ci
+
+    - name: Biome Format
+      run: |
+        npx @biomejs/biome ci ./tests/js/
--- a/.gitignore
+++ b/.gitignore
@@ -13,3 +13,10 @@ dist
 # General
 dev_config.json
 ghidra_import.log
+
+# Playwright
+node_modules/
+/test-results/
+/playwright-report/
+/blob-report/
+/playwright/.cache/
--- a/biome.json
+++ b/biome.json
@@ -1,0 +1,36 @@
+{
+	"$schema": "https://biomejs.dev/schemas/2.0.6/schema.json",
+	"vcs": {
+		"enabled": false,
+		"clientKind": "git",
+		"useIgnoreFile": false
+	},
+	"files": {
+		"ignoreUnknown": false
+	},
+	"formatter": {
+		"enabled": true,
+		"lineWidth": 120,
+		"indentStyle": "space",
+		"indentWidth": 2
+	},
+	"linter": {
+		"enabled": true,
+		"rules": {
+			"recommended": true
+		}
+	},
+	"javascript": {
+		"formatter": {
+			"quoteStyle": "single"
+		}
+	},
+	"assist": {
+		"enabled": true,
+		"actions": {
+			"source": {
+				"organizeImports": "on"
+			}
+		}
+	}
+}
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,0 +1,261 @@
+{
+  "name": "reccmp-js-tests",
+  "version": "1.0.0",
+  "lockfileVersion": 3,
+  "requires": true,
+  "packages": {
+    "": {
+      "name": "reccmp-js-tests",
+      "version": "1.0.0",
+      "license": "AGPL-3.0-or-later",
+      "devDependencies": {
+        "@biomejs/biome": "2.0.6",
+        "@playwright/test": "^1.53.2",
+        "@types/node": "^24.0.8"
+      }
+    },
+    "node_modules/@biomejs/biome": {
+      "version": "2.0.6",
+      "resolved": "https://registry.npmjs.org/@biomejs/biome/-/biome-2.0.6.tgz",
+      "integrity": "sha512-RRP+9cdh5qwe2t0gORwXaa27oTOiQRQvrFf49x2PA1tnpsyU7FIHX4ZOFMtBC4QNtyWsN7Dqkf5EDbg4X+9iqA==",
+      "dev": true,
+      "license": "MIT OR Apache-2.0",
+      "bin": {
+        "biome": "bin/biome"
+      },
+      "engines": {
+        "node": ">=14.21.3"
+      },
+      "funding": {
+        "type": "opencollective",
+        "url": "https://opencollective.com/biome"
+      },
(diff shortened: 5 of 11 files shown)


jonschz:
> Thoughts on shoving all this javascript stuff alongside our python libray? Playwright does have a Python module but I thought it best to keep the worlds separate.

I haven't tried the Playwright/Python module, but my first instinct is the same as yours - keep both worlds separate. Question is whether we want to have all the js project files on the root level or if we want to isolate them to a subfolder (maybe `/playwright`?) This may make the separation clearer and be less confusing to a new user whether this is a Python or a JS project. I'd be in favour of the latter.

> I went with Biome for javascript linting and formatting, with some changes to the default config. This isn't enabled for the actual code in reccmp.js because I wanted to get the tests established first.

Sounds reasonable. I'm more familiar with the established Prettier + ESLint stack but willing to give this one a shot.

> In the CI action, Playwright needs to download browser binaries to run the tests. This takes about 90-120 seconds. They caution against caching these files because they claim it isn't any faster (probably right). Should we configure this test so it only runs when JS files are changed?

Good question. Sounds like a good idea, but I'm not sure how stable it is or how difficult it is to set up properly. Don't forget to include TS files as well in case we want to migrate to Typescript.

webui/playwright.config.js
```diff
@@ -0,0 +1,87 @@
...
+
+    {
+      name: 'firefox',
+      use: { ...devices['Desktop Firefox'] },
+    },
+
+    {
+      name: 'webkit',
+      use: { ...devices['Desktop Safari'] },
+    },
```
jonschz, line 60: Have you tried whether it installs any faster if we only use one browser? Since the JS we use is quite basic, checking one of them would be enough for now IMO.

webui/e2e/reccmp.spec.js
```diff
@@ -0,0 +1,406 @@
+import { expect, test } from '@playwright/test';
+
+test.beforeEach(async ({ page }) => {
+  await page.goto('');
+});
+
+test.describe('Table display options', () => {
+  test('Hide 100%', async ({ page }) => {
+    // TODO: Should use 'cell' role
```
jonschz: Address TODOs (multiple instances). Question is if you want to pull out a few follow-up tasks, or maybe just a task like _Clean up TODOs in Playwright tests_ if you want to have something merged now.

webui/playwright.config.js
```diff
@@ -0,0 +1,87 @@
...
+ */
+// import dotenv from 'dotenv';
+// import path from 'path';
+// dotenv.config({ path: path.resolve(__dirname, '.env') });
+
+/**
+ * @see https://playwright.dev/docs/test-configuration
+ */
+
+const indexPath = path.resolve('./index.html');
```
jonschz: I'd flip around the `const` and the comment because the comment belongs to the `export default defineConfig()`

.github/workflows/playwright.yml
jonschz, line 1: I  think we should add at least a minimal README on how to get the Playwright tests to run locally. It can be reconstructed from the CI script, but it'll be hard to get it running without it.

webui/e2e/issue99.spec.js
jonschz, line 1: Nice!

webui/e2e/reccmp.spec.js
```diff
@@ -0,0 +1,406 @@
...
+
+    // Uncheck the box.
+    await checkbox.click();
+    await expect(checkbox).not.toBeChecked();
+
+    // The rows should return.
+    await expect(stubs).not.toHaveCount(0);
+  });
+
+  test('Show recomp', async ({ page }) => {
```
jonschz: I'd also check for one pair of values in this test

webui/e2e/reccmp.spec.js
```diff
@@ -0,0 +1,406 @@
...
+    // The text we want copied.
+    const addr = '0x401000';
+
+    // TODO: this is a convoluted way to avoid other page elements
+    await page.getByText(addr, { exact: true }).filter({ visible: true }).click();
+
+    // Get the value from the clipboard and confirm that it matches the address.
+    const handle = await page.evaluateHandle(() => navigator.clipboard.readText());
+    const clipboardContent = await handle.jsonValue();
+    expect(clipboardContent).toEqual(addr);
```
jonschz: Have you tried to access the clipboard directly from the Node.js Playwright runner instead of accessing it through the browser? Maybe that also fixes the permissions issue.

webui/e2e/reccmp.spec.js
jonschz, line 1: I think it'd be better if you split this file by the different `describe` sections. It is quite long already.

webui/e2e/reccmp.spec.js
```diff
@@ -0,0 +1,406 @@
...
+    // Name text to click that toggles the diff row.
+    const nameLink = page.locator('func-row').getByText('IsleApp::IsleApp');
+
+    // There are none to start
+    await expect(page.locator('diff-row')).toHaveCount(0);
+
+    // Create diff-row element
+    await nameLink.click();
+    await expect(page.locator('diff-row')).toHaveCount(1);
+
```
jonschz: Maybe also test for a `Identical function - no diff` string to be visible?

webui/e2e/reccmp.spec.js
```diff
@@ -0,0 +1,406 @@
...
+    await expect(searchbox.and(asmPlaceholder)).not.toBeAttached();
+
+    // Same behavior for asm diff option
+    await page.getByRole('radio', { name: 'Asm diffs only' }).click();
+    await expect(searchbox.and(namePlaceholder)).not.toBeAttached();
+    await expect(searchbox.and(asmPlaceholder)).toBeAttached();
+  });
+});
+
+test.describe('Diff rows', () => {
```
jonschz: If you want, you can also add a test that opens a non-trivial diff and verifies that some diff lines are visible (at least by diff text)

webui/playwright.config.js
disinvite: It would be faster, but we should still test all three. Early on we had a case where the page didn't work at all on Safari.

webui/e2e/reccmp.spec.js
disinvite: We'll do a follow-up task for these because it involves refactoring the components. This is ultimately a minor thing because the tests work fine, but these test suites encourage you to use roles in the query selectors wherever possible.

webui/e2e/reccmp.spec.js
disinvite: I think we can leave this as a follow-up issue. It's not an essential feature and we at least have partial testing for it. There are a number of github issues for playwright related to clipboard stuff, including requests for a better API than the `evaluate` call.

disinvite:
@jonschz Do you want to give the readme and other new commits a final look? Thanks!

webui/e2e/reccmp.spec.js
jonschz: We can also leave it like that if there is no easy, better solution right now. No need to open a follow-up. I was just curious if you had considered the option.

webui/README.md
```diff
@@ -0,0 +1,29 @@
...
+3. Using the completed analysis run, create a serialized report file in JSON format, then insert its contents in a second `<script>` element in `template.html`. The result is that the data is assigned to the `report` variable.
+The `pystache` module is used to combine these elements into the final output file.
+
+## Setup
+The tools in this directory require [node.js](https://nodejs.org/). The `package.json` file lists the dependencies and you can use your package manager of choice to install them. For example: `npm install`.
+
+## Formatting and linting
+[Biome](https://biomejs.dev/) validates the test files and javascript source. You can run it yourself with the command: `npx @biomejs/biome ci <list-of-paths>`.
+
+The CI scripts run from the `webui` directory and check these paths: `npx @biomejs/biome ci e2e ./testdata.json`.
```
jonschz: Now that I look at it again, it would probably be cleaner to add scripts to the package.json for linting and for running the tests, and then executing them via `npm run lint` or `npm run test`, both here and in the CI scripts. We can change this as a drive-by in a later PR, though.

webui/testdata.json
```diff
@@ -5,49 +5,49 @@
...
-      "recomp": "0x401000"
+      "recomp": "0x501000"
     },
     {
       "address": "0x4011a0",
       "diff": [],
       "matching": 1.0,
       "name": "IsleApp::~IsleApp",
-      "recomp": "0x4011a0"
+      "recomp": "0x5011a0"
```
jonschz: It's nice to have separate addresses!

jonschz (approved):
Very nice! I've left one comment. It's up to you if you want to address it now or later.

webui/README.md
disinvite: I had that thought too. Let's do it now.

PR #161, opened 2025-07-02
Fix #159 (`EntityBatch` type)

### LINKED ISSUE #159: `EntityBatch` not typed properly
Related to type issues in #152.

Adding the `reveal_type` helper on an `EntityBatch` reference:

```py
    reveal_type(batch)

```

Shows this in `mypy`:
```
reccmp\isledecomp\compare\core.py:732: note: Revealed type is "Any"
```

which is not what we want. I think the answer is to change the return type on `batch()` to `typing.ContextManager[EntityBatch]`, but it also needs to work in these situations:

```py
batch = self._db.batch()
# some stuff
batch.commit()
```

disinvite:
Not sure what test (if any) would fit here. This is more of a fix for `mypy` than an actual bug.

### DIFF
--- a/reccmp/isledecomp/compare/db.py
+++ b/reccmp/isledecomp/compare/db.py
@@ -243,7 +243,7 @@ class EntityBatch:
 
         self.reset()
 
-    def __enter__(self):
+    def __enter__(self) -> "EntityBatch":
         return self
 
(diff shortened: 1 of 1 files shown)


jonschz (approved):
Nice! I probably won't have time for the other PRs before the weekend.

PR #162, opened 2025-07-03
Adds support for c-style static vars

jeff-1amstudios:
Handles `static` variables declared inside C functions by parsing `S_LDATA32` entries in the PDB and tracking them as child entries of the parent function.

Example code:
```
// FUNCTION: CARM95 0x00426cc5
int GetCDPathFromPathsTxtFile(char* pPath_name) {
    // GLOBAL: CARM95 0x50a5e4
    static int got_it_already = 0;
    // GLOBAL: CARM95 0x531e00
    static tPath_name cd_pathname;
    ...
```

PDB dump:
```
(004368) S_GPROC32: [0001:00050A28], Cb: 000000B5, Type:             0x1010, GetCDPathFromPathsTxtFile
...
(0043AC)  S_BPREL32: [00000008], Type:   T_32PRCHAR(0470), pPath_name
(0043C4)  S_LDATA32: [0003:0000B3C4], Type:       T_INT4(0074), got_it_already
(0043E4)  S_LDATA32: [0003:0003C488], Type:             0x100B, cd_pathname
```

### DIFF
--- a/reccmp/isledecomp/cvdump/analysis.py
+++ b/reccmp/isledecomp/cvdump/analysis.py
@@ -154,6 +154,24 @@ class CvdumpAnalysis:
                 node_dict[key].node_type = EntityType.FUNCTION
                 node_dict[key].symbol_entry = sym
 
+                # Iterate through static variables defined in this function
+                # generate decorated name "<variable name>___<func name>" for
+                # match_msvc to find them later
+                for v in sym.static_variables:
+                    key = NodeKey(v.section, v.offset)
+                    if key not in node_dict:
+                        node_dict[key] = CvdumpNode(*key)
+                    node_dict[key].node_type = EntityType.DATA
+                    node_dict[key].decorated_name = f"{v.name}___{sym.name}"
+                    node_dict[key].friendly_name = v.name
+                    try:
+                        v_info = parser.types.get(v.type)
+                        node_dict[key].confirmed_size = v_info.size
+                        node_dict[key].data_type = v_info
+                    except (CvdumpKeyError, CvdumpIntegrityError):
+                        # No big deal if we don't have complete type information.
+                        pass
+
         self.nodes: list[CvdumpNode] = [
             v for _, v in dict(sorted(node_dict.items())).items()
         ]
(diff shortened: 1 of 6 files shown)


reccmp/isledecomp/cvdump/types.py
```diff
@@ -704,19 +704,23 @@ def parse_enum_attribute(self, attribute: str) -> dict[str, Any]:
         logger.error("Unknown attribute in enum: %s", attribute)
         return {}
 
-    def read_union(self, leaf: str) -> dict[str, Any]:
-        """This is a rather barebones handler, only parsing the size"""
-        match = self.LF_UNION_LINE.search(leaf)
-        assert match is not None
+def read_union(self, leaf: str) -> dict[str, Any]:
```
madebr: Did you accidentally dedent this member function?

reccmp/isledecomp/cvdump/analysis.py
```diff
@@ -154,6 +154,24 @@ def __init__(self, parser: CvdumpParser):
...
 
+                # Iterate through static variables defined in this function
+                # generate decorated name "<variable name>___<func name>" for
+                # match_msvc to find them later
+                for v in sym.static_variables:
+                    key = NodeKey(v.section, v.offset)
+                    if key not in node_dict:
+                        node_dict[key] = CvdumpNode(*key)
+                    node_dict[key].node_type = EntityType.DATA
+                    node_dict[key].decorated_name = f"{v.name}___{sym.name}"
```
disinvite, line 166: My only minor issue is that this format for the symbol is there specifically to work with the unusual query in `match_static_symbols`. That area definitely needs improvement. But... I don't think we need to hold this up just for that. Maybe if you could add a `# TODO` in here just so it's on our radar?

reccmp/isledecomp/cvdump/types.py
```diff
@@ -705,18 +705,21 @@ def parse_enum_attribute(self, attribute: str) -> dict[str, Any]:
         return {}
 
     def read_union(self, leaf: str) -> dict[str, Any]:
```
disinvite: What's the connection between the change here and the static variables? Should this be its own PR?

jeff-1amstudios: yes, you are right. I fixed both things at the same time. I'll put this out

PR #163, opened 2025-07-03
Fixes field_list_type in union types

jeff-1amstudios:
Correctly handle `field_list_type` in the same way as `read_class_or_struct`


jonschz:
Thanks! Could you add a test (ideally one that would fail on the old code but succeed on the new one)? Extending/modifying an existing one would also be fine if there is an appropriate one, I haven't checked yet.

jeff-1amstudios:
> Thanks! Could you add a test (ideally one that would fail on the old code but pass on the new one)? Extending/modifying an existing one would also be fine if there is an appropriate one, I haven't checked yet.

done!

Issue #77, opened 2025-01-20
Show const return type for overloaded function


disinvite:
Follow up to #76.

PR #164, opened 2025-07-03
New query to detect overloaded functions

disinvite:
More thoroughly addresses the specific regression indicated in #152 and #153.

The type fix in #161 was needed for `mypy` to alert to the code calling `set_orig` or `set_recomp` where the address parameter might be null. This isn't caught if the assumed type is `Any`, hence the reason for moving this query into its own module.

I'm not sure why the old query only considered functions that had a recomp address, but now we can update using whatever address we have.

If the intended effect is to halt if our typing assert  fails (as shown here) then we should repurpose #153 to do the same but at the sqlite level.

### DIFF
--- a/reccmp/isledecomp/compare/core.py
+++ b/reccmp/isledecomp/compare/core.py
@@ -39,6 +39,7 @@ from .match_msvc import (
 from .db import EntityDb, ReccmpEntity, ReccmpMatch
 from .diff import DiffReport, combined_diff
 from .lines import LinesDb
+from .queries import get_overloaded_functions
 
 
 # pylint: disable=too-many-lines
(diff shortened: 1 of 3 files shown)


tests/test_compare_db_queries.py
```diff
@@ -0,0 +1,62 @@
...
+
+
+def test_overloaded_functions_ignore_non_functions(db: EntityDb):
+    with db.batch() as batch:
+        batch.set_orig(100, name="Hello")
+        batch.set_orig(200, name="Hello")
+        batch.set_recomp(300, name="Hello")
+
+    # Name reused, but no entities are functions.
+    assert len([*get_overloaded_functions(db)]) == 0
```
jonschz: matter of taste, I personally prefer `list(iterator)` over `[*iterator]`, but choose whatever you prefer

reccmp/isledecomp/compare/core.py
```diff
@@ -710,36 +711,24 @@ def _unique_names_for_overloaded_functions(self):
...
-
-            repeat_names.setdefault(name, []).append((recomp_addr, symbol))
-
         with self._db.batch() as batch:
-            for name, items in repeat_names.items():
-                for i, (recomp_addr, symbol) in enumerate(items, start=1):
-                    # Just number it to start, in case we don't have a symbol.
-                    new_name = f"{name}({i})"
+            for func in get_overloaded_functions(self._db):
+                # TODO: Thunk's link to the original function is lost once the record is created.
```
jonschz: Not your change, but what exactly does this comment mean, or what is the consequence of this issue, respectively? "Thunks of overloaded functions cannot be distinguished"?

reccmp/isledecomp/compare/queries.py
```diff
@@ -0,0 +1,38 @@
...
+from .db import EntityDb
+
+
+class OverloadedFunctionEntity(NamedTuple):
+    orig_addr: int | None
+    recomp_addr: int | None
+    name: str
+    symbol: str | None
+    # The number for this repeated name, starting at 1, ordered by orig_addr (nulls last), then recomp_addr.
+    sequence: int
```
jonschz: Maybe `index` instead of `sequence`? If a variable is named `sequence`, I'd expect it to be iterable, like a list

reccmp/isledecomp/compare/core.py
```diff
@@ -710,36 +711,24 @@ def _unique_names_for_overloaded_functions(self):
...
+                new_name = f"{func.name}({func.sequence})"
 
-                    if symbol is not None:
-                        dm_args = get_function_arg_string(symbol)
-                        if dm_args is not None:
-                            new_name = f"{name}{dm_args}"
+                if func.symbol is not None:
+                    dm_args = get_function_arg_string(func.symbol)
+                    if dm_args is not None:
+                        new_name = f"{func.name}{dm_args}"
```
jonschz, line 726: Do we have/need any precautions in case the function arg string collides for two variants? Personally, I think the odds are very small, just something to think about.

jonschz (approved):
Nice refactor, and I like the tests!

reccmp/isledecomp/compare/queries.py
disinvite: `index` conflicts with the `tuple.index` function. How about just `number`? Or `order`?

reccmp/isledecomp/compare/core.py
disinvite: It will collide if, for example, you have a const and non-const `operator[]` function. #77

tests/test_compare_db_queries.py
disinvite: `list()` seems to appear more often in the tests, so let's go with that.

reccmp/isledecomp/compare/core.py
disinvite: The sequence of events is:

1. Create entity `TestFunc` at `0x1234`.
2. Create thunk entity for `TestFunc` at `0x1000`. The thunk gets the name `Thunk of 'TestFunc'`.
3. `TestFunc` is overloaded so we give it the more specific name `TestFunc(int)`.
4. The thunk entity at `0x1000` should get the name `Thunk of 'TestFunc(int)'` but it has no link back to the function at `0x1234` so we can't update it.

I haven't worked out the best way to do this.

reccmp/isledecomp/compare/queries.py
jonschz: `number` is fine with me, other options could be `nth`, `repeated_index`, `repeated_name_index`, or something along these lines

reccmp/isledecomp/compare/core.py
jonschz: Is there a ticket for this? Maybe it'd be best to open one and link that in the code, that way it's documented for future developers

reccmp/isledecomp/compare/queries.py
disinvite: Thanks! I like `nth`.

reccmp/isledecomp/compare/core.py
disinvite: #176

PR #165, opened 2025-07-04
Fixes parsing global variables with underscores

jeff-1amstudios:
The original `global_regex` made an assumption that all global variables follow a pattern like `g_xxx`. 

This assumption caused an incorrect result with a global variable like `char* gBoring_material_names[2];`, which was parsed as `g_material_names`.

This PR uses the original "less strict" regex in all cases, and adds a test which fails on current `master` but passes with this change.


jonschz (approved):
Looks good! I didn't get any regressions on LEGO1 or BETA10.

PR #167, opened 2025-07-06
Adds function filtering to the compare report

jeff-1amstudios:
Project file now supports an `ignore_functions` list for functions that should not appear in the final report, and paves the way for future report-specific fields

Example:
```
targets:
  CARM95:
    filename: carm95.exe
    source_root: src
    hash:
      sha256: c6040203856b71e6a22d2a29053a1eadd1a2ab41bce97b6031d745079bc07bdf
    report:
      ignore_functions: [
        "BrActorToActorMatrix34",
        "BrActorRelink",
        ...
     ]
```

In my case, some functions are coming from a now open source library (https://github.com/dethrace-labs/BRender-v1.3.2), and are not expected to be a binary match since the original was not compiled with MSVC 4.2.

Being able to ignore those functions makes the percentage complete numbers meaningful ✨

### DIFF
--- a/reccmp/project/config.py
+++ b/reccmp/project/config.py
@@ -40,6 +40,17 @@ class YmlGhidraConfig(BaseModel):
         return cls(ignore_types=[], ignore_functions=[])
 
 
+class YmlReportConfig(BaseModel):
+    ignore_functions: list[str] = Field(
+        default_factory=list,
+        validation_alias=AliasChoices("ignore-functions", "ignore_functions"),
+    )
+
+    @classmethod
+    def default(cls) -> "YmlReportConfig":
+        return cls(ignore_functions=[])
+
+
 @dataclass
 class Hash:
     sha256: str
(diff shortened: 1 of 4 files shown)


jeff-1amstudios:
@disinvite or @jonschz could you add me to the list of people where the workflows run automatically? Thanks!

disinvite:
If we fast-track #32 and support setting the `skip` attribute for a given address in the CSV file, that should accomplish the same thing as this PR. What do you think?

disinvite:
I thought about this more last night and I'm now on board with this change. I think we can mitigate the complexity of managing/combining all the YML options by breaking those out into their own functions in a later PR. There will probably be a need or want for more options under the `report` key so the structure is fine too.

jeff-1amstudios:
It could do - I named the field `ignore_functions` to make it clear this affects _functions_ only - I couldn't think of a use case for ignoring non-functions in the report, since only functions contribute to the overall percentage complete number.

How do we square this with #169 though? I'm fine to go with #169 over this one if you prefer, but it doesn't seem like we should do both?

PR #169, opened 2025-07-07
Support loading (some) data from CSV

disinvite:
This adds support for loading annotations for the original binary from a CSV text file. This initial PR supports only limited data as I'll explain in a moment.

## YML

The list of files goes in your `reccmp-project.yml` file like so:

```yml
  BETA10:
    filename: BETA10.DLL
    source-root: LEGO1
    hash:
      sha256: d91435a40fa31f405fba33b03bd3bd40dcd4ca36ccf8ef6162c6c5ca0d7190e7
    csv:
      - LEGO1/annotations/mxatom.csv
```

You can use the keys `csv`, `csv-files` or `csv_files`. We expect a list of paths relative to the project root even if there is only one file to load. The paths should point to each CSV file directly. We don't support globs or loading all files from a directory yet.

## CSV format

When reading the file, we ignore blank lines and lines that begin with comment indicators `#` or `//`.

The options for delimiter in the CSV file are: pipe, comma, or tab. The delimiter is chosen by reading the first (non-blank and non-commented) line.

For example:

```
# STL functions for MxAtom set (BETA10)

addr|symbol
0x101237a0|??Dconst_iterator@?$_Tree@PAVMxAtom@@PAV1@U_Kfn@?$set@PAVMxAtom@@UMxAtomCompare@@V?$allocator@PAVMxAtom@@@@@@UMxAtomCompare@@V?$allocator@PAVMxAtom@@@@@@QBEABQAVMxAtom@@XZ
0x101237d0|?_Value@?$_Tree@PAVMxAtom@@PAV1@U_Kfn@?$set@PAVMxAtom@@UMxAtomCompare@@V?$allocator@PAVMxAtom@@@@@@UMxAtomCompare@@V?$allocator@PAVMxAtom@@@@@@KAAAPAVMxAtom@@PAU_Node@1@@Z
0x101237f0|?end@?$set@PAVMxAtom@@UMxAtomCompare@@V?$allocator@PAVMxAtom@@@@@@QBE?AVconst_iterator@?$_Tree@PAVMxAtom@@PAV1@U_Kfn@?$set@PAVMxAtom@@UMxAtomCompare@@V?$allocator@PAVMxAtom@@@@@@UMxAtomCompare@@V?$allocator@PAVMxAtom@@@@@@XZ
...
```

The CSV file must contain the address as one of the columns, and it must appear only once. You can use either `address` or `addr` as the name for the column. The address is required to be a hex number. The `0x` prefix is optional. I expect that the address will appear as the first column in most cases, but this is not required. (If for example you wanted to sort by alphabetical order.)

## Limitations

To get this out quickly, we only support two column values: `symbol` and `skip`. All other columns are ignored and we don't set anything in the database.

This means that there are only two real use cases at the start:

- Marking templated functions best (or only) identified by symbol, as in the above STL example.
- Specifying that a function should not be be compared or considered for the final accuracy calculation, beyond what is possible in the current code annotation language. (See #167)

### DIFF
--- a/reccmp/isledecomp/compare/core.py
+++ b/reccmp/isledecomp/compare/core.py
@@ -36,6 +36,7 @@ from .match_msvc import (
     match_variables,
     match_strings,
 )
+from .csv import ReccmpCsvParserError, csv_parse
 from .db import EntityDb, ReccmpEntity, ReccmpMatch
 from .diff import DiffReport, combined_diff
 from .lines import LinesDb
@@ -49,6 +50,7 @@ logger = logging.getLogger(__name__)
 
 class Compare:
     # pylint: disable=too-many-instance-attributes
+    # pylint: disable=too-many-arguments
     def __init__(
         self,
         orig_bin: PEImage,
@@ -56,6 +58,7 @@ class Compare:
         pdb_file: Path | str,
         code_dir: Path | str,
         target_id: str | None = None,
+        csv_files: list[Path] | None = None,
     ):
         self.orig_bin = orig_bin
         self.recomp_bin = recomp_bin
@@ -82,6 +85,10 @@ class Compare:
 
         self._load_cvdump()
         self._load_markers(report)
+
+        if isinstance(csv_files, list):
+            for csv_file in csv_files:
+                self._load_csv(csv_file)
 
         # Match using PDB and annotation data
         match_symbols(self._db, report, truncate=True)
@@ -124,6 +131,7 @@ class Compare:
             target.recompiled_pdb,
             target.source_root,
             target_id=target.target_id,
+            csv_files=target.csv_files,
         )
 
     @property
@@ -368,6 +376,19 @@ class Compare:
                     line=line.line_number,
                     type=EntityType.LINE,
                 )
+
+    def _load_csv(self, path: Path):
+        try:
+            with open(path, "r", encoding="utf-8") as f:
+                # TODO: Aborts on any error instead of skipping a row where this is possible
+                rows = list(csv_parse(f))
+        except (FileNotFoundError, ReccmpCsvParserError):
+            logger.error("Failed to parse csv file %s", str(path))
+            return
+
+        with self._db.batch() as batch:
+            for addr, values in rows:
+                batch.set_orig(addr, **values)
 
     def _match_array_elements(self):
         """
--- a/reccmp/isledecomp/compare/csv.py
+++ b/reccmp/isledecomp/compare/csv.py
@@ -1,0 +1,116 @@
+"""Parsing files in csv format.
+The python csv.DictReader class does the heavy lifting.
+We begin with a pre-processing step that removes blank lines or "comment" lines that begin with # or //.
+The delimiter is decided based on the first line (following pre-processing). Options are: pipe, comma, tab
+"""
+
+import csv
+from csv import Error as PythonCsvError
+from typing import Iterable, Iterator
+
+
+class ReccmpCsvParserError(Exception):
+    """Catch-all for csv parsing errors."""
+
+
+class CsvNoAddressError(ReccmpCsvParserError):
+    """This row does not have an address attribute."""
+
+
+class CsvMultipleAddressError(ReccmpCsvParserError):
+    """This row has more than one address attribute
+    (and it's not clear which one we should use)."""
+
+
+class CsvInvalidAddressError(ReccmpCsvParserError):
+    """The address value is not a valid hex number."""
+
+
+class CsvNoDelimiterError(ReccmpCsvParserError):
+    """No obvious delimiter on the first line."""
+
+
+CsvValueOptions = int | str | bool
+CsvValuesType = dict[str, CsvValueOptions]
+
+
+def _boolify(text: str) -> bool:
+    """str to bool conversion. If the string is not in the exclusion list, resolve to True."""
+    return text.strip().lower() not in ("false", "off", "no", "0", "")
+
+
+def _csv_preprocess(lines: Iterable[str]) -> Iterator[str]:
+    """Pre-processing of CSV file."""
+    for line in lines:
+        strip = line.strip()
+        # Skip comment lines
+        if strip.startswith("#") or strip.startswith("//"):
+            continue
+
+        # Skip blank lines.
+        if strip == "":
+            continue
+
+        yield line
+
+
+def _convert_attrs(
+    values: Iterable[tuple[str, str]],
+) -> Iterator[tuple[str, CsvValueOptions]]:
+    """Both a filter and a conversion step for the row values.
+    For the incoming iterable of key/value pairs, only output the ones we want set
+    in the reccmp database. Some keys have their value converted to a different type."""
+    for key, value in values:
+        if key == "symbol":
+            yield (key, value)
+
+        if key == "skip":
+            yield (key, _boolify(value))
+
+
+def _csv_convert(row: dict[str, str]) -> tuple[int, CsvValuesType]:
+    """Pull out the address from the CSV row and convert the remaining key/value pairs."""
+    for key in ("address", "addr"):
+        if key in row:
+            addr_value = row[key]
+            break
+
+    # Addr is always a hex number
+    try:
+        addr = int(addr_value, 16)
+    except ValueError as ex:
+        raise CsvInvalidAddressError from ex
+
+    attrs = list(_convert_attrs(row.items()))
+    return (addr, dict(attrs))
+
+
+def csv_parse(lines: str | Iterable[str]) -> Iterator[tuple[int, CsvValuesType]]:
+    """Read each line from the csv file and output each address and its key/value pairs."""
+    if isinstance(lines, str):
+        lines = lines.split("\n")
+
+    preprocessed = list(_csv_preprocess(lines))
+
+    try:
+        # Use the first line (only) to find the delimiter
+        dialect = csv.Sniffer().sniff(preprocessed[0], delimiters="|,\t")
+    except PythonCsvError as ex:
+        raise CsvNoDelimiterError from ex
+
+    reader = csv.DictReader(preprocessed, dialect=dialect)
+
+    # Could this happen?
+    if reader.fieldnames is None:
+        raise CsvNoDelimiterError
+
+    # We support multiple options for address key, but exactly one must appear.
+    addrs = [key for key in ("address", "addr") if key in reader.fieldnames]
+    if not addrs:
+        raise CsvNoAddressError
+
+    if len(addrs) > 1:
+        raise CsvMultipleAddressError
+
+    for row in reader:
+        yield _csv_convert(row)
--- a/reccmp/project/config.py
+++ b/reccmp/project/config.py
@@ -53,6 +53,10 @@ class ProjectFileTarget(BaseModel):
         validation_alias=AliasChoices("source-root", "source_root")
     )
     hash: Hash
+    csv_files: list[Path] = Field(
+        validation_alias=AliasChoices("csv", "csv_files", "csv-files"),
+        default_factory=list,
+    )
     ghidra: YmlGhidraConfig = Field(default_factory=YmlGhidraConfig.default)
 
 
--- a/reccmp/project/detect.py
+++ b/reccmp/project/detect.py
@@ -117,6 +117,9 @@ class RecCmpPartialTarget:
     recompiled_path: Path | None = None
     recompiled_pdb: Path | None = None
 
+    # Data to set directly in the database (addresses refer to orig binary)
+    csv_files: list[Path] | None = None
+
 
 @dataclass
 class RecCmpTarget:
@@ -146,6 +149,9 @@ class RecCmpTarget:
     original_path: Path
     recompiled_path: Path
     recompiled_pdb: Path
+
+    # Data to set directly in the database (addresses refer to orig binary)
+    csv_files: list[Path] = field(default_factory=list)
 
 
 class RecCmpProject:
@@ -204,6 +210,8 @@ class RecCmpProject:
         else:
             ghidra = GhidraConfig()
 
+        csv_files = target.csv_files or []
+
         return RecCmpTarget(
             target_id=target.target_id,
             filename=target.filename,
@@ -213,6 +221,7 @@ class RecCmpProject:
             recompiled_pdb=target.recompiled_pdb,
             source_root=target.source_root,
             ghidra_config=ghidra,
+            csv_files=csv_files,
         )
 
     def find_build_config(self, search_path: Path) -> BuildFile | None:
@@ -297,7 +306,10 @@ class RecCmpProject:
             else:
                 ghidra = None
 
+            # Assumes these are relative paths. If they are not, the second path
+            # will replace the first instead of adding onto it.
             source_root = project_directory / target.source_root
+            csv_files = [project_directory / csv_path for csv_path in target.csv_files]
 
             project.targets[target_id] = RecCmpPartialTarget(
                 target_id=target_id,
@@ -305,6 +317,7 @@ class RecCmpProject:
                 sha256=target.hash.sha256,
                 source_root=source_root,
                 ghidra_config=ghidra,
+                csv_files=csv_files,
             )
 
         # Apply reccmp-user.yml
--- a/tests/test_csv.py
+++ b/tests/test_csv.py
@@ -1,0 +1,258 @@
+from textwrap import dedent
+import pytest
+from reccmp.isledecomp.compare.csv import (
+    csv_parse,
+    CsvNoAddressError,
+    CsvMultipleAddressError,
+    CsvInvalidAddressError,
+    CsvNoDelimiterError,
+)
+
+
+def test_no_delimiter():
+    """Must have a delimiter. Having just one column isn't very useful."""
+    with pytest.raises(CsvNoDelimiterError):
+        list(csv_parse("addr"))
+
+
+def test_valid_addr_column():
+    """The only requirement is that there is a single address column and a delimiter."""
+    list(csv_parse("addr|symbol"))
+    list(csv_parse("address|symbol"))
+
+
+def test_no_address_column():
+    """Cannot parse any rows if there is no address column."""
+    with pytest.raises(CsvNoAddressError):
+        list(csv_parse("symbol|test"))
+
+
+def test_multiple_address_column():
+    """Cannot parse any rows if there is not a single address column."""
+    with pytest.raises(CsvMultipleAddressError):
+        list(csv_parse("address|symbol|addr"))
+
+
+def test_value_includes_delimiter():
+    """If the value contains the delimiter, we can still parse it correctly
+    if the value is quoted. This is a feature of the python csv module."""
+    values = [
+        *csv_parse(
+            dedent(
+                """\
+        address,symbol
+        1000,"hello,world"
+    """
+            )
+        )
+    ]
+
+    assert values == [(0x1000, {"symbol": "hello,world"})]
+
+
+def test_ignore_columns():
+    """We only parse certain columns that correspond to attribute names in the database."""
+    values = [
+        *csv_parse(
+            dedent(
+                """\
+        address|symbol|test
+        1000|hello|123
+    """
+            )
+        )
+    ]
+
+    assert values == [(0x1000, {"symbol": "hello"})]
+
+
+def test_address_not_hex():
+    """Raise an exception if we cannot parse the address on one of the rows."""
+    with pytest.raises(CsvInvalidAddressError):
+        list(
+            csv_parse(
+                dedent(
+                    """\
+            addr|symbol
+            wrong|test
+        """
+                )
+            )
+        )
+
+
+def test_too_many_columns():
+    """Should ignore extra values in a row."""
+    values = [
+        *csv_parse(
+            dedent(
+                """\
+        addr|symbol
+        1000|hello|world
+    """
+            )
+        )
+    ]
+
+    assert values == [(0x1000, {"symbol": "hello"})]
+
+
+def test_should_output_bool():
+    """Return bool for certain column values, with some flexibility around possible text values."""
+
+    # Using "skip" as an example of a columm where we convert from str to bool:
+    values = [
+        *csv_parse(
+            dedent(
+                """\
+                addr|skip
+                1000|1
+                2000|yes
+                3000|no
+                4000|FALSE
+                5000|0
+                6000|
+            """
+            )
+        )
+    ]
+
+    # To make the following code cleaner
+    skip_map = {addr: row["skip"] for addr, row in values}
+
+    # Any text is considered true...
+    assert skip_map[0x1000] is True
+    assert skip_map[0x2000] is True
+
+    # except the values for these columns
+    assert skip_map[0x3000] is False
+    assert skip_map[0x4000] is False
+    assert skip_map[0x5000] is False
+
+    # Empty string considered false
+    assert skip_map[0x6000] is False
+
+
+def test_ignore_blank_lines():
+    """Parsing should ignore blank lines.
+    n.b. make sure the triple quote string and dedent() remove leading spaces."""
+    values = [
+        *csv_parse(
+            dedent(
+                """\
+
+                addr|symbol
+
+                1000|test
+
+                2000|test
+
+
+                3000|test
+            """
+            )
+        )
+    ]
+
+    assert values == [
+        (0x1000, {"symbol": "test"}),
+        (0x2000, {"symbol": "test"}),
+        (0x3000, {"symbol": "test"}),
+    ]
+
+
+def test_ignore_comments():
+    """We ignore any line starting with // or #."""
+    values = [
+        *csv_parse(
+            dedent(
+                """\
+                # Test CSV file
+                addr|symbol
+                # 1000|test
+                2000|test
+                // 3000|test
+            """
+            )
+        )
+    ]
+
+    assert values == [(0x2000, {"symbol": "test"})]
+
+
+def test_tab_delimiter():
+    """We support tab as an option for delimiter. (It is not covered in the other tests.)"""
+    values = [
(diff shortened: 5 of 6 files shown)


reccmp/isledecomp/compare/csv.py
```diff
@@ -0,0 +1,116 @@
...
+    for key in ("address", "addr"):
+        if key in row:
+            addr_value = row[key]
+            break
+
+    # Addr is always a hex number
+    try:
+        addr = int(addr_value, 16)
+    except ValueError as ex:
+        raise CsvInvalidAddressError from ex
```
madebr: I'd make this intelligent and require a prefix.
```suggestion
    try:
        if addr_value.startswith("0x"):
            addr = int(addr_value[2:], 16)
        else:
            addr = int(addr_value, 10)
    except ValueError as ex:
        raise CsvInvalidAddressError from ex
```
(Do we have a number converter helper function somewhere?)

reccmp/isledecomp/compare/csv.py
```diff
@@ -0,0 +1,116 @@
...
+    if reader.fieldnames is None:
+        raise CsvNoDelimiterError
+
+    # We support multiple options for address key, but exactly one must appear.
+    addrs = [key for key in ("address", "addr") if key in reader.fieldnames]
+    if not addrs:
+        raise CsvNoAddressError
+
+    if len(addrs) > 1:
+        raise CsvMultipleAddressError
```
madebr: Why do we support multiple address keys? 3 extra bytes in the first row will not eat your hard drive.
We're making the code harder to read for little gain imho.

Support for multiple possible csv keys in the yml is nicely abstracted away by pydantic though.
(Does pydantic fail when you have e.g. both `csv` and `csv-files` defined for a target?)

disinvite: There isn't a strong reason to support both `address` and `addr` except to allow for some user preference. Same principle as having multiple options for YML keys.

pydantic accepts multiple keys for `csv_files` but will only use the values for one, based on the order in `AliasChoices`. I think we'd prefer to have it fail if more than one alias is used. I'll see if there's a way to do that.

reccmp/isledecomp/compare/csv.py
disinvite: It should convert the same way for every row, because you could have sloppy data like:
```
addr|test
0x1000|hello
1024|test
```
and the second row would be interpreted as a decimal silently.

Could the choice of column name drive the conversion? `decimal_address`? `segment_address` for 16-bit values? I guess the main concern is to not shoot ourselves in the foot later on by setting expectations for `address` that are too rigid.

madebr: I prefer it to be strict so users can catch errors. (That's also why I'd like to avoid making the source annotations too configurable).

About sloppy data, I think this project can assume the authors have control over all data sources.
If the data source is sloppy, they should fix it.

Thinking about it, perhaps requiring all addresses to be hexadecimal makes sense (let's not add extra columns).
If we ever support 16-bit x86, we can then require the address to be formatted as e.g. `0014:ef24`. (`:` is not a csv separator)

reccmp/isledecomp/compare/csv.py
disinvite: You can do it by adding:
```python
model_config = ConfigDict(extra="forbid")
```
but this also throws a validation error if any extra (unused) fields are in the target.

We should probably break this off into its own issue, unless there's another way to do this just for one field and its aliases.

madebr: Looks like pydantic also supports [validating csv files](https://docs.pydantic.dev/latest/examples/files/#csv-files).
But I don't feel strong about any of this. I just want reccmp to fail hard on invalid input.

Perhaps the file list should be a subitem of csv in the yml?
```yml
- LEGO:
    csv:
      ignore-extra: true
      files: ['data/addresses.csv'] 
```
or make things configurable per-file (idk)

jeff-1amstudios:
Is the intention for the csv file to be a companion or a replacement to the code annotations?

I assume it is the former, since the code annotations are more legible given they are attached in a very visible way to the symbol name, where-as here its just an address?

If so, what is your intention for the meaning of the `skip` field? Just skip from the report, or to negate any code annotation for that address?

reccmp/isledecomp/compare/csv.py
jeff-1amstudios: Agree with @madebr - as I user I might prefer `addr` over `address`, but my stronger desire is to just easily read the examples/code to know what I need to put in those files. (And have it fail quickly and obviously if I screw it up!)

I would prefer the project to dictate clearly the one naming scheme, rather than giving me a lot of options

reccmp/isledecomp/compare/core.py
```diff
@@ -49,13 +50,15 @@
...
     # pylint: disable=too-many-instance-attributes
+    # pylint: disable=too-many-arguments
     def __init__(
         self,
         orig_bin: PEImage,
         recomp_bin: PEImage,
         pdb_file: Path | str,
         code_dir: Path | str,
         target_id: str | None = None,
+        csv_files: list[Path] | None = None,
```
jeff-1amstudios: How about something more meaningful - `overrides_files`, `external_symbol_definitions`, `extra_symbol_files`?

reccmp/project/config.py
```diff
@@ -53,6 +53,10 @@ class ProjectFileTarget(BaseModel):
         validation_alias=AliasChoices("source-root", "source_root")
     )
     hash: Hash
+    csv_files: list[Path] = Field(
+        validation_alias=AliasChoices("csv", "csv_files", "csv-files"),
```
jeff-1amstudios: all these alias's become part of your config surface that is hard to change later without breaking users. Unless theres a strong need, I'd skip aliases for this reason

reccmp/isledecomp/compare/csv.py
```diff
@@ -0,0 +1,112 @@
...
+        if key == "skip":
+            yield (key, _boolify(value))
+
+
+def _csv_convert(addr_key: str, row: dict[str, str]) -> tuple[int, CsvValuesType]:
+    """Pull out the address from the CSV row and convert the remaining key/value pairs."""
+    try:
+        # Addr is always a hex number
+        addr_value = row[addr_key]
+        addr = int(addr_value, 16)
```
jeff-1amstudios, line 156: I think we should enforce a `0x` prefix to make this very obvious. 
Reasoning - I tried out `ghirda_config.ignore_functions` initially and had to look at the code to see that it expected decimal addresses :)

reccmp/isledecomp/compare/csv.py
```diff
@@ -0,0 +1,112 @@
+"""Parsing files in csv format.
+The python csv.DictReader class does the heavy lifting.
+We begin with a pre-processing step that removes blank lines or "comment" lines that begin with # or //.
+The delimiter is decided based on the first line (following pre-processing). Options are: pipe, comma, tab
+"""
+
+import csv
```
jeff-1amstudios, line 7: I understood from your earlier comment about not wanting to add more fields to the project file, but since all other files are in YAML format, still not sure why this one should be different.

YAML isn't perfect ofc, but it does end up more descriptive than csv, supports comments natively and can be schema-validated.

Anyway, not a huge deal, just trying to understand the choice here

reccmp/project/config.py
disinvite: I would agree since these do not work the way I thought they did (many aliases are possible, but only one can appear)

reccmp/isledecomp/compare/csv.py
disinvite: I picked CSV just because it's simple. We can support YML files too.

reccmp/isledecomp/compare/csv.py
jonschz: +1 for not using decimal addresses at all

reccmp/isledecomp/compare/csv.py
jonschz: > that it expected decimal addresses :)

That's not actually correct - it expects a number, and yaml supports both hex and decimal. So
```yml
      ignore-functions:
        - 0x100f8ad0
        - 0x100fa200
```
is what I'd use

jonschz:
> This places more importance on the file extension to decide how to parse each one (if that matters).

I agree - the key should communicate the intention, not technical information. But I think we should document our intents regarding each type of config file (maybe in docstrings? Proper architecture documentation is likely overkill) so there's less debate in case we add more options in the future.

LRFLEW:
This seems like a really useful feature (and would let me replace my 650-ish line file with an easier-to-update one). I did have one thought about this looking over it (other than allowing for different symbol types, which appears to already be a thing).

A use-case I could see myself wanting to use is using a single CSV file for multiple targets. This would be equivalent to combining multiple `LIBRARY` annotations with different targets, like [what's used here in isle decomp](https://github.com/isledecomp/isle/blob/master/LEGO1/library_msvc.h#L134-L136). In these cases, it might be nice if, instead of having an `addr` / `address` column, we use columns for each target instead. It should also support leaving such a column empty to indicate the row doesn't apply to the target. As an example, the linked example could be converted to CSV as something like this:

```csv
LEGO1|BETA10|type|symbol
0x1008c2e0|0x100f9800|library|??_M@YGXPAXIHP6EX0@Z@Z
0x1008c410||library|strlwr
...
```

reccmp/isledecomp/compare/core.py
```diff
@@ -91,6 +94,11 @@ def __init__(
         self._load_cvdump()
         self._load_markers(report)
 
+        if isinstance(data_sources, list):
+            for ds_file in data_sources:
+                if ds_file.suffix.lower() == ".csv":
```
jonschz: maybe log a warning in case the file is ignored due to an unexpected extension?

reccmp/isledecomp/compare/core.py
```diff
@@ -413,6 +422,21 @@ def orig_bin_checker(addr: int) -> bool:
                     type=EntityType.LINE,
                 )
 
+    def _load_csv(self, path: Path):
+        try:
+            with open(path, "r", encoding="utf-8") as f:
+                # TODO: Aborts on any error instead of skipping a row where this is possible
```
jonschz: Address TODO or write a follow-up

reccmp/isledecomp/compare/csv.py
```diff
@@ -0,0 +1,160 @@
...
+    type: NotRequired[EntityType]
+    name: NotRequired[str]
+    size: NotRequired[int]
+    skip: NotRequired[bool]
+    symbol: NotRequired[str]
+
+
+def _boolify(text: str) -> bool:
+    """str to bool conversion. If the string is not in the exclusion list, resolve to True."""
+    return text.strip().lower() not in ("false", "off", "no", "0", "")
```
jonschz: personal opinion: I'd map `"false"` and `""` to `False`, `"true"` to `True`, and raise an error (i.e. skip the line) otherwise. One reason: `"false "` evaluates to `True`, and there may be other unexpected behaviour

tests/test_project_yml.py
jonschz, line 1: Add a test case for a non-csv file

reccmp/isledecomp/compare/core.py
disinvite: Done. Let me know what you think of the `while` loop I used to replace this.

disinvite:
> A use-case I could see myself wanting to use is using a single CSV file for multiple targets. This would be equivalent to combining multiple `LIBRARY` annotations with different targets, like [what's used here in isle decomp](https://github.com/isledecomp/isle/blob/master/LEGO1/library_msvc.h#L134-L136).

Is it okay if we save this for a follow-up? It's a good idea but I want to assess the impact on its own. It would limit the set of valid module names, but I doubt many teams will be be dead set on using `TYPE` or `SYMBOL`.

If you're only combining the files for the library functions, though, it may not prove to be as convenient as you think. At least in the case of `LEGO1` and `BETA10`, the library functions were not in the same order. Having the addresses together in `library_msvc.h` didn't reveal much or save any time. You may find it easier to just copy the list of auto-detected functions from Ghidra to establish the csv file.

reccmp/isledecomp/compare/core.py
jonschz: Looks good! I haven't taken a thorough look at everything, but from a glance, all I'm missing is some documentation at the moment. For my review I'd try to set something up based on the new README.

reccmp/isledecomp/compare/core.py
```diff
@@ -413,6 +427,38 @@ def orig_bin_checker(addr: int) -> bool:
                     type=EntityType.LINE,
                 )
 
+    def _load_csv(self, path: Path):
```
jonschz: I'd prefer not to have more code in `core.py`. Maybe extract this function somewhere else and pass the `DB` as an argument?

reccmp/project/detect.py
```diff
@@ -149,6 +149,9 @@ class RecCmpPartialTarget:
     recompiled_path: Path | None = None
     recompiled_pdb: Path | None = None
 
+    # Data to set directly in the database (addresses refer to orig binary)
+    data_sources: list[Path] | None = None
```
jonschz, line 153: Applies to the previous lines as well: If you put the description into a docstring below, it will be picked up by most IDEs. E.g.
```py
    data_sources: list[Path] | None = None
    """Data to set directly in the database (addresses refer to orig binary)"""
```

docs/csv.md
```diff
@@ -0,0 +1,115 @@
...
+      sha256: 5cf57c284973fce9d14f5677a2e4435fd989c5e938970764d00c8932ed5128ca
+    data-sources:
+      - test.csv
+```
+
+The paths are relative to the directory that contains `reccmp-project.yml`.
+
+## Format
+
+Here's an example of a CSV file using the pipe character[^1] as the delimiter.
```
jonschz: This footnote syntax does not appear to work in VS Code, though it's fine in GitHub. Are you aware of a more compatible format? If not, I'd leave it like that.

docs/csv.md
```diff
@@ -0,0 +1,115 @@
+# CSV import
+
+`reccmp` has the option to parse [comma-separated value (CSV)](https://en.wikipedia.org/wiki/Comma-separated_values) files and load data to supplement code annotations on the original binary.
```
jonschz, line 3: I like the "how" part of this README! What I'm still missing is a bit of the "why", e.g.:
- everything works fine without CSV, so this feature is optional
- you may have additional data from other sources, and export into CSV is easier than into source code
- annotating the size of objects is not supported in code annotations
- ... (I have probably forgotten about several aspects)

docs/csv.md
```diff
@@ -0,0 +1,115 @@
...
+```
+address|name|size
+1008b400|_atol|164
+1008b4b0|_atoi|14
+1008b4c0|_strtok|216
+1008b5a0|_sprintf|103
+1008b608|__ftol|39
+```
+
+The header on the first line defines the columns used in the file. The only required column is `"address"`. You can use the abbreviated `"addr"` if you prefer, but not both in the same file. `"address"` does not need to appear first, but this is common.
```
jonschz: As discussed in the other threads, I'd remove `addr` in favour of `address` both in code an in the docs

docs/csv.md
```diff
@@ -0,0 +1,115 @@
...
+
+`"type"` expects one of these values:
+
+- `function, template, synthetic, library, stub` for function entities
+- `global`
+- `string`
+- `float`
+- `vtable`
+
+These are the same[^3] categories used in the [code annotations](https://github.com/isledecomp/reccmp/blob/master/docs/annotations.md).
```
jonschz: Use a relative link (so this works on a local machine as well)

docs/csv.md
```diff
@@ -0,0 +1,115 @@
...
+Blank lines are ignored entirely.
+
+### Column reference
+
+These fields can appear in the CSV file:
+
+| column | description | possible values |
+| ------ | ----------- | --------------- |
+| `address` | Address of the entity. Required. | Hex number with or without `0x` prefix |
+| `addr` | Alias for `address`. | |
```
jonschz: Remove `addr`

docs/csv.md
```diff
@@ -0,0 +1,115 @@
...
+0x10001070||25
+0x10001090|function|10
+```
+
+If the value contains the delimiter, wrap the entire thing in double quotes. For example:
+
+```csv
+addr,name
+101310a0,"set<MxAtom *,MxAtomCompare,allocator<MxAtom *> >::set<MxAtom *,MxAtomCompare,allocator<MxAtom *> >"
+```
```
jonschz, line 83: Maybe mention that double quotes need to be escaped by using two double quotes (see also the comment on the tests)

docs/csv.md
```diff
@@ -0,0 +1,115 @@
...
+1008b4c0|_strtok|216
+1008b5a0|_sprintf|103
+1008b608|__ftol|39
+```
+
+The header on the first line defines the columns used in the file. The only required column is `"address"`. You can use the abbreviated `"addr"` if you prefer, but not both in the same file. `"address"` does not need to appear first, but this is common.
+
+The address value is always interpreted as a hexadecimal number, even if only digits 0-9 appear. You can use the typical `0x` prefix but it was omitted for this example. The `h` suffix sometimes seen in assembly code is not supported.
+
+The `"name"` field is the name for the entity, as you would expect. `"size"` is the number of bytes used by the entity in virtual memory[^2]. This is always interpreted as a decimal number.
```
jonschz: What isn't immediately obvious to me is the presence of both `name` and `symbol`. Which one takes precedence if both are specified?

docs/csv.md
```diff
@@ -0,0 +1,115 @@
...
+These fields can appear in the CSV file:
+
+| column | description | possible values |
+| ------ | ----------- | --------------- |
+| `address` | Address of the entity. Required. | Hex number with or without `0x` prefix |
+| `addr` | Alias for `address`. | |
+| `name` | Name for the entity (e.g. the fully-qualified function name) | text |
+| `symbol` | Linker name | text |
+| `type` | Entity type | one of: `function, template, synthetic, library, stub, global, string, float, vtable` |
+| `report.ignore` | If `true`, do not include this entity in the final comparison report. | `1` or `0` or `true` or `false` |
```
jonschz: I vaguely remember that we had this debate recently. In my opinion, we should stick to `true`/`false` and disallow anything else (just a remark, feel free to keep it if you strongly prefer your version)

reccmp/isledecomp/compare/csv.py
```diff
@@ -0,0 +1,243 @@
...
+    """Cannot return even a single row from this document."""
+
+
+class CsvNoAddressError(ReccmpCsvFatalParserError):
+    """This file does not have an address attribute."""
+
+
+class CsvMultipleAddressError(ReccmpCsvFatalParserError):
+    """This file has more than one address attribute
+    (and it's not clear which one we should use)."""
```
jonschz: I'd generalise this error to a `CsvDuplicateColumnError` and also raise if another non-mandatory column is duplicated

tests/test_csv.py
```diff
@@ -0,0 +1,590 @@
...
+        except CsvInvalidAddressError as ex:
+            assert ex.illegal_value == "zzzz"
+            assert ex.line_number == 6
+            count += 1
+            continue
+        except CsvInvalidEntityTypeError as ex:
+            assert ex.illegal_value == "libary"
+            assert ex.line_number == 3
+            count += 1
+            continue
```
jonschz: Minor point: This would be more idiomatic without a loop, with manual `next` calls, and `pytest.raises(...) as ex`

tests/test_csv.py
```diff
@@ -0,0 +1,590 @@
...
+        addr,name
+        101310a0,\"set<MxAtom *,MxAtomCompare,allocator<MxAtom *> >::set<MxAtom *,MxAtomCompare,allocator<MxAtom *> >\""""
+    )
+    assert (
+        0x101310A0,
+        {
+            "name": "set<MxAtom *,MxAtomCompare,allocator<MxAtom *> >::set<MxAtom *,MxAtomCompare,allocator<MxAtom *> >"
+        },
+    ) in list(csv_parse(text))
+
```
jonschz, line 505: Add a test case for escaping a double quote, even though this is not explicitly handled by our library

docs/csv.md
```diff
@@ -0,0 +1,115 @@
...
+100db614,string
+100db620,string
+100db628,string
+```
+
+Blank lines are ignored entirely.
+
+### Column reference
+
+These fields can appear in the CSV file:
```
jonschz, line 122: Add a warning that only `address`, `symbol`, `type`, and `skip` currently have an effect (or do any of the others now have one as well? `size` doesn't do anything on functions yet)

reccmp/isledecomp/compare/core.py
jonschz, line 1: After adding a CSV to `LEGO1` with the contents
```csv
address|symbol
0x1008a090|_malloc
```
and commenting out the `_malloc` annotation in `library_msvc.h`, I run into an `AssertionError`:
```
  File "<frozen runpy>", line 198, in _run_module_as_main
  File "<frozen runpy>", line 88, in _run_code
  File "C:\Users\Jonathan\vscode-workspace\isle\.venv\Scripts\reccmp-reccmp.exe\__main__.py", line 7, in <module>
  File "C:\Users\Jonathan\vscode-workspace\reccmp\reccmp\tools\asmcmp.py", line 222, in main
    match = isle_compare.compare_address(args.verbose)
            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "C:\Users\Jonathan\vscode-workspace\reccmp\reccmp\isledecomp\compare\core.py", line 984, in compare_address
    return self._compare_match(match)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "C:\Users\Jonathan\vscode-workspace\reccmp\reccmp\isledecomp\compare\core.py", line 906, in _compare_match
    assert match.entity_type is not None
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
AssertionError
```
To me, it's quite clear that the entity type must be specified in the CSV if it isn't specified anywhere else, but that isn't obvious from either the README or the error message. How would you like to proceed? A follow-up would be fine with me.

docs/csv.md
disinvite: I'm going to just drop the `skip` option altogether so there's no need to parse any bools. It's no longer necessary after #167.

reccmp/isledecomp/compare/core.py
disinvite: That's a separate problem. If we drop the assert and just return `None` when the entity has no type (because we don't know what kind of comparison to run) then the user will get "no match at this address" error message, which is inaccurate.

I guess we could raise different exception types in this case?

This only comes up with CSV because it's now possible to set an orig entity with _only_ a symbol defined. You couldn't do that with code annotations because you always set the entity type when annotating something.

tests/test_csv.py
disinvite: Thanks, I'll try it that way. I don't like how this looks either.

docs/csv.md
disinvite: They're two different things. `symbol` is the linker name and we expect it to be unique. `name` is the name of the function, variable, or class (for vtables) as it appears in the code. This is probably unique but not guaranteed in the case of functions.

jonschz: I meant that it isn't obvious without deeper
reccmp knowledge what happens when specifying name or symbol, and what needs to be specified under which circumstances

reccmp/isledecomp/compare/core.py
jonschz: My suggestion:
- Make it explicit in the documentation that csv files can be used both to modify entities from code and to add new ones. For new ones, the entity type is mandatory. 
- Add an error text to the assertion, like _encountered an entity at address ... without a type. If this was specified in a csv file, please add a type column (if not present) and set the type for this entity_. I think it's okay to abort reccmp in this case for now.

docs/csv.md
disinvite: Oh I see. Thanks. I’ll add an example to the docs that shows the difference

tests/test_csv.py
```diff
@@ -0,0 +1,529 @@
...
+
+def test_docs_example_comments_and_blanks():
+    """Should ignore comment lines and blank lines"""
+    text = dedent(
+        """\
+        address,type
+
+        # Months of the year
+        100db57c,string
```
madebr, line 544: I think test_csv is missing dedicated unit tests for floats, strings and vtables (assuming the test binary even has these).

docs/csv.md
```diff
@@ -6,6 +6,14 @@ Any metadata set from a CSV file will overwrite the same data from code annotati
...
 If the same address is repeated on multiple lines the same CSV file, or across multiple files, the metadata will be overwritten each time.
 
+## Motivation
+
+The CSV feature is optional. In most situations, we can gather enough information to run analysis and provide a diff for each function using only the code annotations and PDB.
+
+However, loading data via CSV allows the user to specify all aspects of entity metadata, including attributes like the entity size, which is not currently possible with code annotations.
+
+You may find it helpful to copy data from Ghidra and use it to bootstrap a new decomp project. For example: CRT or library functions.
+
```
jonschz, line 16: 👍

reccmp/isledecomp/compare/core.py
disinvite: #252

disinvite:
I created some follow-up items so we can get this finished. Any objections to the `TextContainer` abstraction? I can copy it into the tests from #251.

Is it okay to use the backslash for escaping the quotechar and delimiter? I have test and doc examples showing how to use it.

I don't think escaped double quotes are likely to appear in any user data, but I don't want to leave a parsing gap either. I went for the escape character because I couldn't get consistent behavior with parsing:
```csv
address,name
1000,"""Hello, world"""
```
where the expected value is a name of `"Hello, world"` at addr 0x1000.

madebr:
I'm fine with merging this.

I'm currently tagging a binary using public sources. When that is finished, I will stress test this :)

PR #170, opened 2025-07-07
Use `pydemumble` module for MSVC demangling

disinvite:
We have been using the [`pydemangler`](https://github.com/wbenny/pydemangler) module for converting so-called mangled MSVC linked symbols into a more usable format. The three use cases are:
- Deriving the virtual inheritance class for a vtable, as in ``Act2Actor::`vftable'{for `LegoPathActor'}``.
- Getting the list of parameter types for an overloaded function so we can assign a more unique name.
- Getting the name of a vtordisp function. (This is no longer called as of #111)

`pydemangler` is downloaded from github; it is not published to pypi. I came across the [`pydemumble`](https://pypi.org/project/pydemumble/) module as a potential drop-in replacement. Both are based on the [`demumble`](https://github.com/nico/demumble) utility.

The only difference seems to be that a failure to demangle results in `None` with `pydemangler` but an empty string with `pydemumble`. That's easy enough to handle.

I don't see any regressions on the isledecomp binaries but I'll leave this as draft until we can do a more exhaustive comparison. (maybe throw every symbol from cvdump at both and compare what comes back)

### DIFF
--- a/reccmp/isledecomp/cvdump/demangler.py
+++ b/reccmp/isledecomp/cvdump/demangler.py
@@ -5,7 +5,7 @@ https://en.wikiversity.org/wiki/Visual_C%2B%2B_name_mangling
 
 import re
 from typing import NamedTuple
-import pydemangler  # type: ignore
+from pydemumble import demangle  # type: ignore
 
 
 class InvalidEncodedNumberError(Exception):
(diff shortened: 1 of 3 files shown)


jonschz (approved):
Since you already wrote a test, how about you make a proper test out of this using a hard-coded selection of ~10 symbols from LEGO1?

PR #173, opened 2025-07-08
Add `Image.seek()` function

disinvite:
Do we need a separate API for 16-bit segmented reads? Or should everything use `image.read(addr, size)`?

This PR proposes to add an abstract `seek()` function for all the `Image` subclasses. The function should take the provided address, find the correct block of virtual memory (i.e. the section), then return the `memoryview` at the position where we start reading. To account for uninitialized bytes, return the total number of bytes we are allowed to read, including whatever physical bytes exist.

This also moves `read()` and `read_string()` up into the `Image` superclass. The intended benefit is that we only need to implement one function to add support for reading from virtual memory of a particular image type.

What are the potential pitfalls of doing it this way? Or for handling 16-bit addresses in general?

One problem I predict is that if we store the 16-bit address by just shifting the segment left by 16 bytes, we need to careful about calculating the distance between two addresses if they are not from the same segment.


madebr:
Ghidra uses an abstract `Address` type and implements various interfaces:
- `GenericAddress`
- `SegmentedAddress`
- `SpecialAddress`

For reccmp, I think we will only potentially care about generic and segmented addresses.

But as @jonschz mentioned, it's unlikely we will soon/ever support segmented programs.
So I'm fine with doing whatever makes the code easy to work with.

reccmp/isledecomp/formats/image.py
```diff
@@ -1,9 +1,47 @@
...
+r_szstring = re.compile(rb"[^\x00]*\x00")
 
 
 @dataclasses.dataclass
 class Image:
     filepath: Path
     view: memoryview = dataclasses.field(repr=False)
     data: bytes = dataclasses.field(repr=False)
+
+    def seek(self, vaddr: int) -> tuple[bytes, int]:
```
jonschz, line 16: Is it possible to use an abstract base class (`abc`) here? Or does that not work because of the dataclass?

reccmp/isledecomp/formats/raw.py
jonschz, line 1: If this class is for testing only, I'd move it under `/tests` or inline it into `test_image_raw.py`

reccmp/isledecomp/formats/pe.py
```diff
@@ -927,26 +926,9 @@ def uninitialized_ranges(self) -> list[tuple[int, int]]:
     def addr_is_uninitialized(self, vaddr: int) -> bool:
         return any(start <= vaddr < end for start, end in self.uninitialized_ranges)
 
-    def read_string(self, vaddr: int, chunk_size: int = 1000) -> bytes:
```
jonschz, line 930: Does the removal of `chunk_size` have any noticable drawbacks? I'm happy with removing it, but I don't know if there was a good reason for it to be introduced in the first place.

disinvite: That was only there because I didn't know we could use a regular expression on a `memoryview` to capture the null-terminated string. It sort of made sense at the time to put a limit on the data being read while searching for the zero byte.

reccmp/isledecomp/formats/raw.py
disinvite: I moved it to the test file. It's a little out of place right now, but there is a limited use case later on: COM binaries in DOS.

reccmp/isledecomp/formats/image.py
disinvite: It probably should use `abc`, but let's save that for a followup where we can make more general changes.

PR #181, opened 2025-07-31
Match and name thunks using ref attribute

### LINKED ISSUE #176: Retain link between thunk entities and functions
We create thunk entities in the `match_thunks` function. They are named and matched at this point, so this needs to happen after we have created all the other function entities using annotations.

If we saved the link back to the original function address, we could break out these steps instead of doing them all at once.

This limitation is most visible when we have a thunk to an overloaded function. If we provide a more specific name for the overloaded function, we should update any thunks to use that name too.

We could set fields like `ref_orig` or `ref_recomp` to do this. (They can't use the same field name because it will be overwritten with the recomp value when the entity is matched.)

We could use a unique ID as our `ref` value that could be shared by both unmatched and matched entities. The problem is that the only unique IDs we have are orig and recomp addresses. Even if we used the database row ID, one of the rows gets deleted when the function is matched.

disinvite:
Continued from #177 and #180. We now establish entities for each incremental build thunk, but we don't provide a name right away. Because we now retain the link back to the parent entity, we can provide a name _after_ each overloaded function has been assigned a more unique name. This means that thunks to overloaded functions will use the unique name so they can be distinguished. (For example: on `BETA10` this has created new diffs in 36 functions, mostly surrounding 3D geometry functions.)

Fixes #176.

`roadmap` now shows that we establish an entity for each thunk whether we have annotated the parent or not. If we want to exclude these unnamed entities from the output, we can do that with a PR on `roadmap` specifically.

No new diffs on the retail builds on `isledecomp` because none of those were/are built incrementally.

### DIFF
--- a/reccmp/isledecomp/compare/queries.py
+++ b/reccmp/isledecomp/compare/queries.py
@@ -36,3 +36,37 @@ def get_overloaded_functions(db: EntityDb) -> Iterable[OverloadedFunctionEntity]
         assert isinstance(symbol, str) or symbol is None
         assert isinstance(nth, int)
         yield OverloadedFunctionEntity(orig_addr, recomp_addr, name, symbol, nth)
+
+
+class ThunkWithName(NamedTuple):
+    """Entity address(es) and the name (computed or base name)
+    of the referenced (thunked) entity."""
+
+    orig_addr: int | None
+    recomp_addr: int | None
+    name: str
+
+
+def get_thunks_and_name(db: EntityDb):
+    """For each entity with a ref_orig or ref_recomp attribute,
+    if the parent entity is a function, return:
+        - one or both addresses
+        - the name (or computed name) of the parent entity"""
+    for orig_addr, recomp_addr, name in db.sql.execute(
+        """SELECT e.orig_addr, e.recomp_addr,
+        coalesce(json_extract(r.kvstore, '$.computed_name'), json_extract(r.kvstore, '$.name')) name
+        FROM entities e
+        INNER JOIN entities r
+        ON e.ref_orig = r.orig_addr or e.ref_recomp = r.recomp_addr
+        WHERE name is not null
+        -- Do not return rows where (for example) orig_addr and ref_recomp are set.
+        -- If the entity has both ref_orig and ref_recomp, they must point to the same (matched) entity.
+        AND (e.orig_addr is null and e.ref_orig is null or e.ref_orig = r.orig_addr)
+        AND (e.recomp_addr is null and e.ref_recomp is null or e.ref_recomp = r.recomp_addr)
+        AND json_extract(r.kvstore, '$.type') = ?
+    """,
+        (EntityType.FUNCTION,),
+    ):
+        assert isinstance(orig_addr, int) or isinstance(recomp_addr, int)
+        assert isinstance(name, str)
+        yield ThunkWithName(orig_addr, recomp_addr, name)
--- a/reccmp/isledecomp/compare/queries.py
+++ b/reccmp/isledecomp/compare/queries.py
@@ -27,8 +27,11 @@ def get_overloaded_functions(db: EntityDb) -> Iterable[OverloadedFunctionEntity]
             select json_extract(kvstore,'$.name') as name from entities
             where json_extract(kvstore,'$.type') = ?
             and name is not null
+            and ref_orig is null and ref_recomp is null
             group by name having count(name) > 1
-        )""",
+        )
+        and ref_orig is null and ref_recomp is null
+        """,
         (EntityType.FUNCTION, EntityType.FUNCTION),
     ):
         assert isinstance(orig_addr, int) or isinstance(recomp_addr, int)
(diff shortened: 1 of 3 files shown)


reccmp/isledecomp/compare/queries.py
```diff
@@ -27,12 +27,49 @@ def get_overloaded_functions(db: EntityDb) -> Iterable[OverloadedFunctionEntity]
             select json_extract(kvstore,'$.name') as name from entities
             where json_extract(kvstore,'$.type') = ?
             and name is not null
+            and ref_orig is null and ref_recomp is null
```
jonschz, line 34: maybe add comments why this extra check is necessary

reccmp/isledecomp/compare/queries.py
```diff
@@ -27,12 +27,49 @@ def get_overloaded_functions(db: EntityDb) -> Iterable[OverloadedFunctionEntity]
...
+    for orig_addr, recomp_addr, name in db.sql.execute(
+        """SELECT e.orig_addr, e.recomp_addr,
+        coalesce(json_extract(r.kvstore, '$.computed_name'), json_extract(r.kvstore, '$.name')) name
+        FROM entities e
+        INNER JOIN entities r
+        ON e.ref_orig = r.orig_addr or e.ref_recomp = r.recomp_addr
+        WHERE name is not null
+        -- Do not return rows where (for example) orig_addr and ref_recomp are set.
+        -- If the entity has both ref_orig and ref_recomp, they must point to the same (matched) entity.
+        AND (e.orig_addr is null and e.ref_orig is null or e.ref_orig = r.orig_addr)
```
jonschz: Please add parentheses to make the precedence between `and` and `or` explicit

tests/test_compare_db_queries.py
```diff
@@ -60,3 +63,148 @@ def test_overloaded_functions(db: EntityDb):
...
+    """Should follow the ref_orig or ref_recomp attribute back to the
+    parent entity to derive the thunk name."""
+    with db.batch() as batch:
+        batch.set_orig(100, name="Hello", type=EntityType.FUNCTION)
+        batch.set_orig(200, ref_orig=100)
+        batch.set_recomp(500, name="Test", type=EntityType.FUNCTION)
+        batch.set_recomp(600, ref_recomp=500)
+
+    names = list(get_thunks_and_name(db))
+    assert len(names) == 2
```
jonschz, line 97: Minor stylistic point: Can't you assert equality on arrays? Like `assert names == [...]`

tests/test_compare_db_queries.py
```diff
@@ -60,3 +63,148 @@ def test_overloaded_functions(db: EntityDb):
...
+        batch.match(100, 500)
+        batch.match(200, 600)
+
+    names = list(get_thunks_and_name(db))
+    assert len(names) == 1
+    assert names[0].name == "Test"  # Prefer recomp value
+
+
+def test_named_thunks_no_name(db: EntityDb):
+    """Do not return a name unless the parent entity has one."""
```
jonschz, line 122: The semantics sound reasonable, but the function name could be chosen such that it reflects the behaviour better. Maybe rename `get_thunks_and_name` to something like `get_named_thunks` or `get_thunks_with_name`?

tests/test_compare_db_queries.py
```diff
@@ -60,3 +63,148 @@ def test_overloaded_functions(db: EntityDb):
...
+        batch.set_recomp(500, name="Test", type=EntityType.FUNCTION)
+        batch.set_recomp(600, ref_orig=100)
+
+    names = list(get_thunks_and_name(db))
+    assert len(names) == 0
+
+
+def test_named_thunks_crossed_same_addr(db: EntityDb):
+    """The same should be true even if the address values are the same
+    in both virtual address sapces."""
```
jonschz: typo

tests/test_compare_db_queries.py
```diff
@@ -60,3 +63,148 @@ def test_overloaded_functions(db: EntityDb):
...
+        batch.set_recomp(500, name="Test", type=EntityType.FUNCTION)
+        batch.set_recomp(600, ref_recomp=500)
+        # Only thunk is matched
+        batch.match(200, 600)
+
+    names = list(get_thunks_and_name(db))
+    assert len(names) == 0
+
+
+def test_named_thunks_ignore_incomplete_if_matched(db: EntityDb):
```
jonschz, line 191: Should we log warnings in these cases? Right now I am not sure if these issues can only occur due go bugs or also due to incorrect annotations.

tests/test_compare_db_queries.py
disinvite: Yes, but I usually do it this way so we don't have to rebuild the tuple (or namedtuple) and can just check the most pertinent fields. I think this also makes the `pytest` output shorter instead of printing the entire list, but I'd have to go back and check that.

tests/test_compare_db_queries.py
disinvite: Yes. I don't think `match_ref` leaves any gap where this (`ref_orig` and `ref_recomp` pointing to different entities) could happen for the moment. Can we add the check in a follow-up or should it go with this PR?

jonschz: I'm fine with opening a follow-up issue

PR #188, opened 2025-08-13
Recognize widechar strings under `STRING` annotations

disinvite:
First stage of detecting and matching unicode strings. Actually reading the data and decoding to either UTF-16 LE or BE will come next. (Might move #173 up first to support it.)

### DIFF
--- a/reccmp/isledecomp/compare/core.py
+++ b/reccmp/isledecomp/compare/core.py
@@ -343,8 +343,11 @@ class Compare:
                 # Not that we don't trust you, but we're checking the string
                 # annotation to make sure it is accurate.
                 try:
-                    # TODO: would presumably fail for wchar_t strings
-                    orig = self.orig_bin.read_string(string.offset).decode("latin1")
+                    if string.is_unicode:
+                        orig = None  # TODO
+                    else:
+                        orig = self.orig_bin.read_string(string.offset).decode("latin1")
+
                     string_correct = string.name == orig
                 except UnicodeDecodeError:
                     string_correct = False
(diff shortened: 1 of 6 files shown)


reccmp/isledecomp/parser/util.py
```diff
@@ -117,18 +117,20 @@ def get_variable_name(line: str) -> str | None:
     return None
 
 
-def get_string_contents(line: str) -> str | None:
+def get_string_contents(line: str) -> tuple[str | None, bool]:
```
jonschz: Maybe document the return type in the docstring or create a `NamedTuple` for it

reccmp/isledecomp/parser/util.py
```diff
@@ -20,7 +20,7 @@
 regularCommentRegex = re.compile(r"(//.*)")
 
 # Get string contents, ignore escape characters that might interfere
-doubleQuoteRegex = re.compile(r"(\"(?:[^\"\\]|\\.)*\")")
+doubleQuoteRegex = re.compile(r"(L)?(\"(?:[^\"\\]|\\.)*\")")
```
jonschz: I just realised that this regex can be simplified a bit by enclosing it in single quotes.

`r'(L)?("(?:[^"\\]|\\.)*")'` should be equivalent (assuming our linter does not complain, but even then I'd add an exception)

reccmp/isledecomp/compare/core.py
```diff
@@ -343,8 +343,11 @@ def orig_bin_checker(addr: int) -> bool:
                 # Not that we don't trust you, but we're checking the string
                 # annotation to make sure it is accurate.
                 try:
-                    # TODO: would presumably fail for wchar_t strings
-                    orig = self.orig_bin.read_string(string.offset).decode("latin1")
+                    if string.is_unicode:
+                        orig = None  # TODO
```
jonschz, line 347: Maybe write an issue and link it here?

jonschz (approved):
Looks good!

PR #192, opened 2025-08-19
Match widechar strings, improve PDB string detection

### LINKED ISSUE #189: Matching widechar strings
After merging #188, the main thing missing is a function to read widechar strings from the image. (#173 sets this up)

disinvite:
Fixes #189.

This adds the `read_widechar` function to the `Image` base class for reading 2-byte character strings. The `core` module assumes UTF-16 LE encoding to start with, but this function should work if we need to support big endian with other image types. Now that we can read widechar data and get the intended text, we can match widechar strings annotated in code.

The second change is more subtle. To detect strings in the PDB, we rely on string symbols created for images built with the string pooling option enabled. This is the only thing I'm aware of that will tell whether a string is widechar or not. These symbols also include an approximation of the length of the string, but it does not consistently account for the null terminator. I realized that the section contribution data provides a more consistent length, so we now use that if we have it.

This enables us to include strings that include null bytes or are entirely null bytes. In the latter case, we had been reducing those strings to the empty string with `rstrip()`. Another reason they were ignored until now is that our JSON database does not support strings that contain nulls. The escaping added by #191 makes it possible to create these entities.

The strings `"\x00"` and `L"\x00"` are used in the library functions `___crtLCMapStringA` and `___crtGetStringTypeA`. You can see the strings in the `reccmp-roadmap` output with `--verbose` enabled.

### DIFF
--- a/reccmp/isledecomp/compare/core.py
+++ b/reccmp/isledecomp/compare/core.py
@@ -199,26 +199,17 @@ class Compare:
                         )
                         continue
 
-                    # TODO: skip unicode for now. will need to handle these differently.
-                    if string_info.is_utf16:
-                        continue
-
-                    size = sym.size()
-                    assert size is not None
-
-                    raw = self.recomp_bin.read(addr, size)
-
                     try:
-                        # We use the string length reported in the mangled symbol as the
-                        # data size, but this is not always accurate with respect to the
-                        # null terminator.
-                        # e.g. ??_C@_0BA@EFDM@MxObjectFactory?$AA@
-                        # reported length: 16 (includes null terminator)
-                        # c.f. ??_C@_03DPKJ@enz?$AA@
-                        # reported length: 3 (does NOT include terminator)
-                        # This will handle the case where the entire string contains "\x00"
-                        # because those are distinct from the empty string of length 0.
-                        decoded_string = raw.decode("latin1")
+                        # String size is the total memory footprint, including null-terminator.
+                        if string_info.is_utf16:
+                            raw = self.recomp_bin.read_widechar(addr)
+                            decoded_string = raw.decode("utf-16-le")
+                            string_size = len(raw) + 2
+                        else:
+                            raw = self.recomp_bin.read_string(addr)
+                            decoded_string = raw.decode("latin1")
+                            string_size = len(raw) + 1
+
                         rstrip_string = decoded_string.rstrip("\x00")
 
                         # TODO: Hack to exclude a string that contains \x00 bytes
@@ -241,7 +232,7 @@ class Compare:
                             rstrip_string, wide=string_info.is_utf16
                         ),
                         symbol=sym.decorated_name,
-                        size=len(rstrip_string) + 1,
+                        size=string_size,
                         verified=True,
                     )
                 else:
@@ -346,9 +337,13 @@ class Compare:
                 # annotation to make sure it is accurate.
                 try:
                     if string.is_unicode:
-                        orig = None  # TODO
+                        raw = self.orig_bin.read_widechar(string.offset)
+                        orig = raw.decode("utf-16-le")
+                        string_size = len(raw) + 2
                     else:
-                        orig = self.orig_bin.read_string(string.offset).decode("latin1")
+                        raw = self.orig_bin.read_string(string.offset)
+                        orig = raw.decode("latin1")
+                        string_size = len(raw) + 1
 
                     string_correct = string.name == orig
                 except UnicodeDecodeError:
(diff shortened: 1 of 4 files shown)


reccmp/isledecomp/formats/image.py
```diff
@@ -27,9 +30,23 @@ def read_string(self, vaddr: int) -> bytes:
...
+        # Since we are dealing with only UTF-16 LE at the outset,
+        # just make sure we don't deliver a string with an odd number of bytes
+        # that will definitely fail to decode.
+        # Later on we may want to have each image type do things differently
+        # based on the expectations of the platform.
+        if match and len(match.group(0)) % 2 == 0:
+            return match.group(0)
 
-        return bytes(view)
+        return b""
```
jonschz: Maybe log a warning here?

reccmp/isledecomp/compare/core.py
```diff
@@ -199,37 +199,38 @@ def _load_cvdump(self):
...
-                    raw = self.recomp_bin.read(addr, size)
-
                     try:
-                        # We use the string length reported in the mangled symbol as the
-                        # data size, but this is not always accurate with respect to the
-                        # null terminator.
-                        # e.g. ??_C@_0BA@EFDM@MxObjectFactory?$AA@
-                        # reported length: 16 (includes null terminator)
-                        # c.f. ??_C@_03DPKJ@enz?$AA@
-                        # reported length: 3 (does NOT include terminator)
```
jonschz, line 219: Did you remove these examples intentionally?

reccmp/isledecomp/compare/core.py
```diff
@@ -199,37 +199,38 @@ def _load_cvdump(self):
...
+                        if string_info.is_utf16:
+                            if sym.section_contribution is not None:
+                                string_size = sym.section_contribution
+                                # Remove 2-byte null-terminator before decoding
+                                raw = self.recomp_bin.read(addr, string_size)[:-2]
+                            else:
+                                raw = self.recomp_bin.read_widechar(addr)
+                                string_size = len(raw) + 2
+
+                            decoded_string = raw.decode("utf-16-le")
```
jonschz, line 227: Are there any foreseeable use cases for big endian? I don't see any right now. Either way, I'd add a comment that we are only targeting little-endian architectures at the moment.

reccmp/isledecomp/compare/core.py
```diff
@@ -346,9 +347,13 @@ def orig_bin_checker(addr: int) -> bool:
                 # annotation to make sure it is accurate.
                 try:
                     if string.is_unicode:
```
jonschz: Not your change, but how about we rename `is_unicode` to `is_widechar` or `is_utf16`? I think newer C++ versions support UTF-8 as well, which will make `is_unicode` a confusing identifier.

Applies to the whole dependency tree (e.g. `ParserCodeString`, `ParserString`, `util.py:134`).

reccmp/isledecomp/compare/core.py
disinvite: Yes, but only because we aren't using that length value embedded in the symbol any more. I can add them back to illustrate why the value is sometimes off by one.

reccmp/isledecomp/formats/image.py
disinvite: I thought about that while I was figuring this out. With the two regular expressions in their current state, I don't think it's possible to hit the two `return b""` lines. We should always get a match even if it has length zero.

We could use `assert match is not None` here, but I'm leery about using those too often. We've had a few recent cases of fatal errors leaving users with no option to continue except patching `reccmp`.

If we're going to throw an exception, we might as well verify the expected encoding too.

reccmp/isledecomp/compare/core.py
disinvite: There are probably not any big-endian strings on the horizon but I wasn't 100% sure.

reccmp/isledecomp/formats/image.py
jonschz: Alright. In that case I'd log an error but not crash, so it's obvious that something happened that we did not expect

reccmp/isledecomp/compare/core.py
jonschz: Alright, then I'd just stick with the comment

reccmp/isledecomp/formats/image.py
disinvite: I put in a new string exception, we trap it in `core.py` and then write to the log if it happens. I can't think of a way to _make_ the problem happen so there's no test to check for it.

jonschz: I think exhaustiveness is good enough - if the case is sufficiently unexpected but theoretically possible, an error log should do

PR #193, opened 2025-08-19
Don't create unified diff in function comparator

disinvite:
Refactors the function comparator for better separation of concerns. `compare_function` now returns only the `FunctionCompareResult` tuple with the `SequenceMatcher` opcodes and other information. The `core.py` module creates the `DiffReport` object that is used for JSON/HTML/terminal output.

We now return the complete set of asm and opcodes for all functions, even if the match is 100%. At some point we will want the number of context lines to be user-configurable. We might also want the option to show the asm we generate for 100% matched functions. For now, the output is the same, but the call to restrict the diff to 10 lines of context (or none at all, for perfect matches) happens in `core.py`.

This also moves the `DiffOpcode` type and our copy of `get_grouped_opcodes` to its own `difflib` module to prevent circular imports.

Since `compare_function` no longer generates unified diff output, the unit tests in `test_function_comparator` have less intricate data.

### DIFF
--- a/reccmp/isledecomp/compare/core.py
+++ b/reccmp/isledecomp/compare/core.py
@@ -835,10 +835,10 @@ class Compare:
             assert best_name is not None
 
             diff_result = self.function_comparator.compare_function(match)
-            if diff_result.diff:
-                grouped_codes = list(get_grouped_opcodes(diff_result.diff[0], n=10))
+            if diff_result.match_ratio != 1.0:
+                grouped_codes = list(get_grouped_opcodes(diff_result.codes, n=10))
                 udiff = combined_diff(
-                    grouped_codes, diff_result.diff[1], diff_result.diff[2]
+                    grouped_codes, diff_result.orig_inst, diff_result.recomp_inst
                 )
             else:
                 udiff = None
--- a/reccmp/isledecomp/compare/core.py
+++ b/reccmp/isledecomp/compare/core.py
@@ -5,6 +5,7 @@ from pathlib import Path
 import struct
 from typing import Iterable, Iterator
 from reccmp.project.detect import RecCmpTarget
+from reccmp.isledecomp.difflib import get_grouped_opcodes
 from reccmp.isledecomp.compare.functions import FunctionComparator
 from reccmp.isledecomp.formats.detect import detect_image
 from reccmp.isledecomp.formats.pe import PEImage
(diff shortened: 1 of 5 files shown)


tests/test_function_comparator.py
```diff
@@ -222,89 +238,14 @@ def test_impact_of_line_annotation(
...
-            ],
-        ),
+    assert diffreport.codes == [
+        ("equal", 0, 2, 0, 2),
+        ("replace", 2, 9, 2, 3),
+        ("replace", 9, 10, 3, 4),
+        ("equal", 10, 11, 4, 5),
+        ("replace", 11, 14, 5, 8),
+        ("equal", 14, 15, 8, 9),
+        ("replace", 15, 19, 9, 22),
```
jonschz: Did you drop the assertion on the decomps intentionally? It may have been a reasonable choice, just asking.

reccmp/isledecomp/compare/core.py
```diff
@@ -830,7 +831,27 @@ def _compare_match(self, match: ReccmpMatch) -> DiffReport | None:
...
+
+            return DiffReport(
+                match_type=EntityType.FUNCTION,
+                orig_addr=match.orig_addr,
+                recomp_addr=match.recomp_addr,
+                name=best_name,
+                udiff=udiff,
+                ratio=diff_result.match_ratio,
+                is_effective_match=diff_result.is_effective_match,
+            )
```
jonschz, line 859: What I don't quite like is the size increase of `core.py` which I'd like to keep smaller. Maybe we pull this out into a helper function? Doesn't have to move back into the `function_comparator`

tests/test_function_comparator.py
disinvite: I dropped it because the main thing under test is the change to the diff opcodes. However: I expected the asm output to be the same as the previous function, but these two lines are changed:
```
("0x20c", "jmp cppfile.cpp:384 (LINE)"),
("0x407", "mov eax, dword ptr [ebp - 0xc] \t(test.cpp:123, pinned)"),
```
I put both instruction blocks back in just to be comprehensive.

reccmp/isledecomp/compare/core.py
disinvite: I think we should leave it because this (producing the diff report) is one of the things that definitely belongs in the `Compare` module. In contrast, functions that match or detect entities are good candidates for helper functions. Some of those are already in modules like `match_msvc.py` but we just need to do the rest.

PR #196, opened 2025-08-22
HTML table refactor

disinvite:
This makes the following changes:

- The table of entities (internal name: `listing-table`) is now completely controlled by the javascript component code. It had used the `<thead>` element in `template.html` as a sort of hybrid, but now all the internal HTML is set in code.
  - ~Doing it this way means we need to escape the entity name field: template functions will contain angle brackets that disrupt the HTML. There is a new test (and new dummy data) to check for this.~
- Removed the `func-row` and `diff-row` components. They were wrappers for the table rows. This was technically not valid HTML because only `<tr>` elements should be children of `<table>` or `<tbody>`.
   - As a result of this, the query selectors used in the playwright tests now look for rows with the `data-address` or `data-diff` attributes. This still isn't perfect (they would prefer you use `getByRole('row')`) but it's a little better.
- The "show recomp column" option now rewrites the HTML instead of using CSS classes to hide the column in place. The `colspan` used in the diff display rows is now accurate to the number of columns that are there. The upshot here is that the playwright tests succeed even if you remove the `<style>` section completely.

- Custom components no longer use shadow DOM or `<template>` / `<slot>` elements. There wasn't a pressing reason to use them except for maybe the clipboard component. The old design was just hacked together.
- Removed the `no-diff` component. It's just a `<div>` with text.

We end up with less code and redraw performance that should be the same or better. (I'll verify this again before merging just for style points.)

My end goal with this is to give users/teams the option to customize the HTML output. We provide a series of custom components to display the data from the reccmp report and you position them on the page however you like, using whatever CSS style you want.

### DIFF
--- a/reccmp/assets/reccmp.js
+++ b/reccmp/assets/reccmp.js
@@ -144,15 +144,6 @@ function copyToClipboard(value) {
   navigator.clipboard.writeText(value);
 }
 
-function escapeHtml(unsafe) {
-  return unsafe
-    .replace(/&/g, '&amp;')
-    .replace(/</g, '&lt;')
-    .replace(/>/g, '&gt;')
-    .replace(/"/g, '&quot;')
-    .replace(/'/g, '&#039;');
-}
-
 const PAGE_SIZE = 200;
 
 //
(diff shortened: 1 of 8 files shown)


disinvite:
> can we arrange by diff count now too? 👀

Not here, but I can add that in a follow-up.

reccmp/assets/reccmp.js
```diff
@@ -144,6 +144,15 @@ function copyToClipboard(value) {
...
 }
 
+function escapeHtml(unsafe) {
+  return unsafe
+    .replace(/&/g, '&amp;')
+    .replace(/</g, '&lt;')
+    .replace(/>/g, '&gt;')
+    .replace(/"/g, '&quot;')
+    .replace(/'/g, '&#039;');
+}
```
jonschz: How about building the DOM using the API and setting the string using `.textContent` as suggested in https://stackoverflow.com/a/66481918 (probably where this code came from)? We can restrict the refactor to the critical parts for now.

disinvite: Yeah I wasn't thrilled about needing that function either. I switched it back to the element API for now.

PR #198, opened 2025-09-01
Fix Ghidra namespace import paths

### LINKED ISSUE #16: Refactor isle-specific parts in Ghidra import
This is the one place that is still isle-specific. Please tell me if you want me to find a solution before merging.

_Originally posted by @jonschz in https://github.com/isledecomp/reccmp/pull/12#discussion_r1817995984_

### LINKED ISSUE #146: Ghidra import: Wrong `this` argument for functions of nested classes/structs
Case in point:
```c++
class LegoGameState {
	struct History {
		History();
		void WriteScoreHistory();
		// ...
        }
        // ...
}
```
The Ghidra import correctly creates a type named `LegoGameState::History`, but the function `WriteScoreHistory()` expects a `this` argument of type `History`, which it doesn't find. Thus, it creates a new, empty one, and the imported type is not used.

jonschz:
This is a larger Ghidra import refactor addressing two main issues:
- #146 
- #16

There are a few changes in the behaviour of the Ghidra importer:
- Types are no longer imported under `/imported` but under the root of e.g. `LEGO1.DLL/`.
- Types inside of namespaces now get their namespaces mapped to type categories, for example: `Act1State::NamedPlane` is now imported under `/Act1State/NamedPlane`.
  - This is the change that fixes the issue observed in #146.
- Regex substitutions in function imports are now supported. This was the last ISLE-specific code in the `reccmp` project (as far as I am aware)
  - A PR for `isle` will follow. The additional config looks like
    ```yaml
    name-substitutions:
      FUN_([0-9a-f]{8}): LEGO1_\1
    ```

Other than that, there are a few refactors making the Ghidra import more robust and type-safe, mostly with respect to entities (types or functions) inside namespaces or classes.

fixes #146, fixes #16

### DIFF
--- a/reccmp/ghidra_scripts/import_functions_and_types_from_pdb.py
+++ b/reccmp/ghidra_scripts/import_functions_and_types_from_pdb.py
@@ -267,7 +267,7 @@ def log_and_track_failure(
             step_name,
             "Unexpected error: " if unexpected else "",
             error,
-            exc_info=error
+            exc_info=error,
         )
 
 


jonschz:
I have updated the Ghidra repository of `LEGO1` with this PR branch so you can review the effect of this change in Ghidra without having to set it up.

reccmp/ghidra_scripts/lego_util/entity_names.py
jonschz, line 1: This is now a separate file without ghidra imports so it can be unit tested

disinvite (approved):
I don't have time to test it myself until the weekend, but everything here looks good. My one comment is that we may want to run an `re.compile` on the substitution regexes before starting the import, just to make sure each one is valid.

Aside from that, feel free to merge.

jonschz:
I think you don't necessarily need to run it yourself, but it'd be great if you could check out Ghidra and pick a few random functions to look for regressions (LEGO1 and BETA10)

reccmp/ghidra_scripts/import_functions_and_types_from_pdb.py
```diff
@@ -137,8 +139,10 @@ def import_function_into_ghidra(
     api: "FlatProgramAPI",
     pdb_function: "PdbFunction",
     type_importer: "PdbTypeImporter",
-    name_substitutions: dict[str, str],
+    name_substitutions: CompiledRegexReplacements,
 ):
+    logger.debug("Start handling function '%s'", pdb_function.match_info.best_name())
+
```
jonschz, line 145: New drive-by: The log didn't not show if `PdbFunctionImporter` throws

reccmp/ghidra_scripts/lego_util/function_importer.py
```diff
@@ -128,6 +134,13 @@ def __init__(
         self.return_type = type_importer.import_pdb_type_into_ghidra(
             self.signature.return_type
         )
+
+        if "T_NOTYPE(0000)" in self.signature.arglist:
+            # Variadric functions have a T_NOTYPE as their last argument
+            raise TypeNotImplementedError(
+                f"Function '{self.get_full_name()}' is probably variadric, which is not implemented yet."
+            )
```
jonschz, line 142: New drive-by: Better error for variadric functions

Issue #199, opened 2025-09-03
Over-eager effective function matching


jeff-1amstudios:
I'm looking into the effective matching code to add some extra cases I've seen in https://github.com/dethrace-labs/dethrace.

I'm going to add some tests as they seem to be missing in this area, and came across the following test which fails, but should succeed.

```python
def test_basic_jmp_cmp():
    orig_asm = ["cmp eax, ecx", "jg 0x1"]
    recomp_asm = ["cmp ecx, eax", "jg 0x1"]

    diff = difflib.SequenceMatcher(None, orig_asm, recomp_asm)
    assert diff.ratio != 1.0

    is_effective = find_effective_match(
        diff.get_opcodes(), orig_asm, recomp_asm
    )
    assert is_effective == False
```

This code is obvious non-matching, it should only effectively match if recomp_asm contained a `jl` instead of `jg`.

What happens instead is the cmp_swap is correctly not happening, but `naive_register_replacement` marks the first line as correct, then the second line is an exact match, so we show the function as an effective match, which isn't correct.

This seems to be a dangerous behavior with `naive_register_replacement` where it could easily(?) mask problems where an incorrect register was accessed. Am I missing something?

jeff-1amstudios:
Even this example is marked as effective - note the registers are all different

```python
orig_asm = ["cmp eax, ebx", "jg 0x1"]
recomp_asm = ["cmp ecx, edx", "jg 0x1"]
```

jeff-1amstudios:
Maybe a more correct/safe implementation is to track the usage of each register across the function, ensuring it appears in the same place in each line?

eg in the orig "eax" becomes "1", "ebx" becomes "2". In the recmp "ecx" becomes "1", "edx" becomes "2". Now we compare the sanitized form, which respects the order and placement of registers

orig_asm = ["cmp 1, 2", "jg 0x1"]
recomp_asm = ["cmp 1, 2", "jg 0x1"]

jeff-1amstudios:
Added an xfailing test case for it - `test_fixes::test_non_effective_register_replacement_should_not_be_marked_as_effective`

jeff-1amstudios:
Another example we talked about in Discord is the `relocate_instructions` fixup.

An example which was marked as "effective match" but shouldn't be is where code is the same by moved around. There are only a few specific scenarios where this could be safely marked as "effective" and this example is not one of them!

[...]

PR #200, opened 2025-09-05
Adds handling of compiler entropy for mov-cmp-jmp and tests

jeff-1amstudios:
In addition to handling 
```
- cmp eax, ebx,
- jg 0x1
+ cmp ebx, eax
+ jl 0x1
```,

we now handle
```
- mov eax, dword ptr [ebp-4],
- cmp dword ptr [gCurrent_key (DATA)], eax,
- jg 0x1
+ mov eax, dword ptr [gCurrent_key (DATA)],
+ cmp dword ptr [ebp-4], eax,
+ jl 0x1
```

And added tests for both cases

### DIFF
--- a/reccmp/isledecomp/compare/asm/fixes.py
+++ b/reccmp/isledecomp/compare/asm/fixes.py
@@ -43,24 +43,7 @@ def is_operand_swap(a: str, b: str) -> bool:
     return a.partition(", ")[0] != b.partition(", ")[0] and sorted(a) == sorted(b)
 
 
-def can_cmp_swap(orig: list[str], recomp: list[str]) -> bool:
-    # Make sure we have 1 cmp and 1 jmp for both
-    if len(orig) != 2 or len(recomp) != 2:
-        return False
-
-    if not orig[0].startswith("cmp") or not recomp[0].startswith("cmp"):
-        return False
-
-    if not orig[1].startswith("j") or not recomp[1].startswith("j"):
-        return False
-
-    # Checking two things:
-    # Are the cmp operands flipped?
-    # Is the jump instruction compatible with a flip?
-    return is_operand_swap(orig[0], recomp[0]) and jump_swap_ok(orig[1], recomp[1])
-
-
-def patch_jump(a: str, b: str) -> str:
+def get_patched_jump(a: str, b: str) -> str:
     """For jump instructions a, b, return `(mnemonic_a) (operand_b)`.
     The reason to do it this way (instead of just returning `a`) is that
     the jump instructions might use different displacement offsets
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

**Comment by jonschz** [comment]
That is not what I expected - the matches actually only increased:
```
LEGO1:
Increased (5):
0x100019d0 - Score::Paint (99.15% -> 100.00%*) -> correct
0x10044270 - MxControlPresenter::CheckButtonDown (94.18% -> 100.00%*) -> correct
0x1006f4e0 - Infocenter::ReadyWorld (96.75% -> 100.00%*) -> possible false negative, see below
0x10077210 - RegistrationBook::HandleEndAction (96.43% -> 100.00%*) -> correct
0x100ccd00 - MXIOINFO::Advance (96.24% -> 100.00%*) -> correct (looks a bit like the second case, but the two matches are unrelated)

BETA10:
Increased (3):
0x1012078c - MxDirectDraw::SetPaletteEntries (97.66% -> 100.00%*) -> correct
0x1015e6c4 - MXIOINFO::Seek (99.12% -> 100.00%*) -> correct
0x1015ea3e - MXIOINFO::Flush (98.00% -> 100.00%*) -> correct
```
I derived another false negative from `Infocenter::ReadyWorld` which should not match effectively, but it does:
```py
    orig_asm = [
        "mov eax, dword ptr [esi + 0x100]",
        "mov ecx, dword ptr [eax + 0x74]",
        "add eax, 0x74",
        "sub ecx, 3",
        "cmp ecx, 0xc",
        "ja 0x0",
        "mov eax, 0",
        "mov ecx, 1",
        "mov dword ptr [eax], 2",
    ]
    recomp_asm = [
        "mov ecx, dword ptr [esi + 0x100]",
        "mov eax, dword ptr [ecx + 0x74]",
        "add ecx, 0x74",
        "sub eax, 3",
        "cmp eax, 0xc",
        "ja 0x0",
        "mov eax, 0",
        "mov ecx, 1",
        "mov dword ptr [ecx], 2",
    ]
```
The first instructions can be fixed by exchanging `ecx` and `eax`. However, those are changed again in matching code, so they cannot be exchanged below `mov eax, 0`. However, the effective matcher ignores that they have been changed in matching code in between.

It also looks like this issue might not be due to the comparison. The following is a false negative as well:
```py
    orig_asm = [
        "mov eax, 0",
        "mov ecx, 1",
        "mov dword ptr [eax], 2",
    ]
    recomp_asm = [
        "mov eax, 0",
        "mov ecx, 1",
        "mov dword ptr [ecx], 2",
    ]
```
* [ ] Could you check if this issue (or a related one) affects the current `master` as well?
  * [ ] If it does, please open a follow-up ticket and add these as `xfail` test cases.
  * [ ] If it does not, please fix it in your PR.

reccmp/isledecomp/compare/asm/fixes.py
```diff
@@ -95,15 +155,12 @@ def patch_cmp_swaps(
...
-                patched = patch_jump(orig_asm[i + 1], recomp_asm[j + 1])
-                # We only register a fix if it actually matches
-                if orig_asm[i + 1] == patched:
-                    fixed_lines.add(j + 1)
+            for fn in patch_fns:
+                this_patch_lines = fn(orig_asm[i : i + 3], recomp_asm[j : j + 3])
+                # if we have fixed lines by this patcher, add them to the combined `fixed_lines`
+                if len(this_patch_lines) > 0:
+                    fixed_lines.update([j + x for x in this_patch_lines])
+                    break
```
jonschz, line 170: Could you explain the reasoning for this `break`? I don't find it obvious why you would skip the second patch function if the first one finds anything

reccmp/isledecomp/compare/asm/fixes.py
```diff
@@ -95,15 +155,12 @@ def patch_cmp_swaps(
...
-                # Patch cmp
-                fixed_lines.add(j)
-
-                # Patch the jump if necessary
-                patched = patch_jump(orig_asm[i + 1], recomp_asm[j + 1])
-                # We only register a fix if it actually matches
-                if orig_asm[i + 1] == patched:
-                    fixed_lines.add(j + 1)
+            for fn in patch_fns:
+                this_patch_lines = fn(orig_asm[i : i + 3], recomp_asm[j : j + 3])
```
jonschz: This was an issue in the original code, but now it is even more obfuscated: The `3` is sort of a magic value that needs some explanation (or, even better, assign it to a self-explanatory variable)

tests/test_fixes.py
```diff
@@ -0,0 +1,69 @@
+import difflib
+from reccmp.isledecomp.compare.asm.fixes import find_effective_match
+
+
```
jonschz, line 5: Could you also add a test case asserting that `jg 0x1` vs. `jle 0x1` will not be accepted as effective?

jeff-1amstudios: Sure, done 👍

jeff-1amstudios:
> That is not what I expected - the matches actually only increased
Why did you not expect that? That is what I _would_ expect. The only thing that changes ideally is we handle one more case of compiler entropy, so either the matches should remain exactly the same (if that mov/cmp/jmp) pattern is not used) or increase.

For `Infocenter::ReadyWorld`, I think that is an example of this issue which I flagged - https://github.com/isledecomp/reccmp/issues/199 - but I'll check the behavior against master now

jeff-1amstudios:
@jonschz confirmed this behavior exists in master, and added an `xfailing` test for it. I've updated #199 to include a reference to that test

jonschz:
> Why did you not expect that?

I misread your PR and thought that you had attempted to fix (part of) #199. Now that I read it again, it makes a lot more sense - you added additional matching patterns but didn't fix any false effective matches.

reccmp/isledecomp/compare/asm/fixes.py
jeff-1amstudios: Oops missed this one. Functionally it wouldn't make any difference, its more of an optimization. If we've already matched the lines, no need to check the other strategies for matching the lines.

jonschz: > If we've already matched the lines, no need to check the other strategies for matching the lines.

Could you add that as a comment? Then we're good to merge

jeff-1amstudios: Sure, will do

jeff-1amstudios: Done!

PR #203, opened 2025-09-16
Use `TypedDict` for cvdump types parser

disinvite:
The `keys` dict in the cvdump types parser stores all the information about the type leaves we parse. The conglomeration of attributes is now listed under `CvdumpParsedType` as a `TypedDict`. On top of adding type-checking, I made these changes to the existing format:

- Leaves that reference an `LF_FIELDLIST` use the key `"field_list_type"`, except `LF_ENUM` which used `"field_type"`. Now they all use `"field_list_type"`.
- `LF_MEMBER` values now use the namedtuple `FieldListItem` instead of a `dict`. We were already using this type to create the `TypeInfo` object, so we just set it up earlier.
- Similarly, `LF_ENUMERATE` values under an `LF_FIELDLIST` now use `EnumItem` instead of a `dict`.
- Numeric fields such as `"this_adjust"` and `"num_params"` are now converted to `int` instead of being left as `str` when read off the regex match groups. `"this_adjust"` is a hex value (although this may not be obvious at first glance) and there is a new test to make sure we use the correct value.

I also changed the Ghidra scripts to make use of the new types. The `LF_MEMBER` dict had used the `"type"` key to refer to both the cvdump hex key _and_ the Ghidra `DataType` struct interchangeably. There is a new namedtuple `GhidraFieldListItem` to make the conversion more clear.

The last thing was to add the mock type `CvdumpTypeKey` to mark where we expect this instead of just a string. For now it simply resolves to `str`.

### DIFF
--- a/reccmp/isledecomp/cvdump/types.py
+++ b/reccmp/isledecomp/cvdump/types.py
@@ -703,9 +703,9 @@ class CvdumpTypesParser:
         return {
             "type": leaf_type,
             "return_type": match.group("return_type"),
-            "call_type": match.group("call_type"),
             "class_type": match.group("class_type"),
             "this_type": match.group("this_type"),
+            "call_type": match.group("call_type"),
             "func_attr": match.group("func_attr"),
             "num_params": int(match.group("num_params")),
             "arg_list_type": match.group("arg_list_type"),
--- a/reccmp/isledecomp/cvdump/types.py
+++ b/reccmp/isledecomp/cvdump/types.py
@@ -334,7 +334,7 @@ class CvdumpTypesParser:
         members: list[FieldListItem] = []
 
         if "super" in field_obj:
-            for super_id, _ in field_obj["super"].items():
+            for super_id in field_obj["super"].keys():
                 # May need to resolve forward ref.
                 superclass = self.get(super_id)
                 if superclass.members is not None:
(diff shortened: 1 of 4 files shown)


reccmp/isledecomp/cvdump/types.py
```diff
@@ -1,12 +1,21 @@
...
 
 logger = logging.getLogger(__name__)
 
 
+# TODO: For now, this is here just to document where we
+# expect a T_* (scalar) or 0x____ (complex) type key.
+# Not all occurrences are normalized with normalize_type_id().
+# The unit tests show where some hex digits are capitalized.
+# We would need to use typing.NewType for actual type-checking.
+CvdumpTypeKey = str
```
jonschz, line 16: just asking: What happens if you do `class CvdumpTypeKey(str): pass`? It's probably the same as `typing.NewType`, just to be sure

reccmp/isledecomp/cvdump/types.py
```diff
@@ -1,12 +1,21 @@
...
 import logging
-from typing import Any, NamedTuple
+from typing import NamedTuple
+from typing_extensions import NotRequired, TypedDict
 
 
 logger = logging.getLogger(__name__)
 
 
+# TODO: For now, this is here just to document where we
```
jonschz, line 11: How about opening a follow-up issue if you don't want to do that here?

reccmp/isledecomp/cvdump/types.py
```diff
@@ -265,26 +333,19 @@ def _get_field_list(self, type_obj: dict[str, Any]) -> list[FieldListItem]:
...
         members: list[FieldListItem] = []
 
-        super_ids = field_obj.get("super", [])
-        for super_id in super_ids:
-            # May need to resolve forward ref.
-            superclass = self.get(super_id)
-            if superclass.members is not None:
-                members += superclass.members
+        if "super" in field_obj:
+            for super_id, _ in field_obj["super"].items():
```
jonschz: isn't there something like `field_obj["super"].keys()` if you only want to iterate over the keys, not the values?

reccmp/isledecomp/cvdump/types.py
```diff
@@ -442,84 +503,79 @@ def read_all(self, section: str):
...
-            # Add the leaf to our dictionary and add details specific to the leaf type.
-            self.mode = leaf_type
-            self.keys[leaf_id] = {"type": leaf_type}
-
-            this_key = self.keys[leaf_id]
-
             try:
-                if self.mode == "LF_MODIFIER":
-                    this_key.update(self.read_modifier(leaf))
+                if leaf_type == "LF_MODIFIER":
```
jonschz: Drive-by: Can this be refactored to `match-case` now that we have a minimum of Python 3.10? Would that have any benefits?

tests/test_cvdump_types.py
```diff
@@ -916,10 +917,24 @@ def test_enum_with_whitespace_and_comma(
...
+    Make sure we parse it correctly."""
+    empty_parser.read_all(
+        """\
+0x657a : Length = 26, Leaf = 0x1009 LF_MFUNCTION
+    Return type = T_VOID(0003), Class type = 0x15ED, This type = 0x15EE, 
+    Call type = ThisCall, Func attr = none
+    Parms = 3, Arg list type = 0x6579, This adjust = 24"""
+    )
+
+    assert empty_parser.keys["0x657a"]["this_adjust"] == 0x24
```
jonschz, line 940: nice!

jonschz (approved):
Very nice, this is a significant improvement! I also found no regressions in the Ghidra import.

reccmp/isledecomp/cvdump/types.py
disinvite: Yes, I see similar output in `mypy`. I like `class` better than `NewType`, so I'll use that for #211.

reccmp/isledecomp/cvdump/types.py
disinvite: I did it that way because I recalled `pylint` suggesting `.items()` for some kind of iteration, but it looks like `.keys()` is fine to use.

PR #206, opened 2025-09-18
Add --nolib option

LRFLEW:
This adds an option to `reccmp` to exclude functions with the `LIBRARY` annotation. The idea being that excluding library functions, particularly functions from libc, is useful for getting a more accurate accuracy percentage and limits the report to functions that are expected to be reimplemented.

### DIFF
--- a/reccmp/isledecomp/compare/core.py
+++ b/reccmp/isledecomp/compare/core.py
@@ -337,7 +337,10 @@ class Compare:
 
             for fun in codebase.iter_name_functions():
                 batch.set_orig(
-                    fun.offset, type=EntityType.FUNCTION, stub=fun.should_skip()
+                    fun.offset,
+                    type=EntityType.FUNCTION,
+                    stub=fun.should_skip(),
+                    library=fun.is_library(),
                 )
 
                 if fun.name.startswith("?") or fun.name_is_symbol:
@@ -886,6 +889,7 @@ class Compare:
                 udiff=udiff,
                 ratio=diff_result.match_ratio,
                 is_effective_match=diff_result.is_effective_match,
+                is_library=match.get("library", False),
             )
 
         if match.entity_type == EntityType.VTABLE:
--- a/reccmp/isledecomp/parser/node.py
+++ b/reccmp/isledecomp/parser/node.py
@@ -21,6 +21,10 @@ class ParserSymbol:
         """The default is to compare any symbols we have"""
         return False
 
+    def is_library(self) -> bool:
+        """The default is to assume that arbitrary symbols are not library functions"""
+        return False
+
     def is_nameref(self) -> bool:
         """All symbols default to name lookup"""
         return True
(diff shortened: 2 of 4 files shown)


reccmp/isledecomp/parser/node.py
```diff
@@ -21,6 +21,10 @@ def should_skip(self) -> bool:
         """The default is to compare any symbols we have"""
         return False
 
+    def is_library(self) -> bool:
+        """The default is to assume any symbols are not library functions"""
```
jonschz:
```suggestion
        """The default is to assume that arbitrary symbols are not library functions"""
```
nitpick

reccmp/isledecomp/parser/node.py
```diff
@@ -21,6 +21,10 @@ def should_skip(self) -> bool:
         """The default is to compare any symbols we have"""
         return False
 
+    def is_library(self) -> bool:
```
jonschz, line 24: Maybe `is_library_function`? `// LIBRARY` annotations are for functions only

reccmp/isledecomp/compare/core.py
```diff
@@ -337,7 +337,10 @@ def orig_bin_checker(addr: int) -> bool:
 
             for fun in codebase.iter_name_functions():
                 batch.set_orig(
-                    fun.offset, type=EntityType.FUNCTION, stub=fun.should_skip()
+                    fun.offset,
+                    type=EntityType.FUNCTION,
+                    stub=fun.should_skip(),
+                    library=fun.is_library(),
```
jonschz, line 343: I'd also go for `is_library_function` here

jonschz (commented):
Can confirm it works as expected. I only left a few small stylistic remarks. Thanks for the contribution!

reccmp/isledecomp/parser/node.py
LRFLEW: True, but the corollary is that since `LIBRARY` annotations can only apply to functions, adding `function` to the name is redundant. I largely stuck with `is_library` because [that's how the equivalent function is named in `parser/marker.py`](https://github.com/isledecomp/reccmp/blob/42ae00aa410182848ccdde198f8e9d53cd9f6e6e/reccmp/isledecomp/parser/marker.py#L123-L124). If you do want me to make that change, I can, but I'm not sure it's a good idea.

reccmp/isledecomp/compare/core.py
LRFLEW: Same as my previous comment, but with the added point that adding `is_` to the beginning might also be inconsistent. For example, the line above it uses `stub`, not `is_stub`. It just seems more consistent to me to use `library` or even `library_function` here to match.

reccmp/isledecomp/parser/node.py
LRFLEW: That does sound better. Changed it as an amended commit to keep the commit history cleaner.

reccmp/isledecomp/parser/node.py
jonschz: alright, it isn't hard to refactor at a later point anyway

reccmp/isledecomp/parser/node.py
jonschz: No need to do that in the future, we squash on merge anyway

PR #207, opened 2025-09-18
Add Local Variable Support

LRFLEW:
This is a followup to #204, though this one is significantly less important. Basically, this adds support for using `// GLOBAL:` annotations for variables declared `static` (i.e. have local linkage) not defined within a function.

This does have an edge-case where it still fails (for me) if I put the static variable inside a namespace, since cvdump indicates that the PDB drops the namespace in that case, but the source file parsing in reccmp doesn't account for that. It's possible that a fix can be added later, but since it will require parsing more context (i.e. checking if it's also in a class), I figured that was not an edge case worth covering for now.


LRFLEW:
I added a case to  `test_cvdump_symbols.py` to test for local functions using the inciting example from my project. I couldn't find a test for global variables at all, so I added a new test file to test both `S_GDATA32` and `S_LDATA32` records.

jonschz:
Sorry for coming back with another request. I think it'd be more future proof if we at least keep the information whether the entity is a `GDATA` or an `LDATA`. Not sure if we need to separate them completely at some point, I wouldn't go there yet. Maybe rename every `gdata` to `lgdata`, add a `global: bool` to `LgdataEntry`, and parse the relevant information from the regex. You could also improve the test by asserting on the `global` property.

LRFLEW:
Looking at it, I think adding a `global` value is reasonable. While it wouldn't have any practical use at this time, it may be useful in the future, and doing it now will reduce effort later (like how I had to get the property of a function being a library function all the way from the parser to the comparer for #206). I can add that at the same time as I fix the name of the regex (since it's no longer just looking for "gdata32")

As for the test for #204, the `test_local_proc()` test in this PR should be sufficient to cover that PR.

tests/test_cvdump_symbols.py
jonschz, line 1: This test passes on commit b09cc3b64e53dcc5a5237b3a97c7be1cca1a1233, so it does not test the new behaviour

jonschz:
This will be good to merge after fixing the last remaining issue. Thanks for the contribution!

tests/test_cvdump_symbols.py
LRFLEW: Oh, whoops, I realized my mistake now. That was testing `CvdumpSymbolsParser` (which is probably good to keep), but my change was to `CvdumpAnalysis`. I'll see about writing a new test case for that.

tests/test_cvdump_analysis.py
```diff
@@ -0,0 +1,47 @@
...
+        section=1,
+        offset=0x12EC,
+        size=0x56,
+        func_type="0x100D",
+        name="check_managed_app",
+    ),
+]
+
+
+def test_cvdump_analysis_functions():
```
jonschz, line 30: This test indeed fails on the previous `master`. However, it does not hit the branch in
```
                if sym.type == "S_LPROC32" and node_dict[key].symbol_entry is not None:
                    continue
```
i.e. the test still passes when I remove these two lines. Could you also add a case for that? Thanks!

tests/test_cvdump_analysis.py
```diff
@@ -0,0 +1,47 @@
...
+    parser = CvdumpParser()
+    parser.symbols_parser.symbols = TEST_SYMBOLS
+
+    def test_node(node: CvdumpNode, sym: SymbolsEntry):
```
jonschz: Regarding your stylistic question, here is my suggestion:
- Refactor `CvdumpNode` to a `dataclass` (which I consider to be a good refactor anyway)
  - Add `from dataclasses import dataclass`, then `@dataclass \n class CvdumpNode:`, and remove the current constructor `__init__`
- Add the expected outcome like so:
```py
expected = [
    CvdumpNode(
        node_type=EntityType.FUNCTION,
        section=1,
        offset=0xCC3,
        confirmed_size=0xA9,
        symbol_entry=TEST_SYMBOLS[0],
        friendly_name="__setargv",
        estimated_size=1577
    ),
    CvdumpNode(
        node_type=EntityType.FUNCTION,
        section=1,
        offset=0x12EC,
        confirmed_size=0x56,
        symbol_entry=TEST_SYMBOLS[1],
        friendly_name="check_managed_app"
    )
]
```
- Then you can just do `assert analysis.nodes == expected` and you can drop the remaining test code.

LRFLEW: Changing `CvdumpNode` to a more straightforward data type makes sense. From what I can tell, it currently *doesn't* do that specifically because it uses the line `node_dict[key] = CvdumpNode(*key)` to set the `section` and `offset` values, while discarding the rest. Is there a preferred way to replace that if I make it a dataclass?

jonschz: While not strictly necessary, if you'd like to go for the extra mile, it would be somewhat cleaner to do add a dedicated constructor, like
```py
    @classmethod
    def from_node_key(cls, key: NodeKey):
        return cls(section=key.section, offset=key.offset)
```

jonschz: I implemented the change myself now

PR #208, opened 2025-09-18
Fix Matching String Containing Null Characters

### LINKED ISSUE #205: Can't match string literals that contain null characters
I'm working with a game that was compiled with MSVC 7.0. Looking at the source code for libc.lib (which is included when you install Visual Studio .NET 2002), we can see that the definition of `__crtLCMapStringA` looks like this (simplified and truncated for readability):

```c
int __cdecl __crtLCMapStringA(....)
{
        static int f_use = 0;

        if (0 == f_use) {
            if (0 != LCMapStringW(0, LCMAP_LOWERCASE, L"\0", 1, NULL, 0))
                f_use = USE_W;
            else if (GetLastError() == ERROR_CALL_NOT_IMPLEMENTED)
                f_use = USE_A;
        }

        ...
}
```

To get this function to match with reccmp, we need to use an annotation to get it to recognize the third argument to `LCMapStringW` as a string, since it doesn't pass the automatic string recognition.

```diff
---
+++
xor esi, esi
inc esi
push esi
-push 0.0 (FLOAT)
+push L"\x00" (STRING)
push 0x100
push ebx
call dword ptr [->KERNEL32.dll::LCMapStringW (IMPORT)]
```

(note, it appears that the same 4-byte all-zero value was used both as a float constant and as a string literal when the original game was linked, which might cause other issues later, but shouldn't impact this issue directly)

To annotate the referenced address as a string, we can use a `// STRING:` annotation in part of the code that doesn't get compiled like such:

```cpp
#if 0

// STRING: TARGET 0xaaaaaaaa
static const wchar_t *str = L"\0";

#endif
```

However, this results in another problem. When we run reccmp with this annotation, it gives the following error:

```
[ERROR] Data at 0x65e384 does not match string '\x00'
```

This is because, when reccmp tries to read the string from the original binary, it stops at the first null character, assuming that's the end of the string. This results in it reading a value of `L""`, rather than `L"\0"`. This issue should be resolvable by changing the string validation to take into account the length of the string from the annotation, rather than just assuming that the first null character is the null terminator.

EDIT: Also to note, `__crtGetStringTypeA` has the same issue as well.

LRFLEW:
Fixes #205 

The issue was that the strings were read from the original binary up until the first null character, which meant it could never match a string containing a null character. This replaces that read with a fixed-length read to validate the whole length of the expected string, plus the final null terminator.


jonschz:
Could you also add a test here?

LRFLEW:
I looked into added a unit test for this, and I'm not sure it's feasible to add one. The affected code is so deep into `reccmp.isledecomp.compare.core.Compare` that I'm pretty much the only way to test it would be to load up an entire project and test it. I looked into adding a test file to `tests/samples/`, but since those don't get compiled and tested, it would only test the parser, which is not the issue here.

Most likely, the best way to test it would be to use a new binfile, along with an associated pdb (probably need to create one specifically for it), and run the `Compare` init using a mock `report` implementation, so we can check for whether an error is reported on the string. However, that would require a fairly major code change to `Compare` to allow for that mocked report implementation, and should probably be done as part of creating a larger test of the class (which I don't know if I'm the best person to make that test, given I'm not as familiar with the project's inner workings).

As is, I'm not sure this change is involved enough to *require* a test case for it. I agree that one would be good to have, but given the challenges with adding one, I'm not sure it should be a blocker for this PR.

jonschz:
I'll take a look into reproducing the issue locally and thinking about how this can be tested.

jonschz:
My suggestion would be to modularise the string matching code so it can be unit tested. Two options, pick what you prefer:
- Move the string matching code into a public function of `class Compare`, like
  ```py
  def match_string(self, string: ParserString) -> tuple[bool, int]:
    # Not that we don't trust you, but we're checking the string
    try:
        if string.is_widechar:
            string_size = 2 * len(string.name) + 2
            raw = self.orig_bin.read(string.offset, string_size)
            orig = raw.decode("utf-16-le")
        else:
            string_size = len(string.name) + 1
            raw = self.orig_bin.read(string.offset, string_size)
            orig = raw.decode("latin1")

        string_correct = orig[-1] == "\0" and string.name == orig[:-1]

        return string_correct, string_size

    except InvalidStringError:
        logger.warning(
            "Could not read string from orig 0x%x, wide=%s",
            string.offset,
            string.is_widechar,
        )
        return False, 0

    except UnicodeDecodeError:
        return False, 0
  ```
  In order to test this, `self.orig_bin.read()` could be mocked.
- Move the string matching into a separate file (`core.py` is quite large) and pass the read function as a callback, like
  ```py
  def match_string(string: ParserString, read_orig: Callable[[int, int], bytes]) -> tuple[bool, int]:
    # Not that we don't trust you, but we're checking the string
    try:
        if string.is_widechar:
            string_size = 2 * len(string.name) + 2
            raw = read_orig(string.offset, string_size)
            orig = raw.decode("utf-16-le")
        else:
            string_size = len(string.name) + 1
            raw = read_orig(string.offset, string_size)
            orig = raw.decode("latin1")

        string_correct = orig[-1] == "\0" and string.name == orig[:-1]

        return string_correct, string_size

    except InvalidStringError:
        logger.warning(
            "Could not read string from orig 0x%x, wide=%s",
            string.offset,
            string.is_widechar,
        )
        return False, 0

    except UnicodeDecodeError:
        return False, 0
  ```

reccmp/isledecomp/compare/core.py
```diff
@@ -365,22 +365,23 @@ def orig_bin_checker(addr: int) -> bool:
...
                 except InvalidStringError:
                     logger.warning(
                         "Could not read string from orig 0x%x, wide=%s",
                         string.offset,
                         string.is_widechar,
                     )
+                    string_correct = False
 
                 except UnicodeDecodeError:
                     string_correct = False
```
jonschz, line 395: Side note: Now that I see it, could you maybe add a warning log to the `UnicodeDecodeError`? Thanks!

LRFLEW: I believe the following line ends up writing a warning via the report anyways, but I can add a specific one about the unicode decoding error as well.

disinvite:
I did some manual testing and this works as advertised. We don't have the modularization to make testing feasible so we can add this to the list of backfill tests in #218 and come back later.

PR #209, opened 2025-09-18
More helpful feedback for missing files.

NogNoa:


reccmp/project/detect.py
```diff
@@ -467,17 +467,20 @@ def detect_project(
...
+                        build_data.targets.setdefault(
+                            target_id, BuildFileTarget(path=p, pdb=pdb)
+                        )
+                        logger.info("Found %s -> %s", target_id, p)
+                        logger.info("Found %s -> %s", target_id, pdb)
+                        break
+                    missing = pdb.name
             else:
-                logger.warning("Could not find %s", filename)
+                logger.warning("Could not find %s", missing)
```
jonschz: This is less about your change and more about how the code was before: The logic flow and the `missing` variable are a bit confusing in my opinion. Have you considered moving the entirety of the inner `for` loop into a local function and leaving that with `return` instead of `break`? That way, the slightly obscure `for-else` can also be refactored out, and you can simply log `"Could not find %s", filename` at the end of the local function.

Furthermore, I think it's appropriate to log a warning when hitting line 482 under any circumstances. That way, you can also get rid of the `missing` variable.

reccmp/project/detect.py
```diff
@@ -175,7 +175,7 @@ def from_directory(cls, directory: Path) -> "RecCmpBuiltProject":
         )
         if not build_directory:
             raise RecCmpProjectNotFoundException(
-                f"Cannot find {RECCMP_BUILD_CONFIG} under {build_directory}"
+                f"Cannot find {RECCMP_BUILD_CONFIG} in {directory} or any of its parents"
```
jonschz: good catch!

PR #210, opened 2025-09-20
Use the Largest Type for Globals

LRFLEW:
I had encountered a weird issue where references to `stdout` weren't being matched by reccmp in my project that's using `/ML` (i.e. statically linking the runtime). Looking into it, MSVC defines `stdout` is a macro relative to the global variable `_iob`. Looking at the output of cvdump, I noticed that there are two entries for this variable in the GLOBALS section, one right after the other, but with different type IDs.

```
S_GDATA32: [0003:00000050], Type:             0x10C7, _iob
S_GDATA32: [0003:00000050], Type:             0x10BF, _iob
```

Looking at the TYPES section, we can see the types are defined as such:

```
0x10c7 : Length = 14, Leaf = 0x1503 LF_ARRAY
	Element type = 0x10A3
	Index type = T_SHORT(0011)
	length = 640
	Name = 
```
```
0x10bf : Length = 14, Leaf = 0x1503 LF_ARRAY
	Element type = 0x10A3
	Index type = T_SHORT(0011)
	length = 0
	Name = 
```

I believe what's happening here is that the first global is based on the variable's definition in libc.lib, whereas the second global is based on the variable's declaration in stdio.h, which looks like this:

```c
_CRTIMP extern FILE _iob[];
```

Since the declaration is unsized, it seems that the debug information for the obj files that only see the declaration defaulted to a length of zero. This causes it to differ from the obj file where it's actually defined, and when linking all the obj files together, the pdb ends up keeping both types for it, hence the two global entries.

Currently, when reccmp encounters a second global at the same location, it overrides the old size with the new size. Since the declaration's type comes after the definition's one, this results in the global being recorded as having a size of zero. While this particular case could be fixed by simply keeping the first type instead, I chose to make it more robust, and have it keep whichever type is larger. That way, when reccmp encounters multiple globals at the same location, it will always keep whichever size covers the most cases, regardless of the order they appear in.


jonschz:
Same as the other PRs - I'd be great if you could add a test. I think this one shouldn't be too difficult - it may be sufficient run `CvdumpAnalysis` and insert a mock `CvdumpParser` with pre-set fields, most of them empty arrays.

jonschz:
An integration test is also fine. What I'm interested in is any reasonable test that fails on the previous code and passes on the new code.

disinvite:
> Looking at this again, would it be better to simply ignore zero-length arrays, and throw a warning / error if it sees two incompatible globals entries (eg. different names or different non-zero sizes)?

I think this is the way to go with the information we currently have.

In the specific example of the zero length array, it would be better to use the type key with a non-zero size.

If there are two competing types that each have a non-zero size, log a warning so we can review the cvdump output later. Using the largest size (the way this PR is written now) seems like a reasonable choice to me. In the absence of real-world data around this situation, I think we could make the case for any option.

reccmp/cvdump/analysis.py
```diff
@@ -121,16 +121,23 @@ def __init__(self, parser: CvdumpParser):
             if key not in node_dict:
                 node_dict[key] = CvdumpNode(*key)
 
-            node_dict[key].node_type = EntityType.DATA
-            node_dict[key].friendly_name = glo.name
```
disinvite, line 193: The new tests added by #237 fail if we don't have type information. That's because we no longer establish the `DATA` node here and we hit the upcoming `except` block before it gets created.

We _should_ always have type information from the PDB, but it's not currently a requirement to create a global variable entity.

disinvite:
Finished this one up and fixed a test that did not properly exercise the change in #301.

PR #213, opened 2025-09-23
Add Discord link

disinvite:


README.md
```diff
@@ -1,5 +1,7 @@
 # Reccmp Decompilation Toolchain
 
+[![Discord server](https://badgen.net/badge/icon/discord?icon=discord&label)](https://discord.gg/25bNWfZB)
```
madebr, line 3: Should we also add a badge for the Matrix channel ([Matrix channel](https://matrix.to/#/#isledecomp:matrix.org))? There are always devs who have an adverse reaction to closed platforms.

disinvite: Sure. Should we keep using the Lego Island Decomp channel for that purpose? cc @foxtacles

foxtacles: Yes that should be fine

PR #214, opened 2025-09-28
Simplify DB batch functions

disinvite:
This removes the `insert_orig` and `insert_recomp` API calls from the DB batch object. These were there to allow you to abandon an update if there is already an entity at the given address.

It turns out to be only 0-1% slower to just check the address each time and call `set_orig` or `set_recomp`. I made the DB API calls `orig_used` and `recomp_used` public for this purpose. I think this makes the code read better because the distinction between `insert_orig` and `set_orig` is not 100% clear. I removed the tests that used the `insert` functions.

I also changed how matched pairs of addresses are stored. We use a `list` instead of two `dict`s and replace reused addresses with the value from the most recent call to `match()`. Not a significant difference in performance, but I want to make it possible to store more information about a match. (namely: how was the match created, and was it unique?)

### DIFF
--- a/reccmp/isledecomp/compare/db.py
+++ b/reccmp/isledecomp/compare/db.py
@@ -208,23 +208,23 @@ class EntityBatch:
         self._recomp_addr[orig] = recomp
 
     def _finalized_matches(self) -> Iterator[tuple[int, int]]:
-        """If an orig or recomp address is used in two or more matches, use
-        only the most recent match and vacate all others."""
+        """Reduce the list of matches so that each orig and recomp addr appears once.
+        If an address is repeated, retain the first pair where it is used and ignore any others.
+        """
         used_orig = set()
         used_recomp = set()
-        final = []
 
-        # Walk the list in reverse and make sure both addrs are unique.
         # This should have the same effect as the original implementation
         # that used two dicts to check uniqueness during each call to match().
-        for orig, recomp in reversed(self._matches):
+        for orig, recomp in self._matches:
             if orig not in used_orig and recomp not in used_recomp:
                 used_orig.add(orig)
                 used_recomp.add(recomp)
-                final.append((orig, recomp))
-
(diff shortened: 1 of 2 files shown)


reccmp/isledecomp/compare/db.py
```diff
@@ -173,81 +173,70 @@ def matched_entity_factory(_, row: object) -> ReccmpMatch:
...
-        self._orig_to_recomp[orig] = recomp
-        self._recomp_to_orig[recomp] = orig
+        self._matches.append((orig, recomp))
 
     def set_recomp_addr(self, orig: int, recomp: int):
         self._recomp_addr[orig] = recomp
 
+    def _finalized_matches(self) -> Iterator[tuple[int, int]]:
+        """If an orig or recomp address is used in two or more matches, use
+        only the most recent match and vacate all others."""
```
jonschz: what is the rationale for preferring the most recent match over an older one? I'd add that reason to the doc comment.

disinvite: No particular reason. If either address is reused in the same batch, we can either break the previous match or keep it and throw out the new one.

A possible argument for doing it the other way here: once the match is saved in the database, we will not break it if you try to match (for example) the same orig addr to a different recomp addr.

I doubt either situation (breaking a committed or uncommitted match) comes up at all. The queries that derive the list of addresses to match should all check if the address is _already_ matched.

jonschz: In that case I'd definitely log a warning if we get here. Also, if nothing depends on the reversing logic, I'd probably just iterate forwards to keep the code simple. We can even drop the `final` array and simply `yield` instead of `append`

disinvite: Yes, good idea. It does not make a difference and doing it this way is a little easier to explain.

PR #215, opened 2025-09-28
Add support for `pyghidra`

jonschz:
At long last, I have finally been able to make the Ghidra import run from [pyghidra](https://pypi.org/project/pyghidra/). I applied some larger refactoring as well, though I tried to keep this PR on focus. Here are some of the relevant changes:
- I split the extensive `import_functions_and_types_from_pdb.py` into a setup module for the Ghidrathon script and a module for the import logic that was still in there
- I removed the functionality of running the Ghidrathon script outside of a Ghidra context. Previously, I needed it to step through the Ghidra import in a debugger, but that can be done in the pyghidra import now
- I restructured the directories. Note that this has lead to a change in the Ghidrathon script directory, so if you have a current Ghidrathon setup, the script directory needs to be changed.

Notes to the reviewer:
- pyghidra requires Ghidra 11.3 or newer, so the Ghidra version we use in isledecomp (11.1) cannot be used to validate this setup.
- I recommend you start with the updated `/reccmp/ghidra/README.md`, since I hope it will be self-explanatory.
- I have added a basic docker compose setup for a local Ghidra server, which is necessary in order to validate the remote repository functionality without having side effects on production repositories.

This should finally unblock https://github.com/isledecomp/isle/issues/1008, although it isn't really needed there anymore. I'm still planning to add a CI operation to isle as an example for other projects.

**Tasks**
- [x] Write a follow-up for Ghidra integration tests -> #219 
- [x] Add info on the dev Ghidra server to the README
- [x] Consider looking at https://pypi.org/project/ghidra-stubs/
  - If it works: Update README, remove custom Ghidra headers if possible, check if `mypy` is possible
  - -> very nice! All of the Ghidra code is now typechecked, I even found a few minor errors
- [x] Update README + doc strings: The Ghidrathon script now matches based on hashes


reccmp/ghidra/development/server.conf
jonschz, line 1: Note to self: This may no longer be needed. I had it set up as an intermediate step before authentication worked

reccmp/ghidra/importer/importer.py
```diff
@@ -0,0 +1,201 @@
...
+def do_with_error_handling(step_name: str, action: Callable[[], None]):
+    try:
+        action()
+        GLOBALS.statistics.successes += 1
+    except Lego1Exception as e:
+        log_and_track_failure(step_name, e)
+    except RuntimeError as e:
+        cause = e.args[0]
+        if CancelledException is not None and isinstance(cause, CancelledException):
+            # TODO: This looks wrong - should abort
```
jonschz: Address TODO

.gitignore
```diff
@@ -5,7 +5,7 @@ build
 *.pyo
 __pycache__/
 dist
-.venv/
+/.venv*
```
jonschz, line 8: drive-by - I have needed multiple venvs for a while

README.md
```diff
@@ -3,27 +3,31 @@
 [![Discord server](https://badgen.net/badge/icon/discord?icon=discord&label)](https://discord.gg/25bNWfZB)
 [![Matrix channel](https://badgen.net/badge/icon/matrix?icon=matrix&label)](https://matrix.to/#/#isledecomp:matrix.org)
 
-`reccmp` (recompilation comparison) is a collection of tools for decompilation projects. It was born from the [decompilation of LEGO Island](https://github.com/isledecomp/isle). Functions and data are matched based on comments in the source code. For example:
+`reccmp` (recompilation compare) is a collection of tools for decompilation projects. It was born from the [decompilation of LEGO Island](https://github.com/isledecomp/isle). Functions and data are matched based on comments in the source code. For example:
```
jonschz, line 6: I think we usually say `recompilation compare`?

reccmp/ghidra/ghidra_import.py
jonschz, line 1: contains significant code from `/reccmp/ghidra_scripts/import_functions_and_types_from_pdb.py`, but GitHub does not recognise the move

reccmp/ghidra/scripts/import_functions_and_types_from_pdb.py
jonschz, line 1: contains significant code from `/reccmp/ghidra_scripts/import_functions_and_types_from_pdb.py`, but GitHub does not recognise the move

reccmp/ghidra/importer/function_importer.py
```diff
@@ -97,7 +97,7 @@ class ThunkPdbFunctionImport(PdbFunctionImporter):
     Only the name of the function will be imported."""
 
     def matches_ghidra_function(self, ghidra_function: Function) -> bool:
-        name_match = self.name == ghidra_function.getName(False)
+        name_match = self.name == ghidra_function.getName()
```
jonschz, line 100: Both versions exist as overloads in `FunctionDB`, they don't make difference here

reccmp/ghidra/importer/function_importer.py
```diff
@@ -286,7 +286,7 @@ def overwrite_ghidra_function(self, ghidra_function: Function):
                 Function.FunctionUpdateType.DYNAMIC_STORAGE_ALL_PARAMS,  # this implicitly sets custom variable storage to False
                 True,
                 SourceType.USER_DEFINED,
-                [
+                *[
```
jonschz, line 289: I am surprised that this worked before, but either version appears to be fine and this one is correctly typed

reccmp/ghidra/README.md
```diff
@@ -0,0 +1,65 @@
...
+- Click the _Add_ (Plus icon) button.
+- You may have to select _File Chooser Options -> Show '.' Files_.
+  - Case 1: Select `Lib/site-packages/reccmp/ghidra/scripts` inside your _project's_ virtual environment.
+  - Case 2: Select `<this repository's root>/reccmp/ghidra/scripts` (don't select anything within this repository's venv).
+- Close the window and click the _Refresh_ button.
+- This script should now be available under the folder _reccmp_.
+
+### File names
+
+Since it is not possible to configure any script parameters in Ghidra, the correct decompilation target needs to be inferred. Currently, this is done via the file name of the original binary when you first imported the file into Ghidra. This needs to match the file name configured in `reccmp-project.yml` under `targets/<target>/filename`.
```
jonschz: This paragraph is obsolete now that we have hash matching

reccmp/ghidra/README.md
```diff
@@ -0,0 +1,65 @@
...
+
+## Development
+
+- Note that when running the scripts via Ghidrathon, imported modules persist across multiple runs of the script (see also[this issue](https://github.com/mandiant/Ghidrathon/issues/103)).
+    If you indend to modify an imported library, you have to use `import importlib; importlib.reload(${library})` or restart Ghidra for your changes to have any effect. Unfortunately, even that is not perfectly reliable, so you may still have to restart Ghidra for some `reccmp` code changes to be applied.
```
jonschz, line 66: The paragraph on the previous system for type stubs has been removed since there now is a much better solution with the `ghidra-types` package

reccmp/ghidra/importer/importer.py
jonschz: This was obsolete and didn't do anything. I fixed being able to cancel the script.

reccmp/ghidra/development/server.conf
disinvite: It looks like it's working after removing the volume mount, so the default server.conf may be fine.

README.md
disinvite: I don't know if we ever _say_ it 😀. The original file `reccmp.py` had its name before I started making changes to it. The origin might be unix `cmp` rather than the words "comparison" or "compare". But I like "compare" better.

reccmp/ghidra/cli.py
```diff
@@ -0,0 +1,101 @@
...
+    parser.add_argument(
+        "--local-project-dir",
+        metavar="<dir>",
+        help="Defaults to Ghidra's default project directory if omitted",
+    )
+
+    # Optional arguments for remote projects
+    parser.add_argument(
+        "--remote-checkin-comment",
+        metavar="<mesage>",
```
disinvite: typo: message

reccmp/ghidra/importer/exceptions.py
disinvite, line 1: Not part of this change, but it may be time to rename `Lego1Exception`

reccmp/ghidra/cli.py
```diff
@@ -0,0 +1,101 @@
...
+        self,
+        parser: argparse.ArgumentParser,
+        namespace: argparse.Namespace,
+        values: Sequence[str] | None,
+        option_string=None,
+    ):
+        assert isinstance(values, str)
+        # Ghidra's own URL handler is not a good fit because it does not support username and password,
+        # so we'd have to extract those and re-assemble the URL with those removed.
+        parsed_url = urllib.parse.urlparse(values)
```
disinvite, line 45: Do you think it's worth discouraging users from entering the credentials at the command line? (where they might be visible in a CI .yml file, for example)

Granted this restriction would complicate running the script manually, and it's not as though other tools don't allow you to shoot yourself in the foot this way. Just a thought.

reccmp/ghidra/ghidra_import.py
```diff
@@ -0,0 +1,230 @@
...
+
+TEMP_PROJECT_NAME = "temp-reccmp-import"
+# This name is irrelevant, but we have to pick something
+TRANSACTION_NAME = "pyghidra-reccmp-import"
+
+logger = logging.getLogger(__file__)
+
+
+def main():
+    args = parse_reccmp_import_args()
```
disinvite, line 28: We should make sure that the file you selected actually belongs to the reccmp target. I tested locally with `--target BETA10` and `--file ISLE.EXE` and got a slew of errors. (No address space overlap between the two binaries, so no changes were possible)

reccmp/ghidra/cli.py
jonschz: I thought about that too. In a CI run, I'd make the whole URL a secret. What do you think about pulling the credentials from env vars like `RECCMP_GHIDRA_PASSWORD` if they are not provided in the URL?

reccmp/ghidra/ghidra_import.py
jonschz: I'll try to make something work with file hashes. That would also fix one major restriction of the Ghidrathon script.

Edit: This was easier than I first thought. Feel free to check it out.

reccmp/ghidra/development/server.conf
jonschz: Yeah, I originally needed it to get access to anonymous mode, but that is no longer necessary or helpful.

reccmp/ghidra/importer/globals_importer.py
```diff
@@ -49,10 +50,16 @@ def import_global_into_ghidra(
         data_end = glob.orig_addr + data_type.getLength()
 
         while True:
-            # Clear conflicting data (usually auto-generated by Ghidra)
-            next_data_entry = api.getDataAfter(address_ghidra)
+            # Clear conflicting data (usually auto-generated by Ghidra).
+            # Note that we have to downcast the `ghidra.program.model.listing.Data` to a
```
jonschz:
```suggestion
            # Clear conflicting data (usually auto-generated by Ghidra).
```

reccmp/ghidra/ghidra_import.py
disinvite: Nice!

reccmp/ghidra/cli.py
disinvite: Maybe we just need `RECCMP_GHIDRA_URL` as the environment variable. We should log a warning if you use the command line parameter, but it should still work. (My thinking is that we should alert to a problem even when users do env var substitution in the CI action, despite the fact that the values aren't readily visible)

disinvite: But feel free to do it in a follow-up because it's not urgent

disinvite (approved):
Forgot to actually click the approve button yesterday. This is great! Excluding the (now fixed) problem yesterday, I had no trouble using the script with local and shared projects and via Ghidrathon. And thank you for the docker-compose file. It made testing the remote option a lot simpler.

reccmp/ghidra/scripts/import_functions_and_types_from_pdb.py
```diff
@@ -0,0 +1,228 @@
...
+logging.root.setLevel(LOG_LEVEL)
+
+
+def find_target(api: "FlatProgramAPI") -> "RecCmpTarget":
+    """
+    Tries to find a `reccmp` project on a parent path of this script file.
+    This can be achieved by installing `reccmp` into a venv inside a decompilation project
+    and having `reccmp-build.yaml` next to the venv directory.
+
+    Inside that `reccmp` project, a target is chosen based on the name of the executable of the current file.
```
jonschz: Update this part

reccmp/ghidra/cli.py
jonschz: I'd prefer reading the password (and optionally the username) from an env var because currently, the presence of the URL tells the script that the target is a remote repo. I think the CLI will be less clear if neither the local name nor the URL parameter are required.

reccmp/ghidra/scripts/import_functions_and_types_from_pdb.py
```diff
@@ -0,0 +1,228 @@
...
+    Note that the relevant filename is that of the original binary (Ghidra keeps track of that),
+    not what the file is called in the Ghidra project.
+
```
jonschz:
```suggestion
    Inside that `reccmp` project, a target is chosen based on the hash of the executable of the current file.

```

reccmp/ghidra/importer/entity_names.py
```diff
@@ -71,7 +71,7 @@ def sanitize_name(name: str) -> SanitizedEntityName:
     new_name = "::".join(new_name_split)
     if new_name != name:
         logger.info(
-            "Changed class or function name from '%s' to '%s' to avoid Ghidra issues",
+            "Changed entity name from '%s' to '%s' to avoid Ghidra issues",
```
jonschz, line 74: small drive-by

Issue #220, opened 2025-10-05
Break up `core.py`


disinvite:
`core.py` contains the `Compare` class that drives everything in the `reccmp` scripts. It's been hovering around 1k lines for a while now and it's time to break it up into functions grouped by context:
- Populating data from various data sources (code, pdb, binary files)
- Updating the entity db based on data we already have (i.e. the analysis step)
- Reporting issues found during analysis (e.g. we know this vtable has different size in orig and recomp)
- Output of compared entities

These functions should each have parameters for the various aspects of Compare that will be referenced or updated. This allows us to write tests and mock things better. (Everything being run through instance methods is the blocker for #218.) 

The Compare class has (or will have):
- Entity DB
- Type database (Until we have #106, we can continue to use the existing `CvdumpTypesParser`)
- Lines DB. Associates code lines with an address. Translates Windows and Unix paths.
- Function DB. Primarily used by the `stackcmp` script for gathering information about registers and local variables used in each function.

I started hacking this together and I think we have a path forward. You can see that [here](https://github.com/disinvite/reccmp/tree/work-breakup). I'll put in some individual PRs that should hopefully not clash with anything in flight.

Tasks I've identifier so far are:

- [x] Allow Lines DB to be updated in any order. (Meaning: the search should work regardless of whether we add the list of local code paths _or_ the association between PDB-path-and-line to address first) #221 
- [x] Store datatype key alongside variable entities. `datacmp` runs its own second parse of the cvdump to get this information. #222 
- [ ] `CvdumpAnalysis` will become our de-facto way to access data from the PDB. (So it can be mocked for tests)
- [ ] `CvdumpAnalysis` should use addresses rather than section/offset pairs. We need to drag in the associated `PEImage` (or a `Callable`) that can do the conversion. We also need to ignore invalid addresses.
- [ ] `CvdumpAnalysis` nodes should be a `dict` rather than a `list`. We end up building this structure where it is needed anyway. To maintain address order, we can insert into the `dict` after sorting the list.
- [x] Split `_load_cvdump()` into component functions that each update one aspect of the compare core. #231 
- [ ] Split `_load_markers()` into two steps: match using Lines DB and creating entities from the annotations
- [x] Split up operations that do the same thing on each binary. For example: `_find_strings`, `_find_float_const`, `_create_thunks`. This requires a new function in `EntityBatch` that will accept a `bool` or `Enum` parameter to decide which binary to update. #239

PR #221, opened 2025-10-05
LinesDb: Process local and foreign paths in any order

disinvite:
Part of #220. The lines database reads two types of data:
- Local paths: the list of code files that contain annotations, local to the file system where you run reccmp
- Foreign paths: as read from the PDB, so this is always a Windows path even if you are using Wine. For each path, read the non-unique[^1] mapping of line numbers to addresses

The problem is that we expect this specific order of operations:
1. Set up the list of local paths first. This happens in the `Compare` constructor when we pass the list of file paths into the `LinesDb` constructor.
2. For each foreign path, send in the list of line-number-to-address tuples.
3. When reading code annotations, search the lines database for the recomp address to match the orig address.

We would like to make it possible to reverse steps 1 and 2. We now read the list of files at the point where we establish `DecompCodebase`.

This simplifies the `LinesDb` constructor and creates a new `add_files()` method to set up the local paths. My solution to the problem of calling the methods out of order is to save up the two types of data in a queue, then process everything when we need to search.

I restricted the types of `LinesDb` to work with `Path` or `PurePath` only instead of allowing `str` and converting it.

I removed `orig_bin_checker` because we can just pass in the `is_valid_addr` function directly.

No new diffs in `LEGO1` or `BETA10`.

[^1]: Non-unique because the same line number can appear multiple times and point to different addresses. For example: templates, and the start and end of a loop.

### DIFF
--- a/reccmp/isledecomp/compare/lines.py
+++ b/reccmp/isledecomp/compare/lines.py
@@ -37,13 +37,11 @@ class LinesDb:
             self._filenames.setdefault(path.name.lower(), []).append(path)
 
         # Save any failed matches to be processed when we get a new local path.
-        retry = []
-
-        for foreign_path, lines in self._line_queue:
-            if not self._match_foreign_path_to_local(foreign_path, lines):
-                retry.append((foreign_path, lines))
-
-        self._line_queue = retry
+        self._line_queue = [
+            (foreign_path, lines)
+            for foreign_path, lines in self._line_queue
+            if not self._match_foreign_path_to_local(foreign_path, lines)
+        ]
 
     def add_line(self, foreign_path: PureWindowsPath, line_no: int, addr: int):
         """Connect the foreign path to a line number and address pair."""
--- a/reccmp/isledecomp/compare/lines.py
+++ b/reccmp/isledecomp/compare/lines.py
@@ -12,15 +12,15 @@ logger = logging.getLogger(__name__)
 
 
 class LinesDb:
-    # pylint: disable=too-many-instance-attributes
     def __init__(self) -> None:
-        self._new_data = False
         self._path_resolver = cache(convert_foreign_path)
 
-        self._path_queue: list[Path | PurePath] = []
+        # List of foreign paths and associated (line_no, address) pairs that cannot be
+        # matched with the current set of local file paths.
         self._line_queue: list[tuple[PureWindowsPath, list[tuple[int, int]]]] = []
 
-        # Set up memoized map of filenames to their paths
+        # Map of local filenames to their paths. This is used to reduce the list of
+        # possible local paths that can be matched to a foreign path.
         self._filenames: dict[str, list[PurePath]] = {}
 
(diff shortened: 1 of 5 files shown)


reccmp/isledecomp/compare/lines.py
```diff
@@ -35,33 +32,59 @@ def __init__(
...
 
     def add_lines(
-        self, foreign_path: PureWindowsPath, lines: Sequence[tuple[int, int]]
+        self, foreign_path: PureWindowsPath, lines: Iterable[tuple[int, int]]
     ):
         """
         Connect the remote path to a line number and address pair.
         """
-        filename = foreign_path.name.lower()
+        self._new_data = True
```
jonschz: I'm not sure I understand the purpose of this `_new_data` field. Why don't we run `process` immediately after adding lines instead of keeping some dirty flag in this field?

reccmp/isledecomp/parser/codebase.py
```diff
@@ -13,13 +14,13 @@
...
-    def __init__(self, filenames: Iterable[str], module: str) -> None:
+    def __init__(self, paths: Iterable[Path], module: str) -> None:
         self._symbols: list[ParserSymbol] = []
 
         parser = DecompParser()
-        for filename in filenames:
-            parser.reset_and_set_filename(filename)
-            with open(filename, "r", encoding="utf-8") as f:
+        for path in paths:
+            parser.reset_and_set_filename(str(path))
```
jonschz, line 22: This looks like `parser.reset_and_set_filename()` should take an argument of type `Path`, or am I missing something?

jonschz (approved):
Looks good! Wondering if the `_new_data` can be refactored away in an efficient way, but if not, this is good to merge

reccmp/isledecomp/compare/lines.py
disinvite: Thanks! I simplified the code and it's much better now.

reccmp/isledecomp/parser/codebase.py
disinvite: Yes, but we'd have to go down the rabbit hole of making changes in `marker.py` and `linter.py` and then test everything. So we should do that in a followup: #224

reccmp/isledecomp/compare/lines.py
```diff
@@ -35,33 +32,57 @@ def __init__(
...
+    def add_local_paths(self, paths: Iterable[Path] | Iterable[PurePath]):
+        for path in paths:
+            self._filenames.setdefault(path.name.lower(), []).append(path)
+
+        # Save any failed matches to be processed when we get a new local path.
+        retry = []
+
+        for foreign_path, lines in self._line_queue:
+            if not self._match_foreign_path_to_local(foreign_path, lines):
+                retry.append((foreign_path, lines))
```
jonschz: I think it's possible to use a generator expression here, like `[(foreign,lines) for (foreign, lines) in _line_queue if not self.match(...)]`

PR #222, opened 2025-10-05
Create type database as member of `Compare`

disinvite:
Part of #220. Outside of `analysis.py`, the consumers of type information are:
- Ghidra scripts
- `Compare._match_array_elements()` that creates entities for array datatypes
- The `datacmp` script

In each case, we need to access the `Cvdump` parser directly because the type key is not stored alongside the variable entities. `datacmp` re-parses the PDB because we didn't make types accessible at all early on.

This unifies all three consumers to access `Compare.types`. We establish the data in `_load_cvdump` by copying it into a second `CvdumpTypesParser`. We use the `data_type` attribute in the entity to store the key for the types database so we can grab it when iterating on matched variables.

The next step is #106. The `.get()` method in `CvdumpTypesParser` returns `TypeInfo` objects, an attempt to convert the raw data from the PDB into a more general format. However, as demonstrated in `_match_array_elements`, the data we need is a little scattered, so we also need to access the raw data in `.keys` directly.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,6 +1,6 @@
 # Reccmp Decompilation Toolchain
 
-[![Discord server](https://badgen.net/badge/icon/discord?icon=discord&label)](https://discord.gg/25bNWfZB)
+[![Discord server](https://badgen.net/badge/icon/discord?icon=discord&label)](https://discord.gg/aSKCSXwpNp)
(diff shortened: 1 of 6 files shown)


jonschz:
The change in interface looks very good. I haven't looked too closely at the logic changes — my unterstanding is that they are temporary anyway. If `asmcmp` and `datacmp` produce no regressions on LEGO1 and BETA10, feel free to merge.

PR #227, opened 2025-10-10
Remove `self` from core functions

disinvite:
Part of #220. The component functions that load metadata or alter the database now get their dependencies as args instead of relying on `self`. After this, we can break off `core.py` into (at least) two other files. I think this final step should be its own PR because it will require some `git mv` trickery to keep the file history.

Other changes here:
- If you invoke `reccmp-reccmp` and provide the four paths (for orig and recomp binaries, PDB, code directory) instead of a target name, we still need to derive the target name for parsing the code annotations. We assume it is the orig bin base filename, and there is code in `core.py` to alert the user that we have made this assumption. However, the target name would not be null because we have already set the value in `project/detect.py`, so I moved the alert code there.
- I split up `_load_cvdump` into four functions: `_match_entry`, `_load_cvdump_types`, `_load_cvdump` and `_load_cvdump_lines` because these are all independent operations. We could do the same with `_load_markers` but it requires some changes to `DecompCodebase`, so we'll save it for later.
- I moved `function_comparator` up to the `__init__` method instead of creating it after matching.
- The matching functions are now confined to `run()` and `__init__` just sets the class members instead of doing everything all at once.
- Similarly, `cvdump_analysis` is now the arg type for `__init__` instead of just the path to the PDB. We will need some kind of abstract/interface class for mocking this in tests. We can figure that out in the following steps. (The same is true for `DecompCodebase`)


reccmp/project/detect.py
```diff
@@ -388,8 +388,14 @@ def __call__(
...
         original, recompiled, pdb, source_root = list(Path(o) for o in values)
+
+        # Assumes base filename of the original binary is the module name.
+        target_id = original.stem.upper()
+        # This happens before argparse_parse_logging() is called, so it will not match our format.
+        logger.warning('Assuming target name is "%s"', target_id)
+
         target = RecCmpTarget(
-            target_id=original.stem.upper(),
+            target_id=target_id,
```
jonschz, line 398: Potentially controversal opinion, and a bit of a tangent: Do you think we should maintain the 4 file pattern for much longer? In my opinion, it's something we can deprecate (though I'd leave it in for a couple months in case anyone uses it). If we instead commit to that feature, I think we should implement the ability to provide the target name via (optional?) argument.

disinvite: I was thinking about the module option recently too, but from the `reccmp-project.yml` direction. I just put in #230.

On the paths option, I agree. We will need to change it to support projects with more or fewer component files.

PR #228, opened 2025-10-10
Update included binary files

disinvite:
`LINK.EXE` and `MSPDB41.DLL` are no longer bundled with `reccmp`. These were both dependencies of `DUMPBIN.EXE` that I should have removed in #118. Neither are needed to run cvdump.

I added the MIT license for redistributing `cvdump.exe` as provided in [Microsoft's repo](https://github.com/microsoft/microsoft-pdb).

Updated the `MANIFEST.in` file to reflect the changes.


reccmp/bin/LICENSE.cvdump.txt
```diff
@@ -0,0 +1,21 @@
+Copyright (c) 2015 Microsoft Corporation. All rights reserved.
```
madebr, line 1: Perhaps rename this to `LICENSE.cvdump.txt`?
That's what [we are doing in SDL_image](https://github.com/libsdl-org/SDL_image/tree/main/VisualC/external/optional/x64) for optional libraries.

Issue #98, opened 2025-03-25
Better format for asm diffs


disinvite:
The format for diffs in the JSON and HTML reports could use some improvement. It looks like this:

```json
// Exploded to exaggerate the problem
"diff": [
    [
        "@@ -0x401560,22 +0x401560,22 @@",
        [
            {
                "orig": [],
                "recomp": [
                    [
                        "0x401560",
                        "mov edx, dword ptr [esp + 4]"
                    ]
                ]
            },
            {
                "both": [
                    [
                        "0x401560",
                        "push ebx",
                        "0x401564"
                    ],
                    [
                        "0x401561",
                        "push esi",
                        "0x401565"
                    ]
                ]
            },
// You get the idea...
```

You may ask: Why is it like this?

When comparing texts `A` and `B`, the `get_opcodes()` method from `difflib.SequenceMatcher` gives a list of operations to convert `A` to produce `B`. Those are: `equal`, `insert`, `replace`, and `delete`, each with a range of affected lines from `A` and `B`. To limit the size of the diff, `get_grouped_opcodes()` compares the distance between each opcode to a certain number of context lines. If the opcodes are close enough then they are included in the same group. Each group starts with a "diff hunk", which is the line with the double at-signs that shows the range of the given group. Long runs of matching lines are removed from the final result.

`difflib` does all this for you, but we had to modify the stock code a bit so that diffs on asm instructions also include the address in both orig and recomp. I think I put it in this format to make it easier to display the data in both the terminal and in HTML.

The problems are:
- Lots of JSON overhead. Slow to parse. Especially when there are 1024 reports in a run of `reccmp-aggregate`.
- The full asm listing is lost in this conversion. `reccmp-stackcmp` should be able to read the entire function.

Partly blocked by #47 because changing this format also requires non-trivial changes to the javascript code for the HTML view.

PR #238, opened 2025-10-15
Refactor HTML components

### LINKED ISSUE #201: HTML - Detail row doesn't open for 100% match
#193 changed the JSON output so that entities with no diff no longer insert `diff: []` into the object. As a result, opening the detail row for a 100% match displays this error in the HTML view:

`Uncaught TypeError: can't access property "length", obj.diff is undefined`

This wasn't caught because the test data we're using didn't get updated.

disinvite:
This is a major overhaul of code for the HTML report:

- Splits `reccmp.js` into component files. We can't use `<script type="module">` with the `file:///` protocol, so we omit the imports and exports by checking for the `// reccmp-pack-begin` and `// reccmp-pack-end` markers. The files need to be combined in a certian order so that dependencies can be satisfied.

    - Despite relying on a particular order when building the HTML report, the component files are all valid JS modules. In a follow-up we can add a CI action that will run the e2e tests against the minified version.

- Each component gets its data by registering with the main `<reccmp-provider>` component using HTML events. We had been doing something similar, but less elegantly. This also drops the global variable `appState`.

- Components like `<listing-options>` that combined multiple inputs are dropped in favor of small components like: `<prev-page-button>`, `<page-select>`, `<show-recomp>`.

- Components like `<sort-indicator>` are dropped because the functionality just doesn't warrant a component.

- Fixes #201.

Next steps:

- The look and feel could use a refresh (along with the CSS). For this PR, though, I wanted to keep things exactly the same. Hopefully this is true in all browsers.

- There are no e2e tests for the diff display components, so I left them alone for now. Once we have those and can make changes, it will unblock #98.

- With the components broken out like this, we could add unit tests.

### DIFF
--- a/reccmp/assets/template.html
+++ b/reccmp/assets/template.html
@@ -268,7 +268,6 @@
       window.global_reccmp_report = {{{report}}};
     </script>
     <script>{{{reccmp_js}}}</script>
-    </script>
   </head>
   <body>
     <div class="main">


reccmp/isledecomp/utils.py
jonschz, line 1: This looks fine for now. Having a proper web packer would be nice, but I think the downsides (another ecosystem to manage) would overweigh significantly at the moment.

reccmp/isledecomp/utils.py
```diff
@@ -10,17 +11,71 @@
...
+    This only captures lines between "reccmp-pack-begin" and "reccmp-pack-end"
+    and adds a header with the source filename."""
+    js_path = get_asset_file(filename)
+    lines = []
+
+    with open(js_path, "r", encoding="utf-8") as f:
+        lines = list(reccmp_pack_generator(f))
+
+    horiz_line = f"/{'*' * 78}/\n"
+    return horiz_line + f"// {filename}\n" + "".join(lines)
```
jonschz:
```suggestion
    file_header = f"/{'*' * 78}/\n// {filename}\n"
    return file_header + "".join(lines)
```
maybe?

disinvite:
> I noticed one change in behaviour: The search bar is now reset on hitting F5. Previously, the search bar contents would survive hitting F5, but it wouldn't be applied until you changed it (i.e. it was in a glitched state). Not sure what's desired here, tbh.

I would call it a bug because you had to type in the box to make the search run again. The same was true for the other controls when they got stuck. We could do it properly in a follow-up PR and have the results update after the refresh.

jonschz (approved):
I haven't looked at it too thoroughly, but it looks like a move in the right direction overall. It's kind of difficult to see the changes, but we're covered by Playwright to a significant degree, which is very good.

The use of `innerHtml` is okay with me as long as we don't to string interpolation there, which can be tempting in the future. It would be great if we had access to `jsx` / `tsx` files, but that's not possible without Node.js dependencies. So fine with me.

jonschz:
> > I noticed one change in behaviour: The search bar is now reset on hitting F5. Previously, the search bar contents would survive hitting F5, but it wouldn't be applied until you changed it (i.e. it was in a glitched state). Not sure what's desired here, tbh.
> 
> I would call it a bug because you had to type in the box to make the search run again. The same was true for the other controls when they got stuck. We could do it properly in a follow-up PR and have the results update after the refresh.

Not a priority in my opinion. Feel free to merge without creating a follow-up ticket

reccmp/isledecomp/utils.py
disinvite: Yep, agreed. But if we did ever want to do it, the hard work is done.

PR #239, opened 2025-10-18
Split some core analyze functions

disinvite:
The functions in the new `analyze.py` create entities from both of the binary image files. These can be refactored to work with one image at a time, then we just run them twice in `core.py`. I picked the three easiest functions to split: auto-detecting floats, auto-detecting strings, and creating intermediate build thunk functions.
- Creates a generic `db.used()` function on the `EntityDb` to test if an entity exists in the given address space. A new enum picks which one: `ImageId.ORIG` and `ImageId.RECOMP`.
- Creates a generic `batch.set()` function on the `EntityBatch` that works with the `ImageId` enum.
- The `create_thunks` function sets the referenced address (#181) using the `ref_orig` and `ref_recomp` attributes. To make this work with `ImageId`, we now use just `ref` in the `batch.set()` caller then pick the right one inside the function.
- Added tests for the new db and batch functions.
- Added tests for the three analyze functions. Unlike #237, I used a mock `PEImage` because it was simple.

### DIFF
--- a/reccmp/isledecomp/compare/db.py
+++ b/reccmp/isledecomp/compare/db.py
@@ -201,21 +201,21 @@ class EntityBatch:
     def set_recomp(self, addr: int, **kwargs):
         self._recomp.setdefault(addr, {}).update(kwargs)
 
-    def set(self, img: ImageId, addr: int, **kwargs):
+    def set(self, img: ImageId, addr: int, *, ref: int | None = None, **kwargs):
         if img == ImageId.ORIG:
-            if "ref" in kwargs:
-                kwargs["ref_orig"] = kwargs["ref"]
-                del kwargs["ref"]
+            if ref is not None:
+                kwargs["ref_orig"] = ref
 
             self.set_orig(addr, **kwargs)
-            return
 
-        if img == ImageId.RECOMP:
-            if "ref" in kwargs:
-                kwargs["ref_recomp"] = kwargs["ref"]
-                del kwargs["ref"]
+        elif img == ImageId.RECOMP:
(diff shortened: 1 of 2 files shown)


reccmp/isledecomp/compare/db.py
```diff
@@ -201,6 +201,22 @@ def set_orig(self, addr: int, **kwargs):
     def set_recomp(self, addr: int, **kwargs):
         self._recomp.setdefault(addr, {}).update(kwargs)
 
+    def set(self, img: ImageId, addr: int, **kwargs):
```
jonschz: Could you make the `ref` argument explicit as an optional argument? Like
```
    def set(self, img: ImageId, addr: int, ref: int | None, **kwargs):
```
Then you can skip the `del` I think

reccmp/isledecomp/compare/db.py
```diff
@@ -441,6 +457,15 @@ def recomp_used(self, addr: int) -> bool:
...
         return cur.fetchone() is not None
 
+    def used(self, img: ImageId, addr: int) -> bool:
+        if img == ImageId.ORIG:
+            return self.orig_used(addr)
+
+        if img == ImageId.RECOMP:
+            return self.recomp_used(addr)
+
+        return False
```
jonschz: If it's necessary to write anything here (it should be unreachable from a type perspective, but maybe it's still required), I'd `assert False` here

jonschz (approved):
Nice! Didn't take a very detailed look at the tests, do you want a specific opinion on something?

PR #243, opened 2025-10-21
Tests for `match_array_elements`

disinvite:
Style question: do we want tests like this to be scoped using a class or should they all be in their own file?


jonschz:
Personally, I prefer the flat style that most of our existing tests have. I'd also prefer to be consistent across the repository. That being said, any tests are better than no tests, and my opinion on the matter is not particularly strong.

PR #245, opened 2025-10-22
Tests for `name_thunks`

disinvite:
To streamline the new tests, I added a `get()` function to retrieve entities from the DB with the `ImageId` enum as a parameter. This is similar to the functions added in #239. We can then parametrize the tests to demonstrate that they work for both orig and recomp thunks.

I added the type for the `db` fixture to the new tests from #239. I disabled `mypy` for the tests where we intentionally use the wrong type.

In the new tests: I intentionally did not check whether the name is `Thunk of ___`. I think we should consider giving thunks their own entity type and just copying the name of the thunked function. In the asm, the change would be:
```diff
- Thunk of 'PizzaMissionState::ClassName' (FUNCTION)
+ PizzaMissionState::ClassName (THUNK)
```
I did not do that here because the Ghidra scripts check for "Thunk of" specifically, so we have to work that part out first.

### DIFF
--- a/reccmp/isledecomp/formats/pe.py
+++ b/reccmp/isledecomp/formats/pe.py
@@ -410,35 +410,8 @@ class PESection:
         """Get the highest possible offset of this section"""
         return max(self.size_of_raw_data, self.virtual_size)
 
-    def match_name(self, name: str) -> bool:
-        return self.name == name
-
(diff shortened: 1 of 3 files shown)


jonschz:
A dedicated thunk type sounds good!

Issue #235, opened 2025-10-12
Analysis is limited if there is no relocation table


disinvite:
Many of our analysis steps depend on the [relocation table](https://learn.microsoft.com/en-us/windows/win32/debug/pe-format#the-reloc-section-image-only) to decide which values are real addresses. What if the image is not relocatable?

GTA2 is one such case. To solve this we need to "create" our own relocation table. The table is just a list of addresses where the next 4 bytes (depending on pointer size) should be altered when the image is loaded by Windows. We also store the addresses being modified because we know these each point at the start[^1] of an entity.

The "easiest" way is to parse the assembly and grab the value of each pointer. If the address is an immediate value (e.g. `cmp`) then we can make an educated guess using the imagebase and the size of whatever code section.

I say "easiest" because walking the assembly is complicated by jump tables (from switch statements) and jumps directly to a function that occur in SEH unwinds. This is true even if you know the function size ahead of time, and you probably wouldn't unless this has been specifically annotated. (Not possible until #169)

- [ ] Move relocation table to a generic "pointers" structure. This should probably go in the entity database so future queries can make use of this data. After #231 is merged, we can populate this table in `analyze.py`. Any code that references the relocation table should check there instead.
- [ ] The hard part: walk each known function and report the address of each pointer. This may be easier to write if we use `iced_x86` (See #175) but otherwise we can just look at the strings returned by `capstone`.

[^1]: This is true _most_ of the time. We also have indirect calls where the pointer is an offset of a vtable, string instructions that use a pointer with an offset into the string, and loops over an array of structs where the compared pointer value is an irrelevant offset or variable.

Issue #185, opened 2025-08-10
Should not assume symbols with at-sign are functions


disinvite:
MSVC 6 (GTA2) assigns some symbols for float constants. e.g.:
```asm
fmul dword ptr [__real@4@3ff18000000000000000 (FUNCTION)]
```

When reading the PDB we assume that symbol is a function because of the at-sign in the symbol. This is generally correct, but maybe just because most things in most images are functions.

We can at least check whether the symbol appears in a section that has the execute flag set.

disinvite:
> would special casing symbols that start with `__real@` be a valid fix for this?

Yes. And that is what we’ll end up doing, but there are a few other dominoes that need to be in place to actually _match_ a float like this. (The biggest obstacle is that you can’t annotate a float in the original binary right now.)

Since we now incorrectly assume the `__real` symbol is a function, you might be able to work around the missing feature in reccmp with an annotation like this:

```
// STUB: 105 0x12345678 SYMBOL
// __real@___rest_of_the_symbol
```

Of course this isn’t actually a function, but that might create a match and get this diff off of your list for the time being.

PR #250, opened 2025-10-25
Create float entities for `__real` symbols

### LINKED ISSUE #27: Manage const data differently

disinvite:
MSVC versions newer than 4.2 assign a symbol for some (all?) floating point constants. For example: `__real@4@3ff18000000000000000`

The `@4` or `@8` tells whether it is a single or double precision float.

Our search for floating point instructions is dependent on having a relocation table. (See #235.) If your image is not relocatable then we do not create any float entities and you see these symbols in the asm output. (They appear to be functions based on a wrong assumption about what entities have symbols with the at-sign: #185)

This changes the Cvdump analysis step to assign the `FLOAT` entity type for the `__real` symbols. There is a new function to reads from the binary and set the name for any float entities that do not have one. (Until we resolve #27, the data for floats and strings goes in the `name` attribute.)

We can repeat this step for the orig address space when it is possible to annotate floats in code or via CSV.

Also: in `queries.py`, changed `Iterable` to `Iterator` where the latter was intended.

### DIFF
--- a/README.md
+++ b/README.md
@@ -15,6 +15,8 @@ MxCore* MxObjectFactory::Create(const char* p_name)
 
 This allows you to automatically verify the accuracy of re-compiled functions, virtual tables, variable offsets and more. [Click to see the full syntax](docs/annotations.md).
 
+You can supplement the code annotations with metadata from CSV files. The instructions and syntax are [here](docs/csv.md).
+
 At the moment, C++ compiled to 32-bit x86 with old versions of MSVC (like 4.20) is supported. Work on support for newer MSVC versions is in progress - testing and bug reports are greatly appreciated. Other compilers, languages and architectures are not supported at the moment, but feel free to contribute if you wish to do so!
 
 ## Getting started
--- a/docs/csv.md
+++ b/docs/csv.md
@@ -1,0 +1,131 @@
+# CSV import
+
+`reccmp` has the option to parse [comma-separated value (CSV)](https://en.wikipedia.org/wiki/Comma-separated_values) files and load data to supplement code annotations on the original binary.
+
+Any metadata set from a CSV file will overwrite the same data from code annotations on the same address.
+
+If the same address is repeated on multiple lines the same CSV file, or across multiple files, the metadata will be overwritten each time.
+
+## Motivation
+
+The CSV feature is optional. In most situations, we can gather enough information to run analysis and provide a diff for each function using only the code annotations and PDB.
+
+However, loading data via CSV allows the user to specify all aspects of entity metadata, including attributes like the entity size, which is not currently possible with code annotations.
+
+You may find it helpful to copy data from Ghidra and use it to bootstrap a new decomp project. For example: CRT or library functions.
+
+## Setup
+
+Add the list of CSV files under the `data-sources` key for the target in `reccmp-project.yml`. For example:
+
+```yml
+targets:
+  ISLE:
+    filename: ISLE.EXE
+    source-root: .
+    hash:
+      sha256: 5cf57c284973fce9d14f5677a2e4435fd989c5e938970764d00c8932ed5128ca
+    data-sources:
+      - test.csv
+```
+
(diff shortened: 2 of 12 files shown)


disinvite:
I have `create_partial_floats` in the `analyze` module now. It started off in `mutate` last night when I entered the PR. The lines between these modules aren't black and white but the idea is:
- `mutate` updates the DB without using other dependencies
- `analyze` reads from the binary without using other dependencies

So `analyze` fits _better_, but it's not perfect. We are reading in specific spots as directed by the DB entities. We could split to another module if this pattern of using multiple data sources develops further.

reccmp/isledecomp/compare/analyze.py
```diff
@@ -199,3 +209,27 @@ def match_vtordisp(db: EntityDb, orig_bin: PEImage, recomp_bin: PEImage):
                 batch.set_recomp(recomp_vtor.addr, name=new_name)
 
             batch.match(vtor.addr, recomp_vtor.addr)
+
+
+def create_partial_floats(db: EntityDb, image_id: ImageId, binfile: PEImage):
+    """For each float entity without any data,
+    read the value the binary and set the entity name."""
+    if image_id not in (ImageId.ORIG, ImageId.RECOMP):
+        assert False, "Invalid image id"
```
jonschz:
```suggestion
    assert image_id in (ImageId.ORIG, ImageId.RECOMP), "Invalid image id"
```

jonschz:
Do you have a suggestion how to try this out on a real example? I saw that the SmartHeap symbols in ISLE are also annotated this way.

disinvite:
> Do you have a suggestion how to try this out on a real example? I saw that the SmartHeap symbols in ISLE are also annotated this way.

The SH functions that have the `@` should be unchanged. Meaning: they are still flagged as functions by the PDB parsing step.

As for real world testing... I built this with MSVC 6:
```c
float myfloat = 0.0;

// FUNCTION: MYFLOAT 0x401010
int main() {
    myfloat *= 0.5;
    return 0;
}
```

and these commands:
```
cl /Zi .\myfloat.c /link /FIXED
reccmp-reccmp --paths .\myfloat.exe .\myfloat.exe myfloat.pdb . -v 0x401010
```

and got this on master:
```
0x401010 : push ebp     (myfloat.c:4)
0x401011 : mov ebp, esp
0x401013 : -fld dword ptr [<OFFSET1>]
0x401019 : -fmul qword ptr [<OFFSET2>]
0x40101f : -fstp dword ptr [<OFFSET1>]
         : +fld dword ptr [myfloat (DATA)]      (myfloat.c:5)
         : +fmul qword ptr [__real@8@3ffe8000000000000000 (FUNCTION)]
         : +fstp dword ptr [myfloat (DATA)]
0x401025 : xor eax, eax         (myfloat.c:6)
0x401027 : pop ebp      (myfloat.c:7)
0x401028 : ret
```

If you run with this PR branch, you get this instead:
```
0x40101f : -fstp dword ptr [<OFFSET1>]
         : +fld dword ptr [myfloat (DATA)]      (myfloat.c:5)
         : +fmul qword ptr [0.5 (FLOAT)]
         : +fstp dword ptr [myfloat (DATA)]
0x401025 : xor eax, eax         (myfloat.c:6)
```

`/FIXED` is there so we don't autodetect the float on the orig size by checking the instructions. If you remove that, your image now has a relocation table, and you'll see `0.5 (FLOAT)` for both.

PR #251, opened 2025-10-26
Tests for `load_markers`

disinvite:
I thought it would be better to have these tests down before doing any refactoring.

If we read all the code files at once into `DecompCodebase` then we can "replay" the annotations across multiple functions. It also simplifies mocking because you just need `PurePath` and the file contents as a string, the same format I used in these tests.

The only thing not covered is catching an `InvalidStringError`. I'm not sure it's actually possible to get this exception thrown.

### DIFF
--- a/tests/test_compare_ingest_code.py
+++ b/tests/test_compare_ingest_code.py
@@ -1,16 +1,14 @@
 """Tests for creating/matching entities using code annotations."""
 
-from textwrap import dedent
 from pathlib import PurePath, Path, PureWindowsPath
+from textwrap import dedent
+from typing import Iterable
 import pytest
-from reccmp.isledecomp.types import EntityType
+from reccmp.isledecomp.types import EntityType, TextFile
 from reccmp.isledecomp.formats import PEImage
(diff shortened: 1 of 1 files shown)


tests/test_compare_ingest_code.py
```diff
@@ -0,0 +1,653 @@
...
+    load_markers(tmp_path, lines_db, binfile, "TEST", db)
+
+    # No exception raised
+    assert db.get_by_orig(0x11001000) is None
+
+
+def test_load_code_duplicate_addr(
+    tmp_path: Path, db: EntityDb, lines_db: LinesDb, binfile: PEImage
+):
+    """Each address can only be used once.
```
madebr, line 59: What about linking with `/OPT:COMDAT`? Then functions with identical implementations get folded into one.

disinvite: Do you want to open an issue for that? For now, you would have to annotate just one of the functions and leave the rest unmarked. Same as if they were `inline` functions.

PR #256, opened 2025-10-30
Exclude vtordisp functions from unique name check

disinvite:
Vtordisp functions should have a suffix that displays the two displacement values. For example: ``Act3Brickster::Animate`vtordisp{-4, 0}'``. The PDB does not include this value, so the name we read is just `Act3Brickster::Animate`.

Our check for overloaded functions looks at function entities where a name is used more than once. We exclude thunks but not vtordisp functions. This means that some functions get assigned the disambiguation suffix (the parameter list) where this isn't strictly necessary.

This isn't a huge problem, of course, but I want to clear this out now so it doesn't muddy the waters of future changes to thunk and vtordisp naming.

The `roadmap` diff[^1] for `BETA10` looks like this:

```diff
-0001:00000780   0001:00005519      19865  fun  5       Thunk of 'Act3Brickster::Animate(float)'
+0001:00000780   0001:00005519      19865  fun  5       Thunk of 'Act3Brickster::Animate'
 0001:000030a2                             fun  5       
-0001:000187d7   0001:000801e3     424460  fun  2891    Act3Brickster::Animate(float)
+0001:000187d7   0001:000801e3     424460  fun  2891    Act3Brickster::Animate
 0001:0001c4c0   0001:00085200     429376  fun  8       Act3Brickster::Animate`vtordisp{-4, 0}'
```

The unnamed function at `0001:000030a2` (aka `0x100040a2`) is a thunk pointing at the vtordisp function. The vtordisp points at the _named_ thunk, which points at the real function.

We now also create entities for vtordisp functions (and the referenced function, if it does not already exist.) Because of the way we identify vtordisps, the size is known, so we set that too:

```diff
 0001:0001c3a0   0001:000856a0     430848  fun  11      Act3Actor::VTable0x90`vtordisp{-4, 16}'
 0001:0001c3b0   0001:000856b0     430848  fun  8       Act3Cop::HitActor`vtordisp{-4, 0}'
 0001:0001c3c0   0001:000856c0     430848  fun  8       Act3Cop::VTable0x9c`vtordisp{-4, 0}'
+0001:0001c3d0                             fun  8       
+0001:0001c480                             fun  11      
+0001:0001c490                             fun  11      
 0001:0001c4a0   0001:000851e0     429376  fun  8       Act3Brickster::ParseAction`vtordisp{-4, 0}'
+0001:0001c4b0                             fun  11      
 0001:0001c4c0   0001:00085200     429376  fun  8       Act3Brickster::Animate`vtordisp{-4, 0}'
+0001:0001c4d0                             fun  11      
+0001:0001c4e0                             fun  11      
 0001:0001c4f0   0001:00085230     429376  fun  8       Act3Brickster::HitActor`vtordisp{-4, 0}'
```

[^1]: I'm cheating here slightly. This is what you _would_ see if `roadmap` displayed the unique name of an entity when there is one. Change line 454 from `match.name` to `match.best_name()` to see that. I'll put in a PR for this.

### DIFF
--- a/reccmp/isledecomp/compare/analyze.py
+++ b/reccmp/isledecomp/compare/analyze.py
@@ -2,7 +2,13 @@
 These functions update the entity database based on analysis of the binary files.
 """
 
+import logging
+import struct
 from reccmp.isledecomp.formats.pe import PEImage
+from reccmp.isledecomp.formats.exceptions import (
+    InvalidVirtualAddressError,
+    InvalidVirtualReadError,
+)
 from reccmp.isledecomp.types import EntityType, ImageId
 from reccmp.isledecomp.analysis import (
     find_float_consts,
@@ -11,6 +17,10 @@ from reccmp.isledecomp.analysis import (
     is_likely_latin1,
 )
 from .db import EntityDb, entity_name_from_string
+from .queries import get_floats_without_data
+
+
+logger = logging.getLogger(__name__)
 
 
 def match_entry(db: EntityDb, orig_bin: PEImage, recomp_bin: PEImage):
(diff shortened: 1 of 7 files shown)


jonschz (approved):
Looks good! One question: Is it intended that the vtordisps have no name in the `roadmap` output? They aren't the only functions without a name, just asking.

PR #262, opened 2025-11-27
Improve thunk matching and naming

disinvite:
- New entity types: `THUNK` and `VTORDISP`. `THUNK` will be used for incremental build (debug) thunks. Import thunks will keep using the `FUNCTION` type. The distinction is that `THUNK` entities acquire the `Thunk of` prefix.

- Created a new `refs` SQL table for storing links between entities. We no longer need `ref_orig` and `ref_recomp` attributes on the main `entities` table. `refs` has a column for the `ImageId` enum to specify which address space we are in.

- The new `set_ref` function on the DB batch creates the link between entities. Ref information now includes the two displacement values so we can match vtordisp functions along with thunks. Thunks have a displacement value of `(0, 0)`.

- The new approach for naming thunks and vtordisps is: start by copying each function name into a new `names` SQL table. Using the data from `refs`, copy the parent's names to each child function. We can run this repeatedly to cover chains of thunks. `BETA10` has many instances of: thunk -> vtordisp -> thunk -> function.

- The database function `get_functions()` returns all matched `FUNCTION`, `THUNK`, or `VTORDISP` entities. The ghidra script will need some changes to handle thunks differently so I thought it best to just keep the same behavior for now.

- In chat we discussed omitting vtordisps from the diff report. I decided not to do it here but it's a one line change when we're ready.

My end goal for this is to have a series of "stage" tables where we collect data points (names, sizes, types) from all available sources, decide which are the most likely to describe the entity (e.g. prefer user data over autodetect), then match and build the final list of entities.

To achieve that, I need to change the behavior of `get_by_orig` and `get_by_recomp`, and not using them while matching thunks is the first step.

### DIFF
--- a/tests/test_compare_db_queries.py
+++ b/tests/test_compare_db_queries.py
@@ -110,7 +110,7 @@ def test_get_referencing_entity_matches(db: EntityDb):
         batch.set_ref(ImageId.RECOMP, 300, ref=100)
 
     # Can match these child entities that point to the same matched parent.
-    assert not list(get_referencing_entity_matches(db))
+    assert list(get_referencing_entity_matches(db)) == [(200, 300)]
 
     # Create the match as directed by the query.
     with db.batch() as batch:


reccmp/isledecomp/compare/match_msvc.py
```diff
@@ -394,15 +395,14 @@ def match_ref(
...
+    Repeats until there are no new matches."""
+    for __ in range(10):
+        new_matches = False
+        with db.batch() as batch:
+            for orig_addr, recomp_addr in get_referencing_entity_matches(db):
+                new_matches = True
+                batch.match(orig_addr, recomp_addr)
+
+            if not new_matches:
+                break
```
jonschz, line 410: I presume that you wanted to avoid an infinite loop like this? a `while True` would probably be fine here. If you want to keep it like this, at the very least I'd throw or log an error in case we get all the way to the end of the loop.

disinvite: Yeah, I put a range on it just to be safe. I don't think we need more than 4 iterations, but you never know. I'll add an error log on there.

reccmp/isledecomp/compare/core.py
```diff
@@ -313,7 +311,7 @@ def _compare_match(self, match: ReccmpMatch) -> DiffReport | None:
                 is_stub=True,
             )
 
-        if match.entity_type == EntityType.FUNCTION:
+        if match.entity_type in (EntityType.FUNCTION, EntityType.VTORDISP):
```
jonschz, line 316: Just curious: What's the reason to not incude thunks here? If there is one, maybe add a comment why

reccmp/isledecomp/compare/db.py
```diff
@@ -10,27 +10,37 @@
...
+        primary key (img, addr)
     );
 
-    CREATE VIEW orig_refs (orig_addr, ref_id, nth) AS
-        SELECT thunk.orig_addr, ref.rowid,
-            Row_number() OVER (partition BY thunk.ref_orig order by thunk.orig_addr) nth
-        FROM entities thunk
-        INNER JOIN entities ref on thunk.ref_orig = ref.orig_addr
-        WHERE json_extract(thunk.kvstore, '$.vtordisp') IS NULL;
+    CREATE VIEW matches (match_id, addr_x, addr_y) AS
```
jonschz: Is there a reason to use the names `addr_x` and `addr_y` instead of `orig_addr` and `recomp_addr`? If there isn't, I'd stick to the established names for consistency.

reccmp/isledecomp/compare/queries.py
```diff
@@ -51,36 +45,38 @@ class ThunkWithName(NamedTuple):
...
+        GROUP BY e.rowid""",
+        (EntityType.THUNK, EntityType.VTORDISP),
     ):
-        assert isinstance(orig_addr, int) or isinstance(recomp_addr, int)
+        assert img in (ImageId.ORIG, ImageId.RECOMP)
+        assert isinstance(addr, int)
         assert isinstance(name, str)
-        yield ThunkWithName(orig_addr, recomp_addr, name)
+
+        if disp0 == 0 and disp1 == 0:
```
jonschz, line 76: While I like the general structure, I don't quite like using `0, 0` as a magic value for `this is a thunk instead of a vtordisp`. I have a slight preference to making `disp0` and `disp1` nullable and having `disp0 = None` encode a thunk, though there are other data models as well (like adding a match type enum).

If you prefer to keep the `0,0` magic value or refactor to the `None` approach, please add a comment to the `CREATE TABLE` entry to explain the semantics. If you go for a self-explanatory enum, a comment isn't necessary.

reccmp/isledecomp/compare/queries.py
```diff
@@ -105,3 +101,42 @@ def get_floats_without_data(
...
+            ON r.img = x.img AND r.addr = x.addr
+        )
+        SELECT x.addr, y.addr FROM
+        (SELECT * FROM linked_refs WHERE img = 0) x
+        INNER JOIN
+        (SELECT * FROM linked_refs WHERE img = 1) y
+        ON  x.disp0 = y.disp0
+        AND x.disp1 = y.disp1
+        AND x.nth = y.nth
+        AND x.match_id = y.match_id
```
jonschz, line 137: 1. Could you put all SQL keywords in uppercase (e.g. `PARTITION BY`, `ORDER BY`)?
2. I spent a few minutes and I think I understood the gist of it, but this statement is still quite complex. I'll partially trust in your experience, the outcome, and the tests. Please let me know in case you want a more detailed review.
3. I haven't thought too much about whether the test coverage by the existing and new tests is sufficient. Do you think this function would benefit from dedicated unit tests?

tests/test_compare_analyze.py
```diff
@@ -142,9 +141,7 @@ def test_create_analysis_vtordisps(db: EntityDb, binfile: PEImage):
     # Using the first vtordisp as an example
     e = db.get_by_orig(0x1000FB50)
     assert e is not None
-    assert e.get("type") == EntityType.FUNCTION
-    assert e.get("ref_orig") == 0x1000FB60
-    assert e.get("vtordisp") is True
+    assert e.get("type") == EntityType.VTORDISP
```
jonschz, line 168: since `batch.set_ref` is called in `create_analysis_vtordisps`, there should be a new assertion here that covers the new ref.

tests/test_match_msvc.py
```diff
@@ -681,19 +752,78 @@ def test_match_ref(db):
...
+
+    # No entities to match: should not report.
+    match_ref(db, report)
+    report.assert_not_called()
+
+    # Run one iteration: should not report.
+    with patch(
+        "reccmp.isledecomp.compare.match_msvc.get_referencing_entity_matches",
+        return_value=iter([(100, 100)]),
+    ) as getter:
```
jonschz, line 817: I think this is good enough. If you want to go all out, you could provide an optional callback argument to `match_ref` which defaults to `get_referencing_entity_matches`. It would be more stable under refactoring, but it's not a necessity in my opinion.

tests/test_match_msvc.py
jonschz, line 1: Nice tests overall!

PR #264, opened 2025-12-02
Read data for partially created string entities

disinvite:
This adds the ability to finish a partially created string entity. Partially created means: we have the address, type, and (optionally) the size, but no data has been read from the binary.

This enables us to create string entities in CSV. This might have been possible already by setting the `name` field directly, but that's a hack. (See #27)

Unlike with `// STRING` code annotations, there is no way in the CSV to indicate that the string is UTF-16LE, so I added another entity type, `WIDECHAR`.

I moved the string reads out of the `load_cvdump` function, so this now only establishes the string/widechar entities and the new `create_partial_strings` function does the rest. We can do the same for `load_markers`, but there is the added stipulation of checking whether the text in the code file matches the data at the given address. I'm not exactly sure yet how to manage this as part of the larger refactor in #220 so the existing behavior stays for now.

Also in `load_cvdump`: we no longer bother demangling the string symbol to get the approximate length. As the comment explains, this has never been consistent and the section contribution value is much better for our purposes. If there is no section contribution, we default to reading data up to the null-terminator, which is probably still better than relying on the string length encoded in the symbol.

We do not alert "Failed to match string" for strings _unless_ they come from code annotations. The reason is that you may want to mark all strings in the original binary in advance, before they exist in recomp code, and avoid the deluge of errors that would otherwise appear.

We were missing tests for the similar `create_partial_floats` function so I added a couple.

One loose end: `entity_name_from_string` adds the `L` prefix to widechar strings (as in `L"Test"`) and this is the only thing preventing an `EntityType.STRING` from being matched to an `EntityType.WIDECHAR`.

### DIFF
--- a/tests/test_compare_analyze.py
+++ b/tests/test_compare_analyze.py
@@ -13,7 +13,6 @@ from reccmp.isledecomp.compare.analyze import (
     create_analysis_strings,
     create_thunks,
     create_analysis_vtordisps,
-    complete_partial_floats,
     complete_partial_strings,
 )
 
@@ -197,36 +196,6 @@ def test_create_analysis_vtordisps_no_overwrite(db: EntityDb, binfile: PEImage):
     e = db.get_by_orig(0x1000FB60)
     assert e is not None
     assert e.get("type") != EntityType.FUNCTION
-
-
-def test_complete_partial_floats(db: EntityDb):
-    """Should read data for a partially-initialized float entity."""
-    binfile = Mock(spec=[])
-    binfile.read = Mock(return_value=b"\x00\x00\x00\x3f")
-
-    with db.batch() as batch:
-        batch.set(ImageId.ORIG, 100, type=EntityType.FLOAT, size=4)
-
-    complete_partial_floats(db, ImageId.ORIG, binfile)
-
-    e = db.get_by_orig(100)
-    assert e is not None
-    assert e.name == "0.5"
-
-
-def test_complete_partial_floats_double(db: EntityDb):
-    """Should read data for a partially-initialized float entity (double precision)."""
-    binfile = Mock(spec=[])
-    binfile.read = Mock(return_value=b"\x18\x2d\x44\x54\xfb\x21\x09\x40")
-
-    with db.batch() as batch:
-        batch.set(ImageId.ORIG, 100, type=EntityType.FLOAT, size=8)
-
-    complete_partial_floats(db, ImageId.ORIG, binfile)
-
-    e = db.get_by_orig(100)
-    assert e is not None
-    assert e.name == "3.141592653589793"
 
 
 def test_complete_partial_strings(db: EntityDb):
--- a/tests/test_compare_analyze_partial_floats.py
+++ b/tests/test_compare_analyze_partial_floats.py
@@ -1,3 +1,4 @@
+from unittest.mock import Mock
 import pytest
 from reccmp.isledecomp.compare.analyze import (
     complete_partial_floats,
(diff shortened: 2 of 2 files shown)


reccmp/isledecomp/compare/analyze.py
```diff
@@ -210,3 +211,58 @@ def create_partial_floats(db: EntityDb, image_id: ImageId, binfile: PEImage):
                     image_id.name.lower(),
                     addr,
                 )
+
+
+def create_partial_strings(db: EntityDb, image_id: ImageId, binfile: PEImage):
+    """For each string/widechar entity without any data,
+    read the value the binary and set the entity name.
+    If the entity has no size, read until we hit a null-terminator."""
```
jonschz, line 219: Just to verify: Conceptually, this is also what we did before, right?

reccmp/isledecomp/compare/ingest.py
```diff
@@ -72,74 +68,28 @@ def load_cvdump(cvdump_analysis: CvdumpAnalysis, db: EntityDb, recomp_bin: PEIma
...
+                # Use the section contribution size if we have it. It is more accurate
+                # than the number embedded in the string symbol:
+                #
+                #     e.g. ??_C@_0BA@EFDM@MxObjectFactory?$AA@
+                #     reported length: 16 (includes null terminator)
+                #     c.f. ??_C@_03DPKJ@enz?$AA@
+                #     reported length: 3 (does NOT include terminator)
+                #
+                # Using a known length enables us to read strings that include null bytes.
+                # string_size is the total memory footprint, including null-terminator.
```
jonschz: Now that the logic isn't here anymore, we should mention that strings without size are read until a terminator is found

reccmp/isledecomp/compare/ingest.py
```diff
@@ -72,74 +68,28 @@ def load_cvdump(cvdump_analysis: CvdumpAnalysis, db: EntityDb, recomp_bin: PEIma
...
+                string_size = None
+                if sym.section_contribution is not None:
+                    string_size = sym.section_contribution
```
jonschz:
```suggestion
                string_size = None if sym.section_contribution is None else sym.section_contribution
```

reccmp/isledecomp/compare/analyze.py
```diff
@@ -210,3 +211,58 @@ def create_partial_floats(db: EntityDb, image_id: ImageId, binfile: PEImage):
                     image_id.name.lower(),
                     addr,
                 )
+
+
+def create_partial_strings(db: EntityDb, image_id: ImageId, binfile: PEImage):
```
jonschz: Let me get this straight: What is the call order of `create_partial_strings` vs. `load_cvdump`? In other words, where is the "partial" string completed to a "full" string? Shouldn't this be called something like `complete_partial_strings`? Would apply to the `float` case as well.

reccmp/isledecomp/cvdump/analysis.py
```diff
@@ -58,11 +58,11 @@ def set_decorated(self, name: str):
             self.node_type = EntityType.DATA
             self.friendly_name = demangle_vtable(self.decorated_name)
 
-        elif self.decorated_name.startswith("??_C@"):
+        elif self.decorated_name.startswith("??_C@_0"):
             self.node_type = EntityType.STRING
-            demangled = demangle_string_const(self.decorated_name)
-            assert demangled is not None
-            self.confirmed_size = demangled.len
+
```
jonschz, line 63: Not your change, but is it fine to not have a final `else` in this clause? Should we log a warning there?

disinvite: I think it's fine without the `else`. All we're doing here is setting additional attributes if the symbol matches some known pattern.

reccmp/isledecomp/compare/analyze.py
disinvite: Yes, thanks. Those were poorly named. `complete_partial_strings` is the correct description for what it's doing.

reccmp/isledecomp/compare/analyze.py
disinvite: Yes

Issue #257, opened 2025-11-02
Improve `NEImage` support


disinvite:
Our priority is still `PEImage`, but I'm curious to see if some small improvements would allow us to get _something_ from a 16-bit NE executable.

Things that are definitely missing:

- [ ] Need to implement `seek()` for `NEImage` so we can read data.
- [ ] Relocations are handled differently with NE. There is no image base, so segments in each far pointer start as zeroes. We need to modify the binary after reading to set some value.
- [ ] On the topic of segments: Ghidra starts at `0x1000` for each segment and increments by 8 (`0x1008`, `0x1010`, `0x1018`...) for each new segment. This is arbitrary and is just mean to keep the address spaces separate. We can do the same thing easily, but we may not want to lock into this pattern if other tools do things differently.
- [ ] We need another copy of `asm/instgen.py` and `asm/parse.py` that will work with 16-bit instructions. These are somewhat well-structured so we can just copy the interface.
- [ ] Asm sanitize will be tricky in the short term. If we (naively) assume that `call` instructions use the current `CS` then we need to provide this to the sanitize function. We could disable sanitize or only do far pointers to start.
- [ ] I haven't been able to create a PDB with MSVC 1.52 that we can read with cvdump. We could parse the .map file instead, but this is another new thing to add. We also need to change the project detection and yaml reading code to allow the PDB file to be blank.
- [ ] Virtually everything in the app is set up for `PEImage` so we need to skip steps that will not be compatible. This would be easier if we had a more fleshed out `Image` interface.

PR #266, opened 2025-12-07
Use `Image` type in compare core

disinvite:
The minimum adjustments required to run the main `reccmp` script with a generic `Image` instead of `PEImage`. Of course, you won't get any results until we make more changes to the image types. I have my sights set on NE support (#257) and already have things hacked together in another branch.

### DIFF
--- a/docs/csv.md
+++ b/docs/csv.md
@@ -55,6 +55,7 @@ The `"name"` field is the name for the entity, as you would expect. `"size"` is 
 - `function, template, synthetic, library, stub` for function entities
 - `global`
 - `string`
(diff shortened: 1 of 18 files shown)


madebr (approved):
`abc.abstractmethod` is an alternative, but it's a mess imho with it awkwardly requiring  a abc metaclass.

disinvite:
> `abc.abstractmethod` is an alternative, but it's a mess imho with it awkwardly requiring a abc metaclass.

I looked into that briefly but had the same reservations about using it. We could always do it later if the need comes up.

PR #277, opened 2025-12-22
Add entities for SEH items

### LINKED ISSUE #274: Mark `funcinfo` structs
`funcinfo` structs are easy to detect because of their magic number:
```
0x19930520: up to VC6,
0x19930521: VC7.x(2002-2003),
0x19930522: VC8 (2005)
```
More info here: https://www.openrce.org/articles/full_view/21

If there's no relocation table in the image, we don't identify the address as a pointer and so you see the raw number in the asm output. We should be able to detect these easily and provide a name for asm sanitize. (See also: #235)

disinvite:
Fixes #274.

Functions with SEH (structured exception handling) have this format:

```asm
    mov     eax, fs:[0]
    push    ebp
    mov     ebp, esp
    push    -0x1
    push    __ehhandler
; ...

__ehhandler:
    mov     eax, __ehfuncinfo
    jmp     ___CxxFrameHandler
; ...
```

`__ehfuncinfo` is a structure that begins with a magic number and contains information about each of the `Unwind` labels for the function. These are the sections that look like this:

```asm
    lea     ecx,[ebp + -0x10]
    jmp     MxAtomId::~MxAtomId
```

There's more detailed information here: https://www.openrce.org/articles/full_view/21

This update detects the VC6-and-older format for `FuncInfo` and creates entities for both it and the `__ehhandler` label. This will have the most significant benefit to binaries without a relocation table because neither the `push` nor the `mov eax, ____` instructions are identified as having a pointer to resolve during the asm sanitize step.

For binaries with a relocation table, there is the potential for a small improvement where orig and recomp have a diff, resulting in each side not having the same number of sanitize replacements. For example: in `LEGO1`, `OrientableROI::OrientableROI` at `0x100a4420`:

```
0x100a460f : ret
0x100a4610 : -mov eax, <OFFSET10>
           : +mov eax, <OFFSET9>
0x100a4615 : jmp ___CxxFrameHandler (FUNCTION)
0x100a461a : mov ecx, dword ptr [ebp - 0x10]
0x100a461d : jmp ROI::~ROI (FUNCTION)
```

### DIFF
--- a/reccmp/isledecomp/compare/analyze.py
+++ b/reccmp/isledecomp/compare/analyze.py
@@ -79,9 +79,15 @@ def create_seh_entities(db: EntityDb, img_id: ImageId, binfile: PEImage):
             batch.set(
                 img_id,
                 handler_addr,
+                type=EntityType.LABEL,
                 name="__ehhandler",
             )
-            batch.set(img_id, funcinfo.addr, name="__ehfuncinfo")
+            batch.set(
+                img_id,
+                funcinfo.addr,
+                type=EntityType.DATA,
+                name="__ehfuncinfo",
+            )
 
 
 def create_imports(db: EntityDb, image_id: ImageId, binfile: Image):
--- a/reccmp/isledecomp/compare/db.py
+++ b/reccmp/isledecomp/compare/db.py
@@ -537,7 +537,7 @@ class EntityDb:
         """Return the original address (matched or not) that follows
         the one given. If our recomp function size would cause us to read
         too many bytes for the original function, we can adjust it.
-        Skips LINE-type symbols since these these are always contained
+        Skips LINE and LABEL type entities since these these are always contained
         within functions.
         """
         result = self._sql.execute(
@@ -546,10 +546,12 @@ class EntityDb:
             WHERE
               orig_addr > ?
             AND
-              json_extract(kvstore,'$.type') != ?
+              json_extract(kvstore,'$.type') IS NOT NULL
+            AND
+              json_extract(kvstore,'$.type') NOT IN (?, ?)
             ORDER BY orig_addr
             LIMIT 1""",
-            (addr, EntityType.LINE),
+            (addr, EntityType.LINE, EntityType.LABEL),
         ).fetchone()
 
         return result[0] if result is not None else None
--- a/reccmp/isledecomp/types.py
+++ b/reccmp/isledecomp/types.py
@@ -20,6 +20,7 @@ class EntityType(IntEnum):
     VTORDISP = 10
     WIDECHAR = 11
     IMPORT_THUNK = 12
+    LABEL = 13
 
 
 class ImageId(IntEnum):
--- a/tests/test_compare_db.py
+++ b/tests/test_compare_db.py
@@ -4,7 +4,7 @@ import sqlite3
 from unittest.mock import patch
 import pytest
 from reccmp.isledecomp.compare.db import EntityDb
-from reccmp.isledecomp.types import ImageId
+from reccmp.isledecomp.types import EntityType, ImageId
 
 
 @pytest.fixture(name="db")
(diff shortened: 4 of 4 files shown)


reccmp/isledecomp/analysis/funcinfo.py
```diff
@@ -0,0 +1,71 @@
...
+def find_funcinfo_offsets_in_buffer(buf: bytes) -> Iterator[int]:
+    """Return offsets of the FuncInfo magic number."""
+    for match in FUNCINFO_MAGIC_RE.finditer(buf):
+        yield match.start()
+
+
+def find_funcinfo_in_buffer(buf: bytes, base_addr: int) -> Iterator[FuncInfo]:
+    """Parse the FuncInfo struct and return its location."""
+    for ofs in find_funcinfo_offsets_in_buffer(buf):
+        # TODO: The structure may vary depending on the magic string.
```
jonschz, line 40: Add a ticket (can be tagged as "low prio"). Also, from what I read at https://www.openrce.org/articles/full_view/21, the info we read (`unwind_addr`) does not appear to depend on the version. Hard to tell without validating on an example, though.

reccmp/isledecomp/analysis/funcinfo.py
```diff
@@ -0,0 +1,71 @@
...
+        # We support format 19930520 to start.
+        (n_unwind, unwind_addr) = struct.unpack_from("<4x2I", buf, offset=ofs)
```
jonschz: We don't support 16 bit at this time, right? This is something that is sensitive to 16/32/64 bit. If so, no need to change anything right now

jonschz:
```suggestion
        # max_state: number of entries in unwind table
        # p_unwind_map: table of unwind destructors
        (max_state, p_unwind_map) = struct.unpack_from("<4x2I", buf, offset=ofs)
```
I'd stick to the terminology of openRCE

reccmp/isledecomp/analysis/funcinfo.py
```diff
@@ -0,0 +1,71 @@
...
+
+        # Unwind offset is an absolute address.
+        unwind_ofs = unwind_addr - base_addr
+        unwinds = list(
```
jonschz: I'd add a `NamedTuple` for the `UnwindMapEntry` here. The two integers are difficult to interpret without consulting openRCE

reccmp/isledecomp/analysis/funcinfo.py
```diff
@@ -0,0 +1,71 @@
...
+FUNCINFO_MAGIC_RE = re.compile(rb"\x20\x05\x93\x19", flags=re.S)
+
+
+# Match `mov eax, ____` instructions followed by jmp opcode. `B8 .... E9`
+MOV_EAX_RE = re.compile(rb"(?=\xb8(.{4})\xe9)", flags=re.S)
+
+
+class FuncInfo(NamedTuple):
+    addr: int
+    unwinds: list[tuple[int, int]]
```
jonschz:
```suggestion
    unwinds: list[UnwindMapEntry]
```
see also the comment on line 41

reccmp/isledecomp/analysis/funcinfo.py
```diff
@@ -0,0 +1,71 @@
...
+
+# Magic strings:
+# - 0x19930520: up to VC6
+# - 0x19930521: VC7.x(2002-2003)
+# - 0x19930522: VC8 (2005)
+FUNCINFO_MAGIC_RE = re.compile(rb"\x20\x05\x93\x19", flags=re.S)
+
+
+# Match `mov eax, ____` instructions followed by jmp opcode. `B8 .... E9`
+MOV_EAX_RE = re.compile(rb"(?=\xb8(.{4})\xe9)", flags=re.S)
```
jonschz, line 18:
```suggestion
MOV_EAX_RE = re.compile(rb"\xb8(.{4})\xe9", flags=re.S)
```
Is there a reason this wouldn't work?

reccmp/isledecomp/compare/analyze.py
```diff
@@ -68,6 +69,21 @@ def create_analysis_floats(db: EntityDb, img_id: ImageId, binfile: PEImage):
...
 
 
+def create_seh_entities(db: EntityDb, img_id: ImageId, binfile: PEImage):
+    """Create entities for the SEH (structured exception handling)
+    handler and funcinfo struct. For images without a relocation table,
+    this will allow us to replace the addresses for both items."""
+    with db.batch() as batch:
+        for handler_addr, funcinfo in find_eh_handlers(binfile):
+            # Using names derived from symbols in .cpp.s generated asm.
+            batch.set(
```
jonschz, line 79: Just curious: Do these entities have no type? Is that something we support, or that has any use to us?

If there is some default type they are set to, I'd prefer to have that explicit.

reccmp/isledecomp/analysis/funcinfo.py
disinvite: #280

reccmp/isledecomp/compare/analyze.py
disinvite: Good catch: the handler entity not having a type will cause problems for estimating function size using `get_next_orig_addr`. `LINE` might make sense here, but I'll have to see how that works with unmatched entities.

As for what we could derive from this data: automatic pinned lines for each of the unwind maps, matched by target state.

reccmp/isledecomp/analysis/funcinfo.py
disinvite: That was done to allow for overlapping matches in the unlikely event that this occurs. I'll break this out into its own function so we can add a test to demonstrate.

reccmp/isledecomp/compare/analyze.py
disinvite: This turned out not to be a problem because of the way SQL equality works with `NULL`. We don't have tests for `get_next_orig_addr` so I will add some. I'll also change the query so our exclusion of no-type-entities is more explicit.

Issue #278, opened 2025-12-22
print location of unicode error in decomplint


bananapizzuh:
When there is a unknown unicode character in a file, the decomplint fails and only prints the position of the character in the file, but not the filename. (the line would be nice as well)

```
Traceback (most recent call last):
  File "~/.local/bin/reccmp-decomplint", line 7, in <module>
    sys.exit(main())
             ~~~~^^
  File "~/.local/share/pipx/venvs/reccmp/lib/python3.13/site-packages/reccmp/tools/decomplint.py", line 121, in main
    (warning_count, error_count) = process_files(files_to_check, module=args.module)
                                   ~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "~/.local/share/pipx/venvs/reccmp/lib/python3.13/site-packages/reccmp/tools/decomplint.py", line 84, in process_files
    success = linter.check_file(filename, module)
  File "~/.local/share/pipx/venvs/reccmp/lib/python3.13/site-packages/reccmp/isledecomp/parser/linter.py", line 143, in check_file
    return self.read(f.read(), filename, module)
                     ~~~~~~^^
  File "<frozen codecs>", line 325, in decode
UnicodeDecodeError: 'utf-8' codec can't decode byte 0xc0 in position 6160: invalid start byte
```

disinvite:
I think the core problem here is that we naively assumed all code files in a project would be ASCII or UTF-8. See also #229. But we can start with a non-fatal error.

disinvite:
@bananapizzuh I just merged #279 which will report an error but skip the files with a decoding issue. Does that solve the problem on your end?

Issue #229, opened 2025-10-10
Encoding utf-8 may be wrong for old windows files


gynt:
I have windows headers from 2005 with `©` character in them (`\xa9`). The open(encoding=`utf-8`) that reccmp makes is therefore wrong unfortunately.

```
File "C:\Users\Gynt\AppData\Roaming\Python\Python313\site-packages\reccmp\isledecomp\parser\codebase.py", line 23, in __init__
```

The correct encoding should be windows-1252 or ANSI. The easiest solution is that I can specify this option in one of the yml files.

I patched it for myself using a `try-except` block around the statements found in codebase.py, falling back to ANSI.

Issue #127, opened 2025-05-29
QOL features for `decomplint`


disinvite:
- Should support the `--target` parameter similar to other tools.
- Show the fully resolved path for each file. I run `decomplint` from the build directory so the paths shown have the backref like `C:\path\to\lego-island\isle\build\..\ISLE\isleapp.cpp`.
- Group warnings and errors for each file so there's less stuff on screen. We can just show the line number without printing the full file path every time. (IIRC I was imitating `clang-format` with how the alerts are displayed.)
- Show the module for each alert if there is one. For example, `bad_decomp_marker` appears any time we detect it. There is no specific module for the alert. But `duplicate_offset` is scoped by module.

PR #279, opened 2025-12-23
Retool source code file reading to avoid errors

### LINKED ISSUE #224: Use `pathlib.Path` and `PurePath` everywhere
We should use `Path` and `PurePath` over `str` in every place where we need to represent a path.

The main problem is `walk_source_dir` from `dir.py` and all functions using its data. I looked into this a while back and IIRC we get worse performance if we use `Path.walk` or `Path.glob`, but we can just convert to `Path` if we can't work that out.

disinvite:
Fixes #224. Includes some `decomplint` items from #127.

Potential fix for #278 and #229. Specifically that: if a source code file cannot be decoded as UTF-8, we skip it and move on. Supporting more encodings would be a second update. I'm not sure on the best way to do it. We could use [`chardet`](https://github.com/chardet/chardet) to identify the encoding at runtime, or parametrize it in the project yml.

Items of interest for this change:

- `DecompCodebase` now uses the `TextFile` wrapper instead of opening and reading files directly. This is another facet of the broader refactor in #220. Future changes are still on the drawing board but this one feels correct because it improves testability.

- To that point: the tests for source code ingestion no longer create test files using the `pytest` temporary directory. I future-proofed the test in anticipation of this change, so there was not much to remove, although `black` reformatting makes the diff appear bigger.

- There was a bug where `decomplint` would read source files multiple times if you did not specify a path. The current behavior is to combine source paths for each target in the project, but we did not remove duplicates if the paths overlapped. Now fixed.

- General improvement to `decomplint` display, including removing `..` path segments and grouping alerts by file instead of printing the full path for each.

### DIFF
--- a/reccmp/tools/decomplint.py
+++ b/reccmp/tools/decomplint.py
@@ -11,7 +11,7 @@ from reccmp.isledecomp.parser import DecompLinter
 from reccmp.isledecomp.parser.error import ParserAlert
 from reccmp.project.logging import argparse_add_logging_args, argparse_parse_logging
 from reccmp.project.detect import RecCmpProject
-from reccmp.isledecomp.types import TextFile
+from reccmp.isledecomp.formats import TextFile
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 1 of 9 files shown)


reccmp/isledecomp/parser/linter.py
```diff
@@ -13,7 +14,7 @@ class DecompLinter:
     def __init__(self) -> None:
         self.alerts: list[ParserAlert] = []
         self._parser = DecompParser()
-        self._filename: str = ""
+        self._filename: PurePath = PurePath("")
```
jonschz, line 17: Is this intentionally an OS-dependent `PurePath`? Usually, you either have a `Path` (OS-dependent, real path) or one of `PurePosixPath` and `PureWindowsPath` (path that may not correspond to anything on the present system)

reccmp/isledecomp/parser/linter.py
```diff
@@ -119,7 +120,7 @@ def _check_byname_allowed(self):
                     )
                 )
 
-    def read(self, code: str, filename: str, module=None) -> bool:
+    def read(self, code: str, filename: PurePath, module: str | None = None) -> bool:
```
jonschz, line 123: This looks like it should be a `Path`

reccmp/tools/decomplint.py
```diff
@@ -90,35 +100,61 @@ def process_files(files, module=None):
...
+            logger.error("Could not open '%s'", filename)
+
+        except UnicodeDecodeError as ex:
+            logger.error(
+                "Failed to decode '%s' as %s (reason: %s, position: %d)",
+                filename,
+                ex.encoding,
+                ex.reason,
+                ex.start,
+            )
```
jonschz: This error handling is duplicated with `core.py`. Maybe they can be unified in some util function?

jonschz (approved):
Looks very good! I assume you have also checked for regressions manually (e.g. by introducing linter violations to `LEGO1`)?

PR #283, opened 2025-12-25
Option to run Ghidra import despite hash mismatch

### LINKED ISSUE #267: reccmp-ghidra-import: Ignore the executable's hash via a switch
My target EXE keeps changing due to manual byte patches, I'd like an option to ignore the hash from the YAML.

https://github.com/isledecomp/reccmp/blob/9cb8414968bcc9d01a0d292de39a04e8cfb60dcf/reccmp/ghidra/ghidra_import.py#L55

disinvite:
Fixes #267. The new option in the YML is `allow-hash-mismatch`. If set to `true`, we do not abort the Ghidra import if the hash reported by Ghidra does not match the one we have on file in `reccmp-project.yml`.

This only works with the `reccmp-ghidra-import` command line tool. If run from the Ghidra GUI, we depend on a hash match to pick the correct target.

I tested by changing the hash of a target in the project file so I did not have to create a dummy binary with some minor change.

Sample project file:
```yml
targets:
  ISLE:
    filename: ISLE.EXE
    source-root: .
    hash:
    ghidra:
        allow-hash-mismatch: true
```

### DIFF
--- a/docs/annotations.md
+++ b/docs/annotations.md
@@ -153,7 +153,7 @@ MxAtomId* g_introScript = NULL;
 
 ## Strings
 
-String values should be annotated using the `STRING` marker, which includes the module name and address of the text content. Note that this is usually not required since most strings can be auto-detected. If you want, you can use this for bookeeping, but it will usually not affect the `reccmp` match.
+String values should be annotated using the `STRING` marker, which includes the module name and address of the text content. Note that this is usually not required since most strings can be auto-detected. If you want, you can use this for bookkeeping, but it will usually not affect the `reccmp` match.
 
 ```c++
 inline virtual const char* ClassName() const override // vtable+0x0c
--- a/reccmp/assets/components/clickToCopy.js
+++ b/reccmp/assets/components/clickToCopy.js
@@ -14,7 +14,9 @@ class ClickToCopy extends window.HTMLElement {
       this.setAttribute('copied', '');
       // Display "copied!" message for up to 2 seconds
       // if the user keeps their mouse on the link.
-      setTimeout(() => { this.removeAttribute('copied'); }, 2000);
+      setTimeout(() => {
+        this.removeAttribute('copied');
+      }, 2000);
     });
   }
 }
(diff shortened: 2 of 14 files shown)


reccmp/ghidra/ghidra_import.py
```diff
@@ -52,7 +52,7 @@ def main():
         logger.info("Program opened. Starting reccmp import...")
 
         program_hash = program.getExecutableSHA256()
-        if target.sha256 != program_hash:
+        if target.ghidra_config.require_hash_match and target.sha256 != program_hash:
             logger.critical(
                 "The program hashes mismatch (Ghidra: '%s', reccmp project: '%s')",
                 program_hash,
```
madebr, line 66: I'd log always. and do something like this:
```python
        if target.sha256 != program_hash:
            logger.log(
                logging.CRITICAL if target.ghidra_config.require_hash_match else logging.WARNING,
                "The program hashes mismatch (Ghidra: '%s', reccmp project: '%s')",
                program_hash,
                target.sha256,
            )
            if target.ghidra_config.require_hash_match:
                return 1
```
wdyt?

madebr (approved):

reccmp/project/config.py
```diff
@@ -38,10 +38,19 @@ class YmlGhidraConfig(BaseModel):
...
+    )
 
     @classmethod
     def default(cls) -> "YmlGhidraConfig":
-        return cls(ignore_types=[], ignore_functions=[], name_substitutions=[])
+        return cls(
+            ignore_types=[],
+            ignore_functions=[],
+            name_substitutions=[],
+            require_hash_match=True,
```
jonschz: Matter of taste: "true if absent" booleans can be a bit confusing semantically. How about `ignore_hash_mismatch` with inverted logic and default `false`?

disinvite: What do you think of `allow_hash_mismatch`? To avoid the double negative of:
```python
if not target.ghidra_config.ignore_hash_mismatch:
    return 1
```

madebr: Good idea about trying to keep all defaults and/or non-specified options false-like.

PR #291, opened 2026-01-09
Create `LABEL` entities for SEH unwind funclets

disinvite:
More help for binaries without a relocation table (#235).

The new entity name includes the unwind state, but this is not necessarily unique. Long functions are a good example, like `Act1State::PlaceActors` in `LEGO1`, with many unwinds to state `-1` that each call `MxString::~MxString`.

There are a few improvements in `LEGO1` and many in `BETA10`. It seems like these are due to the replaced name solving a mismatch in the numbering of `<OFFSET>` replacements  -- for example, because orig or recomp inlines a function that the other does not.

`BETA10` has two regressions that are caused by the unwind target state not matching, where this had previously been ignored because the `<OFFSET>` numbers _did_ line up. These are caused by the larger difference in retail vs. beta code.


jonschz (approved):
Haven't tried it or looked too deeply into it, but looks reasonable. The regressions you describe also sound reasonable.

Would you consider test coverage to be overkill? If it's a feature someone would rely on in the future, it may be worth considering.

Issue #190, opened 2025-08-15
Revamp `Image` subclasses


disinvite:
Following #173. Some ideas:

- [ ] `Image` is effectively an abstract class so it should use the Python `abc` module.
- [ ] We are using `dataclass` for simplicity when creating the image objects. We might be able to move some shared logic into regular `__init__` functions.
- [ ] The images should have the option to load from both memory (byte string) and from a file.

disinvite:
We've made some strides here. Let's close until there are new specific items to address.

PR #296, opened 2026-01-11
Generic `ImageSection` dataclass

disinvite:
Replaces `PESection` with a more generic `ImageSection` dataclass. Also adds a generic `is_valid_vaddr` function to the main `Image` class. Both were intended to address the `roadmap` issues in #295 but it's not a priority to get that working so this should be considered as its own change.

I don't think this is a potential footgun because Ghidra has basically the same setup under Window > Memory Map.

Ghidra goes a step further and splits the uninitialized tail of some sections into a second mock section. I didn't do that here because it would complicate accessing by the section index and we already have methods to tell if an address falls in a `bss` area.

I still need to test this with my stashed changes for `NEImage` support to make sure it will be compatible. (IIRC the relocations are embedded in the section dataclass so I could either extend `ImageSection` or keep the information separate.)

Could be considered part of #190.

### DIFF
--- a/reccmp/isledecomp/formats/pe.py
+++ b/reccmp/isledecomp/formats/pe.py
@@ -433,11 +433,12 @@ class ExportDirectoryTable:
     ordinal_table_rva: int
 
 
-def create_pe_sections(
+def get_pe_sections(
     section_headers: Iterable[PEImageSectionHeader],
     image_base: int,
     view: memoryview,
 ) -> Iterator[ImageSection]:
+    """Helper that creates ImageSections based on the PE section headers."""
     for s in section_headers:
         virtual_start = image_base + s.virtual_address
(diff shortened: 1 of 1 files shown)


reccmp/isledecomp/formats/pe.py
```diff
@@ -453,15 +433,44 @@ class ExportDirectoryTable:
     ordinal_table_rva: int
 
 
+def create_pe_sections(
```
jonschz: Minor point: `create` sounds a bit like something is created as a side effect. How about `get_pe_sections` or something along these lines?

jonschz (approved):
Nice! I can't speak for how well this applies to other formats beyond PE, but the interface can always be changed later if needed.

Issue #272, opened 2025-12-19
Improve `datacmp` display of raw data blobs


disinvite:
Some variables do not have type information and so we cannot present a structured diff using `datacmp`. This most commonly occurs with library variables that appear only as a public symbol in the PDB without a type. If we define them as variables in the orig annotation data, they will be compared by `datacmp`.

The only thing we can do in this case is diff the raw data. If the size is not known (or even when it is) this can result in a huge block of text. This is especially bad if it is a library variable because there is currently no way to provide a type and improve the display.

Ideas:
1. Hide raw data diffs behind the `--verbose` flag.
2. Truncate to some arbitrary number of bytes (and optionally display the entire thing with `--verbose`).
3. Display as a hex dump instead of printing the raw `bytes` string.

PR #298, opened 2026-01-17
Improve `datacmp` display

### LINKED ISSUE #271: `datacmp` hides value for initialize mismatch
Consider this case:
```
pgbh_InitDLL (0x626c54) ... DIFF
      0 (value)                        (initialized) : (uninitialized)
```
In orig, address `0x626c54` has the initial value of a pointer to the function at `0x5ea4d0`. In the recomp build, it is uninitialized.

This diff display makes the problem look like recomp is uninitialized but should be initialized to zero.

In general, this presentation could be improved.

disinvite:
Fixes #271 and #272.

- Comparisons of raw data (variables without a data type) are now treated as arrays of `u8` with more surgical diffs instead of dumping the entire data blob.
- `CARM95` has a variable with a bitfield. We do not support these so an exception is raised by the type db. We had previously refused to compare the variable, but we now compare as raw and simply log the error.
- `105` (GTA2) has a struct with no members. In the PDB this appears as an `LF_STRUCTURE` of size 1 with only methods in the `LF_FIELDLIST`. If we cannot pull any members for the struct, we compare as raw.
- We now have a more rigid structure for determining whether a variable is uninitialized or if it merely _could_ be. If we can tell for certain that the variable is (un)initialized in one binary and not the other, this is raised as a `DIFF`. We had previously flagged these as `WARN` which makes it unclear what the user can do about it.
- Per #271, we now display all initialized values.

### DIFF
--- a/reccmp/tools/datacmp.py
+++ b/reccmp/tools/datacmp.py
@@ -385,10 +385,6 @@ def do_the_comparison(target: RecCmpTarget) -> Iterable[ComparisonItem]:
         )
 
 
-def value_get(value: str | None, default: str):
-    return value if value is not None else default
-
-
 def colorize_match_result(result: CompareResult, no_color: bool) -> str:
     """Helper to return color string or not, depending on user preference"""
     if no_color:
@@ -403,6 +399,38 @@ def colorize_match_result(result: CompareResult, no_color: bool) -> str:
             color = colorama.Fore.YELLOW
 
     return f"{color}{result.name}{colorama.Style.RESET_ALL}"
+
+
+def compared_offset_string(c: ComparedOffset, no_color: bool) -> str:
+    """Display the offset, name, and value diff for each compared item.
+    Scalar variables have only a single item, the value itself."""
+
+    def ansi_wrap(c: str):
+        """Easily remove the ANSI codes in --no-color mode."""
+        return "" if no_color else c
+
+    offset = f"+ 0x{c.offset:02x}"
+    header_chunk = [ansi_wrap(colorama.Fore.LIGHTBLACK_EX), f"{offset:>10}"]
+
+    name_chunk = [
+        ": " if c.name else "  ",
+        ansi_wrap(colorama.Fore.WHITE),
+        f"{c.name if c.name else '':30}",
+    ]
+
+    (value_a, value_b) = c.values
+    values_chunk = [ansi_wrap(colorama.Fore.LIGHTWHITE_EX), value_a]
+    if not c.match:
+        values_chunk.extend([" : ", ansi_wrap(colorama.Fore.LIGHTBLACK_EX), value_b])
+
+    return " ".join(
+        [
+            "".join(header_chunk),
+            "".join(name_chunk),
+            "".join(values_chunk),
+            ansi_wrap(colorama.Style.RESET_ALL),
+        ]
+    )
 
 
 def main():
@@ -444,13 +472,7 @@ def main():
             if not args.verbose and c.match:
                 continue
 
-            (value_a, value_b) = c.values
-            if c.match:
-                print(f"  {c.offset:5} {value_get(c.name, '(value)'):30} {value_a}")
-            else:
-                print(
-                    f"  {c.offset:5} {value_get(c.name, '(value)'):30} {value_a} : {value_b}"
-                )
+            print(compared_offset_string(c, args.no_color))
 
         if args.verbose:
             print()


reccmp/tools/datacmp.py
```diff
@@ -158,140 +228,103 @@ def do_the_comparison(target: RecCmpTarget) -> Iterable[ComparisonItem]:
...
-                continue
+
+                # Make sure we can retrieve struct or array members.
+                if isle_compare.types.get_format_string(type_name):
+                    raw_only = False
+
+            except (CvdumpKeyError, CvdumpIntegrityError):
+                # TODO: This may occur even when nothing is wrong, so permit a raw comparison here.
+                # For example: we do not handle bitfields and this complicates fieldlist parsing
+                # where they are used.
```
jonschz: I don't quite understand this TODO. If there is something that you know could be done reasonably, please open a ticket. Otherwise, I'd be fine with leaving this comment in without the `TODO` prefix.

reccmp/tools/datacmp.py
```diff
@@ -70,13 +73,63 @@ def parse_args() -> argparse.Namespace:
...
 
 
+class BssState(Enum):
+    """Determination of whether this variable is uninitialized.
+    - NO:    At least one byte between the variable's start and the
+             end of the section is non-zero.
+    - MAYBE: The variable is fully initialized to zero. All remaining
+             initialized bytes in the section are zero.
+    - YES:   All or part of the variable is uninitialized.
+    """
```
jonschz, line 136: Optional: Maybe add 1-2 sentences describing what the bss section in Win32 PE does, and why we aren't sure whether it is initialised.

reccmp/tools/datacmp.py
```diff
@@ -70,13 +73,63 @@ def parse_args() -> argparse.Namespace:
...
+    @classmethod
+    def read(cls, addr: int, size: int, image: Image) -> Self:
+        data = image.read(addr, size)
+        # Per the seek() API, phys_data is a memoryview of the remaining
+        # physical bytes in this section.
+        (phys_data, _) = image.seek(addr)
+
+        # If we find any non-zero bytes then this variable must be initialized
+        # even if all of the variable's values are zero.
+        init_bytes_remain = re.search(b"[^\x00]", phys_data) is not None
```
jonschz, line 164: Do I understand correctly that within a bss section, the initialised bytes are always at the front and the uninitialised ones are always in the back? Is this an MSVC convention?

reccmp/tools/datacmp.py
```diff
@@ -158,140 +228,103 @@ def do_the_comparison(target: RecCmpTarget) -> Iterable[ComparisonItem]:
...
+            compare_items = [
+                DataOffset(offset=sc.offset, name=sc.name or "", pointer=sc.is_pointer)
+                for sc in isle_compare.types.get_scalars_gapless(type_name)
+            ]
+            format_str = isle_compare.types.get_format_string(type_name)
+
+            orig_data = unpack(format_str, orig_block.data)
+            recomp_data = unpack(format_str, recomp_block.data)
 
         # Could zip here
```
jonschz: Either zip or remove the comment

reccmp/tools/datacmp.py
```diff
@@ -158,140 +228,103 @@ def do_the_comparison(target: RecCmpTarget) -> Iterable[ComparisonItem]:
...
                 data_type = isle_compare.types.get(type_name)
                 assert data_type.size is not None
                 data_size = data_type.size
-            except (CvdumpKeyError, CvdumpIntegrityError) as ex:
-                yield create_comparison_item(var, error=repr(ex))
-                continue
+
+                # Make sure we can retrieve struct or array members.
+                if isle_compare.types.get_format_string(type_name):
+                    raw_only = False
```
jonschz, line 296: Would the `else` case here warrant a warning / error log as well? I.e. are there any expected non-error cases where `type_name is not None` but we don't get a result from `get_format_string`? I added an `else` for testing but couldn't hit it on `LEGO1` or `BETA10`.

reccmp/tools/datacmp.py
```diff
@@ -158,140 +228,103 @@ def do_the_comparison(target: RecCmpTarget) -> Iterable[ComparisonItem]:
...
+                value_b = str(recomp_data[i])
+
+            # Invalidate the match if there is a definite conflict between
+            # the initialized state in orig and recomp.
+            if (orig_block.bss == BssState.NO and recomp_block.bss == BssState.YES) or (
+                recomp_block.bss == BssState.NO and orig_block.bss == BssState.YES
+            ):
+                match = False
+
+            if orig_data[i] == 0 and orig_block.bss == BssState.YES:
```
jonschz: Is the check for `orig_data[i] == 0` required? Intuitively, I would have thought that `BssState.YES` wouldn't be possible if  `orig_data[i] != 0`. Same below for recomp.

reccmp/tools/datacmp.py
```diff
@@ -141,6 +194,23 @@ def create_comparison_item(
     )
 
 
+def pointer_display(isle_compare: IsleCompare, addr: int, is_orig: bool) -> str:
```
jonschz, line 250: Suggestion: Use `reccmp.isledecomp.types.ImageId` instead of `is_orig: bool`

reccmp/tools/datacmp.py
```diff
@@ -344,6 +376,9 @@ def display_match(result: CompareResult) -> str:
         if item.error is not None:
             print(f"  {item.error}")
 
+        if item.raw_only:
+            print("  Comparing raw data only.")
```
jonschz:
```suggestion
            print("  Unknown or unsupported data type, comparing raw data only.")
```
maybe something along those lines?

reccmp/tools/datacmp.py
```diff
@@ -312,16 +345,15 @@ def display_match(result: CompareResult) -> str:
...
-        return f"{match_color}{result.name}{colorama.Style.RESET_ALL}"
+        match result:
+            case CompareResult.MATCH:
+                color = colorama.Fore.GREEN
+            case CompareResult.ERROR | CompareResult.DIFF:
+                color = colorama.Fore.RED
+            case _:
+                color = colorama.Fore.YELLOW
+
+        return f"{color}{result.name}{colorama.Style.RESET_ALL}"
```
jonschz: Matter of taste: This function could also be pulled to the root level, with `args.no_color` becoming an ordinary function argument

reccmp/tools/datacmp.py
```diff
@@ -158,140 +228,103 @@ def do_the_comparison(target: RecCmpTarget) -> Iterable[ComparisonItem]:
...
+                DataOffset(offset=sc.offset, name=sc.name or "", pointer=sc.is_pointer)
+                for sc in isle_compare.types.get_scalars_gapless(type_name)
+            ]
+            format_str = isle_compare.types.get_format_string(type_name)
+
+            orig_data = unpack(format_str, orig_block.data)
+            recomp_data = unpack(format_str, recomp_block.data)
 
         # Could zip here
+        compared = []
```
jonschz, line 347: Is this variable needed? Why not yield directly in the loop?

reccmp/tools/datacmp.py
```diff
@@ -158,140 +228,103 @@ def do_the_comparison(target: RecCmpTarget) -> Iterable[ComparisonItem]:
...
-            )
-
-            if ptr_match is not None:
-                return f"Pointer to {ptr_match.match_name()}"
-
-            # This variable did not match if we do not have
-            # the pointer target in our DB.
-            return f"Unknown pointer 0x{addr:x}"
+            compare_items = [
+                DataOffset(offset=i, name=f"[{i}]", pointer=False)
```
jonschz: Personally, I find the diff output for raw a bit confusing, e.g.:
```
__OP_EXPjmptab (0x1010296a) ... WARN
  Comparing raw data only.
     16 [16]                           1 : 65
     17 [17]                           36 : 236
     18 [18]                           13 : 12
     20 [20]                           221 : 157
     21 [21]                           48 : 34
     24 [24]                           226 : 162
     25 [25]                           48 : 34
     28 [28]                           65 : 129
     29 [29]                           37 : 237
     30 [30]                           13 : 12
```
The first and second number (`16 [16]`) are always the same, but the display suggests that they convey different information. I'd probably remove the one in brackets entirely. Maybe we can also go for hex with the offset to make it more obvious, e.g.
```
__OP_EXPjmptab (0x1010296a) ... WARN
  Comparing raw data only.
     +0x10                           1 : 65
```
We could e.g. use `print(f" +{c.offset:#7x} [...]` in lines 388 and 390, and set `name` to `""` here.

disinvite:
> However, If I change all `MxAtomId* g_[...] = NULL;` to `MxAtomId* g_[...];` in the file (changing one in the middle appears to have no effect), I get a diff
> 
> ```
> g_infomainScript (0x102114dc) ... DIFF 
>   +    0x0                                (uninitialized) : Unknown pointer 0x5c646c69
> ```

This might be a similar situation to #210. The `Script` variables all have an `extern` reference in `scripts.h` and so they appear twice in the PDB:

```
(GLOBALS section, sorted by seg:offset)
S_GDATA32: [0004:0000C718], Type:             0x10D5, g_infomainScript
...
S_GDATA32: [0004:0002F6BC], Type:             0x10D5, g_infomainScript
```

Only the second entry has a corresponding symbol from `PUBLICS`:
```
S_PUB32: [0004:0002F6BC], Flags: 00000000, ?g_infomainScript@@3PAVMxAtomId@@A
```

Output from `datacmp` with `--print-rec-addr`:
```
g_infomainScript (0x102114dc / 0x101df718) ... DIFF
      0                                (uninitialized) : Unknown pointer 0x5c646c69
```

The tail end of the address shows we are matching against the first entry.

reccmp/tools/datacmp.py
disinvite: This is true for [PE images](https://learn.microsoft.com/en-us/windows/win32/debug/pe-format#section-table-section-headers):

> The size of the section (for object files) or the size of the initialized data on disk (for image files). For executable images, this must be a multiple of FileAlignment from the optional header. **If this is less than VirtualSize, the remainder of the section is zero-filled.** Because the SizeOfRawData field is rounded but the VirtualSize field is not, it is possible for SizeOfRawData to be greater than VirtualSize as well. When a section contains only uninitialized data, this field should be zero.

jonschz: Nice, maybe add that link somewhere in the file (here in that comment or elsewhere)

reccmp/tools/datacmp.py
disinvite: We would hit it for a struct without any members (e.g. one with only functions). I'll add a `logging.info` message since this is not really a user problem.

reccmp/tools/datacmp.py
disinvite: The only known problem to fix would be bitfields. I'll add a reference to #299.

reccmp/tools/datacmp.py
disinvite: Going to leave this for now because it would require a change to `core.py` and I don't want to disturb another thing I have in the pipeline. But I have been meaning to address these. #300

disinvite:
> I really like the refactor, the readability is greatly improved.
> 
> General remark: Do you think it would be possible to separate this file into two, one for the data comparison logic and one for the CLI parsing / output generation? That way, it may be possible to unit test the former. Right now, your fixes do not have any test coverage.
> 
> Good to merge once my comments have been resolved. Please ping me if you'd like to get a re-review after a large change.

Thanks as always for the detailed review! I agree we should break the compare logic out so we can test it, but I'll do it in a second PR.

reccmp/tools/datacmp.py
disinvite: It ended up being more than 1-2 sentences 😅 but hopefully this clarifies the assumption we're making.

reccmp/tools/datacmp.py
disinvite: We have to build the list of compared offsets first, then add this to the larger `ComparisonItem` with the variable details.

I tried breaking this out into a function just now, but there's too many dependencies to drag in so it doesn't really work. I agree we should refactor it further after we add some tests.

reccmp/tools/datacmp.py
disinvite: Thanks, I added this and it looks a lot better. I went with a minimum of two digits for the offset (`0x00` not `0x0`) similar to the convention we used over in `isle`.

I also darkened some of the text and highlighted the orig value so the diff is more clear. What do you think?

jonschz: Wow, that looks very nice indeed!

reccmp/tools/datacmp.py
jonschz: You're right. Let's get this one merged and consider further refactoring later.

PR #302, opened 2026-01-19
Detect switch data tables accessed with `movzx`

disinvite:
Data tables are a byte array lookup table that converts sparse `case` values to an offset for a jump table of addresses.

The typical setup is:
```
Excerpt from "Score::Notify"

1000143b 33 c0      XOR     EAX,EAX
1000143d 8a 81      MOV     AL,byte ptr [this + 0x100014ec]=
        ec 14 
        00 10
                switchD_10001443::switchD
10001443 ff 24      JMP     dword ptr [EAX*0x4 + switchD_100  = 1000144a
        85 d4 
        14 00 10
```

This will now detect a data table if a `movzx` or `mov` instruction is used with a displacement pointer.

### DIFF
--- a/README.md
+++ b/README.md
@@ -31,14 +31,14 @@ The next steps differ based on what kind of project you have.
 ### Contributing to a project that already uses `reccmp`
 
 1. Compile the C++ project.
-2. Run `reccmp-project detect --search-path path/to/folder/with/original/binaries`.
+2. Run `reccmp-project detect --search-path "path/to/folder/with/original/binaries"`.
 3. If there is no `reccmp-build.yml` after building: Navigate to the recompiled binaries folder and run `reccmp-project detect --what recompiled`.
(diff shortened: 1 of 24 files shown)


tests/test_instgen.py
```diff
@@ -214,3 +214,41 @@ def test_action_case():
...
+)
+
+
+def test_movzx_data_table():
+    """Data table is accessed with a MOVZX instruction:
+    00451df7 0f b6 88        MOVZX      this,byte ptr [EAX + 0x451ed5]=>PTR_caseD_7_00
+             d5 1e 45 00
+    """
+    ig = InstructGen(REPORT_SAVE_LOAD_EXCEPTION, 0x451D73)
+    assert len(ig.sections) == 3
```
jonschz, line 254: I'd also assert on the types of the three sections, like (schematically)
```py
assert [section.type for section in ig.sections] == [code, tab, tab]
```

Issue #93, opened 2025-03-05
Convert report JSON to other output artifacts


disinvite:
We use `reccmp-reccmp` to generate SVG and HTML files, but it should be possible to do this using the report JSON file. We can't do it yet because some of the required information is not there.

For HTML output, we can drop in the report JSON as it is now, but it is missing the diff so it's not really that useful. We may also want to change the format to store the diff data more efficiently. There is a lot of JSON overhead in the way we do it now, and the type hints on the Python side are similarly convoluted. We can offload some of the processing of preparing the display onto the browser. (Might do a separate issue just for this.)

For the SVG, we need to indicate which entities are functions so we can calculate total effective accuracy and other values. We store vtables alongside functions in the current report, so we can't derive the number with what we have. (Although a simple text match against "vtable" would work with the current naming conventions.)

disinvite:
#486 provides the report changes needed for this. The only thing left will be a `reccmp-report` or similar tool that translates between the artifact formats.

PR #307, opened 2026-01-24
Generate unified diff only when producing output

disinvite:
The process of generating a diff for a function or vtable is:
1. Get the `ReccmpMatch` object from the database. Assuming that exists:
2. Get the raw list of "items" for the entity. For a function this is each address and instruction. For a vtable this is the offset and function of each entry in the table.
3. Use `difflib.SequenceMatcher` to compare the raw text (i.e. not the addresses or offsets) and produce diff opcodes. These are instructions from the set `(equal, insert, delete, replace)` that describe how to turn sequence A (orig) into sequence B (recomp).
4. Generate a unified diff with a certain number of context lines. This breaks the diff into groups and eliminates long sequences of matched items. For functions, a group will begin or end with up to 10 lines of matched lines for context. For vtables, we retain all matched lines but still use the udiff structure, so it is one big group.
5. Set the result in a `DiffReport` object.
6. Output the diff. If we are creating a JSON or HTML report, convert the `DiffReport` to a `ReccmpComparedEntity` object first, then follow the serialization logic.

Some data is lost between steps 3 and 4. We don't need to generate the unified diff until step 6 when it is time to output to a file or to the screen.

I changed `stackcmp` to use the entire diff report instead of the grouped version. Is that useful at all? Or not needed?

My end goal is to create report format version 2 that will store this data more efficiently. (#98) JSON deserialization becomes a performance problem if you are aggregating a lot of entropy runs. For the 1024 sample set, IIRC the bulk of the time is spent in JSON parsing.

We also do not store the entity type in the JSON report. If we had this, it would unblock #93.

This is a draft for now because I'm still trying to figure out what kind of automated tests would help verify this change. Ideally we would commit the tests as a separate PR first. The _manual_ testing is straightforward but tedious. (Diffing JSON files from many decomp projects.)

I would appreciate any and all preliminary feedback!

### DIFF
--- a/reccmp/tools/stackcmp.py
+++ b/reccmp/tools/stackcmp.py
@@ -346,6 +346,8 @@ def main():
 
     assert match.result is not None
     # Analyze the entire function, including long sections that already match.
+    # This comment explains why this is necessary:
+    # https://github.com/isledecomp/reccmp/pull/307#issuecomment-3796146436
     udiff = raw_diff_to_udiff(match.result.diff, grouped=False)
 
     function_data = next(


reccmp/compare/report.py
```diff
@@ -107,6 +123,30 @@ def combine_reports(samples: list[ReccmpStatusReport]) -> ReccmpStatusReport:
     return output
 
 
+def get_udiff_for_entity(entity: ReccmpComparedEntity) -> CombinedDiffOutput | None:
```
jonschz, line 126: I don't have any good ideas right now, but can we name this function better so it is clearer under which circumstances it returns `None`? Or add a docstring?

reccmp/compare/report.py
```diff
@@ -107,6 +123,30 @@ def combine_reports(samples: list[ReccmpStatusReport]) -> ReccmpStatusReport:
...
+    if entity.udiff is not None:
+        # An aggregate report may already have a deserialized udiff.
+        return entity.udiff
+
+    if entity.rdiff is None:
+        # We need data to create the unified diff.
+        return None
```
jonschz, line 142: Is this function supposed to be called when both `udiff` and `rdiff` are `None`? Is that a potential error where an assertion would be more appropriate? If not, I'd add a comment to explain under which circumstances we can end up in this branch.

reccmp/tools/stackcmp.py
```diff
@@ -343,7 +344,9 @@ def main():
         print(f"Failed to find a match at address 0x{args.address:x}")
         return 1
 
-    assert match.udiff is not None
+    assert match.result is not None
+    # Analyze the entire function, including long sections that already match.
+    udiff = raw_diff_to_udiff(match.result.diff, grouped=False)
```
jonschz, line 351: Nice!
```suggestion
    assert match.result is not None
    # Analyze the entire function, including long sections that already match.
    # See https://github.com/isledecomp/reccmp/pull/307#issuecomment-3796146436 why that is useful.
    udiff = raw_diff_to_udiff(match.result.diff, grouped=False)
```

reccmp/compare/report.py
```diff
@@ -16,13 +17,19 @@ class ReccmpReportSameSourceError(Exception):
...
+    # Version 1 files have no type, so it is optional.
+    type: EntityType | None = None
     recomp_addr: str | None = None
     is_effective_match: bool = False
     is_stub: bool = False
-    diff: CombinedDiffOutput | None = None
+    rdiff: RawDiffOutput | None = None
+
+    # Legacy field for importing version 1 files (aggregate).
+    udiff: CombinedDiffOutput | None = None
```
jonschz, line 32: I am a little confused: Does this PR introduce a version 2 for the file type? I don't see it referenced anywhere.

reccmp/compare/report.py
```diff
@@ -132,7 +172,7 @@ class JSONReportVersion1(BaseModel):
 def _serialize_version_1(
     report: ReccmpStatusReport, diff_included: bool = False
 ) -> JSONReportVersion1:
-    """The HTML file needs the diff data, but it is omitted from the JSON report."""
+    """The JSON report can exclude the diff to make deserialization faster."""
     entities = [
         JSONEntityVersion1(
```
jonschz, line 186: This still references "Version 1" (`_serialize_version_1`, `JSONReportVersion1`)

reccmp/compare/core.py
```diff
@@ -220,7 +220,7 @@ def debug(self, debug: bool):
         self._debug = debug
         self.function_comparator.debug = debug
 
-    def _compare_vtable(self, match: ReccmpMatch) -> DiffReport:
+    def _compare_vtable(self, match: ReccmpMatch) -> FunctionCompareResult:
```
jonschz: Is the class name `FunctionCompareResult` still appropriate if it can also hold the compare result of a VTable?

jonschz (approved):
Looks very solid overall! I have faith in your manual testing, I didn't do any.

reccmp/compare/report.py
disinvite: No, still just version 1. That field is there to future-proof version 2+ where we will (in some fashion) store the entire diff. The internal representation of the report is `ReccmpComparedEntity` so it needs to support both formats.

`rdiff` can be converted to `udiff` but the reverse is not possible because the matching sections outside the diff groups have already been lost.

reccmp/compare/report.py
disinvite: The version hasn't changed yet so this should be fine. For this diff in particular, I changed the docstring because it's not accurate. We do store the diff in the JSON report unless you provide the `--json-diet` switch.

reccmp/compare/core.py
disinvite: They are the same for now, so maybe a type alias would work?

reccmp/compare/report.py
disinvite: Yes, although I wasn't sure until I found the gap in our tests: `reccmp-aggregate` with a matching function.
1. Write report file with a null udiff for function entities that match 100%.
2. Deserialize the sample reports. Both `udiff` and `rdiff` are `None` for entities like the one in step 1.
3. Pick the highest scoring entities and write the aggregate report.

reccmp/compare/core.py
jonschz: How about `EntityCompareResult`?

PR #309, opened 2026-01-25
Remove references to `Lego` and `Isle` in names

### LINKED ISSUE #202: Remove `Lego*` names from `reccmp`
This mostly affects the Ghidra import, but there are a few places in the documentation (Readme + docstrings) which should be updated as well.

Needs to be done after #198 to reduce collisions.

jonschz:
Closes #202 

Sorry for all the merge conflicts 😅 Almost all of the changes come from renaming the `isledecomp` module to `recomp`, so it shouldn't be too hard to fix those.

### DIFF
--- a/reccmp/tools/roadmap.py
+++ b/reccmp/tools/roadmap.py
@@ -397,8 +397,10 @@ def main() -> int:
     module_map = ModuleMap(target.recompiled_pdb, recomp_bin)
 
     def is_same_section(orig: int, recomp: int) -> bool:
-        """Compare the section name instead of the index.
-        LEGO1.dll adds extra sections for some reason. (Smacker library?)"""
+        """
+        It is better to compare the sections by name since orig and recomp might not have the same number of sections.
+        We have encountered this e.g. for LEGO1.
+        """
 
         try:
             orig_name = orig_bin.sections[orig - 1].name


reccmp/tools/roadmap.py
jonschz: This is the only reference to `LEGO` where I wasn't sure what to do. @disinvite can you comment on that?

tests/test_compare_ingest_cvdump.py
```diff
@@ -453,7 +453,7 @@ def test_variable_size_with_type_info(binfile: PEImage):
                 Element type = T_RCHAR(0070)
                 Index type = T_SHORT(0011)
                 length = 1024
-                Name = 
+                Name =
```
jonschz: A few whitespace differences got in. Should I revert those?

reccmp/tools/roadmap.py
disinvite: We could generalize it to say that orig and recomp may not have the same number of sections and that's why it's better to use the name.

tests/test_compare_ingest_cvdump.py
disinvite: Those could disrupt the regular expressions that parse cvdump output. However... all the tests still pass, so this has apparently not happened.

reccmp/isledecomp/__init__.py
jonschz, line 1: I removed this entirely. Less than 20 lines of code got a little bit longer because of it. Worth it in my opinion.

PR #312, opened 2026-01-26
Docs: add quotes to path arguments for cross-platform compatibility

MystikReasons:
I'm mostly on Linux but due to an old Windows game which I wanted to test this tool with I've decided to do this on Windows itself.

CMD and PowerShell behave a bit differently when it comes to paths, they need to be quoted. Some of the existing examples work fine on Linux/macOS but can be confusing or fail on Windows because of this.

This PR tweaks the documentation to make things clearer by:

- Separating Linux/macOS and Windows (CMD / PowerShell) command examples
- Showing properly quoted paths for Windows
- Keeping the Unix examples exactly as they were

No functional changes, just making the docs easier to follow and more copy-paste friendly for Windows users.

### DIFF
--- a/README.md
+++ b/README.md
@@ -31,14 +31,14 @@ The next steps differ based on what kind of project you have.
 ### Contributing to a project that already uses `reccmp`
 
 1. Compile the C++ project.
-2. Run `reccmp-project detect --search-path path/to/folder/with/original/binaries`.
+2. Run `reccmp-project detect --search-path "path/to/folder/with/original/binaries"`.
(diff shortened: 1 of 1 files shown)


disinvite:
What kinds of paths did not work on Windows?

MystikReasons:
@disinvite 

We could also keep the commands together and just add the quoting to the paths. This would work for all platforms and make it more clear.

MystikReasons:
@disinvite done 🙂

PR #314, opened 2026-02-06
More "safe" asm diff fixes for MSVC entropy

rlerrr:
This applies a couple extra fixes for MSVC entropy I've encountered that seem very safe:

1. Treats "test" instruction just like "cmp"

MSVC likes to do the same shenanigans "randomly" with test as for cmp.
BTW can patch_mov_compare_jmp conceivably cause false negatives?  The reordered instructions don't leave the register in the same state.  Intermediate reuse largely doesn't exist in unoptimized code, but maybe with optimizations?

2. Add fix for consecutive mov/add (and other commutative integer instructions) with swapped operands.

Similar to fld/fmul fix but has to contend with different target registers and more operators.

### DIFF
--- a/tests/test_fixes.py
+++ b/tests/test_fixes.py
@@ -374,3 +374,25 @@ def test_this_should_not_be_marked_as_effective():
     is_effective = find_effective_match(diff.get_opcodes(), orig_asm, recomp_asm)
 
     assert is_effective is False
+
+
+@pytest.mark.xfail(reason="Limitation of fix_mov_cmp_jmp")
+def test_fix_mov_cmp_jmp_unsafe_intermediate_reuse():
+    # These are NOT equivalent since eax is used after the jmp
+    orig_asm = [
+        "mov eax, dword ptr [ebp - 8]",
+        "cmp eax, dword ptr [ebp - 4]",
+        "jl 0x2",
+        "mov dword ptr [ebp - 0xc], eax",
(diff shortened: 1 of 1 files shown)


rlerrr:
Tests added!

> Can you give an example of what you mean? That might be useful to test as well.

Sorry, struggling to explain precisely. Here's a test with some contrived assembly that hopefully shows the theoretical issue:

```
def test_fix_mov_cmp_jmp_unsafe_intermediate_reuse():
    # These are NOT equivalent since eax is used after the jmp
    orig_asm = [
        "mov eax, dword ptr [ebp - 8]",
        "cmp eax, dword ptr [ebp - 4]",
        "jl 0x2",
        "mov dword ptr [ebp - 0xc], eax"
    ]
    recomp_asm = [
        "mov eax, dword ptr [ebp - 4]",
        "cmp eax, dword ptr [ebp - 8]",
        "jg 0x2",
        "mov dword ptr [ebp - 0xc], eax"
    ]

    diff = difflib.SequenceMatcher(None, orig_asm, recomp_asm)
    is_effective = find_effective_match(diff.get_opcodes(), orig_asm, recomp_asm)

    assert is_effective is False
```

`patch_mov_cmp_jmp` says this is an effective match  (this test fails), but the result in `[ebp - 0xc]` is actually different if the branch isn't taken.

Issue #299, opened 2026-01-18
Cvdump: Handle `LF_BITFIELD`


disinvite:
When parsing cvdump output, we do not parse bitfield leaves at all. This creates a broken link in the fieldlist structure and so loading a datatype that contains a bitfield (at any level) will raise an exception.

We could treat the bitfield as raw data but we have to determine the memory footprint. If any compilers would allow you to mix bitfields and regular struct members then this will be tricky. (Need to check on this.)

Real world example from `CARM95`:

```cpp
typedef struct tCar_controls { // size: 0x4
    int joystick_acc : 8;
    int joystick_dec : 8;
    unsigned int left : 1;        // 0x10000   bit 17
    unsigned int right : 1;       // 0x20000   bit 18
    unsigned int acc : 1;         // 0x40000   bit 19
    unsigned int dec : 1;         // 0x80000   bit 20
    unsigned int brake : 1;       // 0x100000   bit 21
    unsigned int up : 1;          // 0x200000   bit 22
    unsigned int down : 1;        // 0x400000   bit 23
    unsigned int holdw : 1;       // 0x800000   bit 24
    unsigned int backwards : 1;   // 0x1000000   bit 25
    unsigned int change_up : 1;   // 0x2000000   bit 26
    unsigned int change_down : 1; // 0x4000000   bit 27
    unsigned int horn : 1;        // 0x8000000   bit 28
} tCar_controls;
```

```
0x16cb : Length = 10, Leaf = 0x1205 LF_BITFIELD
	bits = 8, starting position = 0, Type = T_INT4(0074)

0x16cc : Length = 10, Leaf = 0x1205 LF_BITFIELD
	bits = 8, starting position = 8, Type = T_INT4(0074)

0x16cd : Length = 10, Leaf = 0x1205 LF_BITFIELD
	bits = 1, starting position = 16, Type = T_UINT4(0075)

0x16ce : Length = 10, Leaf = 0x1205 LF_BITFIELD
	bits = 1, starting position = 17, Type = T_UINT4(0075)

0x16cf : Length = 10, Leaf = 0x1205 LF_BITFIELD
	bits = 1, starting position = 18, Type = T_UINT4(0075)

0x16d0 : Length = 10, Leaf = 0x1205 LF_BITFIELD
	bits = 1, starting position = 19, Type = T_UINT4(0075)

0x16d1 : Length = 10, Leaf = 0x1205 LF_BITFIELD
[...]

PR #315, opened 2026-02-07
Use `int` as basis for `CvdumpTypeKey`

### LINKED ISSUE #211: Type checking for keys in cvdump `TYPES` parser
By "keys" I mean the identifiers for each type, as in:
```
0x1003 : Length = 26, Leaf = 0x1009 LF_MFUNCTION
	Return type = T_VOID(0003), Class type = 0x1000, This type = 0x1001, 
	Call type = ThisCall, Func attr = none
	Parms = 0, Arg list type = 0x1002, This adjust = 0
```
The keys in this excerpt are:
- `0x1003` (for this `MFUNCTION` leaf)
- `0x1000` class type
- `0x1001` this type
- `0x1002` arg list type
- and also `0x0003` indicating `void` return type

Right now the keys are stored as strings after a normalization step. This is needed because hex digits can be lower or upper case. We store the scalar `T_` types as a string also.

We would like to enforce type checking to make sure we are using valid keys. The current implementation does not normalize the keys in each occurrence. (Follow-up to #203)

We may also want to consider using an `int` instead. You can see the pattern for scalar type ids in [cvinfo.h](https://github.com/microsoft/microsoft-pdb/blob/master/include/cvinfo.h) and they should all be under `0x1000`. We could preload the type database with some of these scalar values and possibly resolve the type issues in #85.

disinvite:
Fixes #211.

This is a huge change and it could easily be split into a few smaller ones if needed.

## Intro

Type keys are used throughout the `TYPES` section of cvdump, along with other sections that refer to a type, such as `SYMBOLS` and `GLOBALS`. They come in three forms:

- Scalar types like `T_NOTYPE(0000)`
- "Unknown" scalar types like `???(047C)`
- User-defined types like `0x1234`

Until now we have used a string representation for this key, but the use of upper or lower-case hex digits is not consistent. We use the `normalize_type_id` function to handle this, but also manually convert using `str.lower()` in many spots. #203 added the `CvdumpTypeKey` annotation but this was just an alias for `str` so it was not checked with `mypy`.

The obvious improvement is to exploit the fact that scalar types occupy the range from `0x0000` to `0x1000`, so the key could just be an `int`. The obstacle there is that we depend on the string version of the `T_*` scalar types to decide the type's footprint and other attributes. [`cvinfo.h`](https://github.com/microsoft/microsoft-pdb/blob/master/include/cvinfo.h) contains bit-shifting macros that do this using integers. Considering that the range of scalar types is only 10% full (about 300 total) it is reasonable to use const data for this instead of calculating it on the fly.

## This change

- All scalar types now have `CvInfoType` tuples in a new file `cvinfo.py`, which includes the type's size and format chars for `struct.unpack`.

- `CvdumpTypeKey` is now a class that extends `int` so `mypy` will type check it. We need to make sure it is an `int` when storing in the Entity DB and convert back to `CvdumpTypeKey` when reading data back. For now, this happens automatically, but we could use conversion methods too.

- The parametrized tests in `test_cvdump.py` may not have much value now that we store the type attributes directly in `cvinfo.py`. One change to test data is that `float` and `double` are now considered unsigned to match `cvinfo.h`. The only impact was the choice of format char, but we now store this directly rather than computing it.

- On the topic of `struct.unpack` format chars, I decided to use `i/I` for the `INT` types and `l/L` for `LONG`. We had previously used `l/L` for any 4-byte integer. This is more of a style thing and I don't expect any practical difference.

- `CvdumpTypeKey.from_str()` replaces `normalize_type_id` and it's used everywhere we need to convert a type key. Similarly, the `is_scalar()` function replaces magic string checks for the `"T_"` prefix.

- Because `CvdumpTypeKey` is now a class, we need to use it to initialize any `int` variable that will serve as a type key. This change is most prevalent in unit tests.

- Similarly, whenever we need to refer to a specific type, there is an `Enum` based on `CvdumpTypeKey`. This is created dynamically with `types.new_class` to save space.

- The Ghidra scripts use the string representation of scalar types to decide which Ghidra primitive to use, and to add a `*` for pointers. We can access the `T_` string in the `CvInfoType` tuple so the conversion should be the same.

- I fixed a bug with `LF_ENUM` where forward refs were not followed. Forward refs have an underlying type of `T_NOTYPE(0000)` but this slipped by unnoticed because we assumed a size of 4 most types.

- Fixed another bug: we did not correctly handle type keys larger than `0xffff` because they use more characters than expected.

## Future

Some types might be handled incorrectly in the new system, but we won't know for sure until we see them in a PDB. I marked most scalar types with a `weird` attribute and the logger will display an info message if we see any.

It is probably time to separate cvdump text parsing from the general type database.

Not all data is exposed through the `get()` API and we need to access the `keys` dict directly. e.g. We don't store modifiers like `const` or `volatile` or the destination type of pointers. See #106.

`test_cvdump_types.py` is now over 1000 lines, so I added the `pylint:disable` option for now. The problem is that we need the huge blocks of text to set up the parser, but this could probably be broken out into tests for each leaf type.

### DIFF
--- a/reccmp/compare/asm/fixes.py
+++ b/reccmp/compare/asm/fixes.py
@@ -30,6 +30,19 @@ def jump_swap_ok(a: str, b: str) -> bool:
     return (jmp_a, jmp_b) in ALLOWED_JUMP_SWAPS
 
 
+def _mnemonic(inst: str) -> str:
+    if not inst:
+        return ""
+    return inst.split(" ", 1)[0].lower()
+
+
+def _split_operands(inst: str) -> list[str]:
+    (_, _, operand_str) = inst.partition(" ")
+    if not operand_str:
+        return []
+    return [operand.strip() for operand in operand_str.split(",") if operand.strip()]
+
+
 def is_operand_swap(a: str, b: str) -> bool:
     """This is a hack to avoid parsing the operands. It's not as simple as
     breaking on the comma because templates or string literals interfere
@@ -56,6 +69,16 @@ def get_patched_jump(a: str, b: str) -> str:
 
 
 def patch_mov_cmp_jmp(orig: list[str], recomp: list[str]) -> set[int]:
+    return patch_mov_compare_jmp(orig, recomp, "cmp")
+
+
+def patch_mov_test_jmp(orig: list[str], recomp: list[str]) -> set[int]:
+    return patch_mov_compare_jmp(orig, recomp, "test")
+
+
+def patch_mov_compare_jmp(
+    orig: list[str], recomp: list[str], cmp_instruction: str
+) -> set[int]:
     """Can we resolve the diffs between orig and recomp by patching
     swapped cmp instructions?
     For example:
@@ -66,15 +89,17 @@ def patch_mov_cmp_jmp(orig: list[str], recomp: list[str]) -> set[int]:
     Returns set of fixed lines
     """
 
-    # find the first "cmp" instruction
-    cmp_index = next((i for i, s in enumerate(orig) if s.startswith("cmp")), -1)
+    # find the first "cmp"/"test" instruction
+    cmp_index = next(
+        (i for i, s in enumerate(orig) if s.startswith(cmp_instruction)), -1
+    )
 
     # return if not found, or only found on first or last line
     if (
         cmp_index in (-1, 0, len(orig) - 1)
         or
         # recomp should also have a cmp in the same line
-        not recomp[cmp_index].startswith("cmp")
+        not recomp[cmp_index].startswith(cmp_instruction)
         or
         # line before cmp must be a mov
         not orig[cmp_index - 1].startswith("mov")
@@ -98,7 +123,72 @@ def patch_mov_cmp_jmp(orig: list[str], recomp: list[str]) -> set[int]:
     return set()
 
 
+def patch_mov_commutative(orig: list[str], recomp: list[str]) -> set[int]:
+    """Can we resolve the diffs between orig and recomp by patching
+        swapped operands in mov + commutative ops (add, and, or, xor, imul).
+    For example:
+        mov eax, dword ptr [ebp - 0x4]      mov eax, dword ptr [ebp - 0x8]
+        add eax, dword ptr [ebp - 0x8]      add eax, dword ptr [ebp - 0x4]
+
+    Returns set of fixed lines
+    """
+
+    valid_mnemonics = ("add", "and", "or", "xor", "imul")
+    inst_index = next(
+        (i for i, s in enumerate(orig) if _mnemonic(s) in valid_mnemonics), -1
+    )
+
+    # commutative op must exist and have a preceding line in both slices
+    if inst_index in (-1, 0) or inst_index >= len(recomp):
+        return set()
+
+    # this pattern only handles mov + {valid_mnemonics}
+    if (
+        _mnemonic(recomp[inst_index]) != _mnemonic(orig[inst_index])
+        or _mnemonic(orig[inst_index - 1]) != "mov"
+        or _mnemonic(recomp[inst_index - 1]) != "mov"
+    ):
+        return set()
+
+    orig_mov_ops = _split_operands(orig[inst_index - 1])
+    recomp_mov_ops = _split_operands(recomp[inst_index - 1])
+    orig_ops = _split_operands(orig[inst_index])
+    recomp_ops = _split_operands(recomp[inst_index])
+
+    if len(orig_mov_ops) < 2 or len(recomp_mov_ops) < 2:
+        return set()
+
+    # MOV destination must be the same register in both versions.
+    mov_dest_norm = orig_mov_ops[0].lower()
+    if mov_dest_norm != recomp_mov_ops[0].lower() or mov_dest_norm not in REGISTER_SET:
+        return set()
+
+    # Must target the same register and swap sources exactly.
+    op_layout_ok = (
+        len(orig_ops) == 2
+        and len(recomp_ops) == 2
+        and orig_ops[0].lower() == mov_dest_norm
+        and recomp_ops[0].lower() == mov_dest_norm
+    )
+    swap_ok = orig_ops[1] == recomp_mov_ops[1] and recomp_ops[1] == orig_mov_ops[1]
+
+    if op_layout_ok and swap_ok:
+        return {inst_index - 1, inst_index}
+
+    return set()
+
+
 def patch_cmp_jmp(orig: list[str], recomp: list[str]) -> set[int]:
+    return patch_compare_jmp(orig, recomp, "cmp")
+
+
+def patch_test_jmp(orig: list[str], recomp: list[str]) -> set[int]:
+    return patch_compare_jmp(orig, recomp, "test")
+
+
+def patch_compare_jmp(
+    orig: list[str], recomp: list[str], cmp_instruction: str
+) -> set[int]:
     """Can we resolve the diffs between orig and recomp by patching
     swapped cmp instructions?
     For example:
@@ -111,14 +201,16 @@ def patch_cmp_jmp(orig: list[str], recomp: list[str]) -> set[int]:
     Returns set of fixed lines
     """
 
-    # find the first "cmp" instruction
-    cmp_index = next((i for i, s in enumerate(orig) if s.startswith("cmp")), -1)
+    # find the first "cmp"/"test" instruction
+    cmp_index = next(
+        (i for i, s in enumerate(orig) if s.startswith(cmp_instruction)), -1
+    )
     # return if not found, or only found on the last line
     if (
         cmp_index in (-1, len(orig) - 1)
         or
         # recomp should also have a cmp in the same line
-        not recomp[cmp_index].startswith("cmp")
+        not recomp[cmp_index].startswith(cmp_instruction)
         or
         # if the last lines are not a compatible jump difference
         not jump_swap_ok(orig[cmp_index + 1], recomp[cmp_index + 1])
@@ -148,7 +240,7 @@ def patch_fld_fmul(orig: list[str], recomp: list[str]) -> set[int]:
 
     valid_following_ops = ["fmul", "fadd"]
 
-    # find the first "cmp" instruction
+    # find the first "fld" instruction
     fld_index = next((i for i, s in enumerate(orig) if s.startswith("fld")), -1)
     # return if not found, or only found on the last line
     if (
@@ -190,7 +282,14 @@ def patch_cmp_swaps(
 
     fixed_lines = set()
 
-    patch_fns = [patch_cmp_jmp, patch_mov_cmp_jmp, patch_fld_fmul]
+    patch_fns = [
+        patch_cmp_jmp,
+        patch_test_jmp,
+        patch_mov_cmp_jmp,
+        patch_mov_test_jmp,
+        patch_fld_fmul,
+        patch_mov_commutative,
+    ]
 
     for code, i1, i2, j1, j2 in codes:
         # To save us the trouble of finding "compatible" cmp instructions
@@ -355,6 +454,7 @@ def relocate_instructions(
 DWORD_REGS = ("eax", "ebx", "ecx", "edx", "esi", "edi", "ebp", "esp")
 WORD_REGS = ("ax", "bx", "cx", "dx", "si", "di", "bp", "sp")
 BYTE_REGS = ("ah", "al", "bh", "bl", "ch", "cl", "dh", "dl")
+REGISTER_SET = set(reg for reg in (DWORD_REGS + WORD_REGS + BYTE_REGS))
 
 
 def naive_register_replacement(orig_asm: list[str], recomp_asm: list[str]) -> set[int]:
--- a/tests/test_fixes.py
+++ b/tests/test_fixes.py
@@ -13,6 +13,16 @@ def test_fix_cmp_jmp():
     assert is_effective is True
 
 
+def test_fix_test_jmp():
+    orig_asm = ["mov eax, 1", "mov ebx, 2", "test eax, ebx", "jg 0x1"]
+    recomp_asm = ["mov eax, 1", "mov ebx, 2", "test ebx, eax", "jl 0x1"]
+
+    diff = difflib.SequenceMatcher(None, orig_asm, recomp_asm)
+    is_effective = find_effective_match(diff.get_opcodes(), orig_asm, recomp_asm)
+
+    assert is_effective is True
+
+
 def test_fix_mov_cmp_jmp_mem_with_different_operands():
     """This should not be fixed up, since the operands are different"""
     orig_asm = [
@@ -80,6 +90,25 @@ def test_fix_mov_cmp_jmp_mem_valid():
     recomp_asm = [
         "mov eax, dword ptr [gCurrent_key (DATA)]",
         "cmp dword ptr [ebp-4], eax",
+        "jne 0x1",
+    ]
+
+    diff = difflib.SequenceMatcher(None, orig_asm, recomp_asm)
+    is_effective = find_effective_match(diff.get_opcodes(), orig_asm, recomp_asm)
+
+    assert is_effective is True
+
+
+def test_fix_mov_test_jmp_mem_valid():
+
+    orig_asm = [
+        "mov eax, dword ptr [ebp-4]",
+        "test dword ptr [gCurrent_key (DATA)], eax",
+        "jne 0x1",
+    ]
+    recomp_asm = [
+        "mov eax, dword ptr [gCurrent_key (DATA)]",
+        "test dword ptr [ebp-4], eax",
         "jne 0x1",
     ]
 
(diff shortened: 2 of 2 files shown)


reccmp/cvdump/cvinfo.py
```diff
@@ -0,0 +1,518 @@
...
+    """The integer type key."""
+    name: str
+    """The name of this type as given in cvinfo.h."""
+    fmt: str
+    """The struct.unpack format char(s) for this type."""
+    size: int
+    """The type's footprint in bytes."""
+    pointer: CvdumpTypeKey | None
+    """If set, this type is a pointer to another CVinfo type."""
+    weird: bool
```
jonschz: Gave me a chuckle :D how about `never_seen`, or something along these lines? Or the inverse boolean `validated`, `known_to_exist`, etc.?

reccmp/cvdump/cvinfo.py
```diff
@@ -0,0 +1,518 @@
...
+
+class CvInfoType(NamedTuple):
+    key: CvdumpTypeKey
+    """The integer type key."""
+    name: str
+    """The name of this type as given in cvinfo.h."""
+    fmt: str
+    """The struct.unpack format char(s) for this type."""
+    size: int
+    """The type's footprint in bytes."""
```
jonschz, line 37: I'd probably leave that as-is, but could there be an issue with bitfields and fractions of a byte? I don't know how they are represented in cvdump.

reccmp/cvdump/cvinfo.py
```diff
@@ -0,0 +1,518 @@
...
+from types import MappingProxyType, new_class
+from typing import NamedTuple
+
+
+class CvdumpTypeKey(int):
+    def is_scalar(self) -> bool:
+        return self < 0x1000
+
+    @classmethod
+    def from_str(cls, key: str) -> "CvdumpTypeKey":
```
jonschz, line 16:
```suggestion
    def from_str(cls, key: str) -> Self:
```
and `from type_extensions import Self`

reccmp/cvdump/cvinfo.py
```diff
@@ -0,0 +1,518 @@
...
+def scalar_type_pointer(key: CvdumpTypeKey) -> bool:
+    return get_cvinfo(key).pointer is not None
+
+
+def scalar_type_size(key: CvdumpTypeKey) -> int:
+    return get_cvinfo(key).size
+
+
+def scalar_type_signed(key: CvdumpTypeKey) -> bool:
+    return key in (
```
jonschz: Maybe add the sign info to the `CvInfoType`?

reccmp/cvdump/cvinfo.py
```diff
@@ -0,0 +1,518 @@
...
+# fmt: on
+
+
+_CVInfoTypeEnum = new_class("_CVInfoTypeEnum", bases=(CvdumpTypeKey, Enum))
+CVInfoTypeEnum = _CVInfoTypeEnum(
+    "CVInfoTypeEnum", {cv.name: cv.key for cv in _CVINFO_TYPES}
+)
+
+
+_TYPE_ENUM_E = MappingProxyType({cv.key: cv for cv in _CVINFO_TYPES})
```
jonschz: I'm quite impressed that it works this way, but I got nerd-sniped and may have found a better solution. My issue is that this isn't type safe. The following passes `mypy` and the IDE doesn't show anything either:
```py
_CVInfoTypeEnum = new_class("_CVInfoTypeEnum", bases=(CvdumpTypeKey, Enum))
CVInfoTypeEnum = _CVInfoTypeEnum(
    "CVInfoTypeEnum", {cv.name: cv.key for cv in _CVINFO_TYPES}
)

CVInfoTypeEnum.ABCDE
```

The best I could come up with is something along the following lines:
```py
CVInfoTypeEnum = Enum(
    "CVInfoTypeEnum",
    [
        ("T_NOTYPE", CvInfoType(key=CvdumpTypeKey(0x0000), fmt="",     size=0,   pointer=None,                   weird=False)), # uncharacterized type (no type)
# ...
    ]
)

_TYPE_ENUM_E = MappingProxyType({ x.value.key: x.value for x in CVInfoTypeEnum })
_NAME_FOR_TYPE = MappingProxyType({ x.value.key: x.name for x in CVInfoTypeEnum })

# ...

def cvinfo_type_name(key: CvdumpTypeKey) -> str:
    assert key.is_scalar()
    return _NAME_FOR_TYPE[key]
```
replacing lines 40-483. This requires a few minor changes elsewhere, e.g.
```py
            return self.get(CVInfoTypeEnum.T_64NCVPTR)
# to
            return self.get(CVInfoTypeEnum.T_64NCVPTR.value.key)
```
and one instance whe `.name` needs to be replaced by `cvinfo_type_name`, but I think it may be worth it overall.

reccmp/cvdump/cvinfo.py
```diff
@@ -0,0 +1,518 @@
...
+class CvdumpTypeKey(int):
+    def is_scalar(self) -> bool:
+        return self < 0x1000
+
+    @classmethod
+    def from_str(cls, key: str) -> "CvdumpTypeKey":
+        if key[0] == "0":
+            return cls(int(key, 16))
+
+        # Should cover both "T_" and "???" cases.
```
jonschz: Could there be more than `0x9999` types?

reccmp/cvdump/cvinfo.py
```diff
@@ -0,0 +1,518 @@
...
+    CvInfoType(key=CvdumpTypeKey(0x0120),  name="T_PUCHAR",       fmt="H",    size=2,   pointer=CvdumpTypeKey(0x0020),  weird=True ), # 16 bit pointer to 8 bit unsigned
+    CvInfoType(key=CvdumpTypeKey(0x0220),  name="T_PFUCHAR",      fmt="I",    size=4,   pointer=CvdumpTypeKey(0x0020),  weird=True ), # 16:16 far pointer to 8 bit unsigned
+    CvInfoType(key=CvdumpTypeKey(0x0320),  name="T_PHUCHAR",      fmt="I",    size=4,   pointer=CvdumpTypeKey(0x0020),  weird=True ), # 16:16 huge pointer to 8 bit unsigned
+    CvInfoType(key=CvdumpTypeKey(0x0420),  name="T_32PUCHAR",     fmt="I",    size=4,   pointer=CvdumpTypeKey(0x0020),  weird=False), # 32 bit pointer to 8 bit unsigned
+    CvInfoType(key=CvdumpTypeKey(0x0520),  name="T_32PFUCHAR",    fmt="6x",   size=6,   pointer=CvdumpTypeKey(0x0020),  weird=True ), # 16:32 pointer to 8 bit unsigned
+    CvInfoType(key=CvdumpTypeKey(0x0620),  name="T_64PUCHAR",     fmt="Q",    size=8,   pointer=CvdumpTypeKey(0x0020),  weird=True ), # 64 bit pointer to 8 bit unsigned
+
+
+#      really a character types
```
jonschz:
```suggestion
#      "really a character types" (according to `cvinfo.h`)
#      Note that cvdump does not understand these and prints them e.g. like `???(007C)`
```
I wondered what you meant by the comment until I looked into the source

reccmp/cvdump/types.py
```diff
@@ -58,22 +55,13 @@ class ScalarType(NamedTuple):
...
-    @property
-    def format_char(self) -> str:
-        return scalar_type_format_char(self.type)
-
-    @property
-    def is_pointer(self) -> bool:
-        return scalar_type_pointer(self.type)
+    size: int
+    format_char: str
+    is_pointer: bool
```
jonschz: You could also include the `CvdumpTypeInfo` here instead of copying a subset of the fields under different names

reccmp/cvdump/cvinfo.py
```diff
@@ -0,0 +1,518 @@
...
+        # Should cover both "T_" and "???" cases.
+        return cls(int(key[-5:-1], 16))
+
+
+class CvInfoType(NamedTuple):
+    key: CvdumpTypeKey
+    """The integer type key."""
+    name: str
+    """The name of this type as given in cvinfo.h."""
+    fmt: str
```
jonschz, line 34:
```suggestion
    struct_unpack_fmt: str
```
maybe? More self-documenting

reccmp/ghidra/importer/pdb_extraction.py
```diff
@@ -57,30 +57,26 @@ class PdbFunctionExtractor:
...
-            if type_name is None
-            else self.compare.types.keys.get(type_name.lower())
-        )
+    def _get_cvdump_type(
+        self, type_name: CvdumpTypeKey | None
+    ) -> CvdumpParsedType | None:
+        return None if type_name is None else self.compare.types.keys.get(type_name)
 
     def get_func_signature(self, fn: SymbolsEntry) -> FunctionSignature | None:
         function_type_str = fn.func_type
```
jonschz:
```suggestion
        function_type_key = fn.func_type
```

reccmp/ghidra/importer/type_importer.py
```diff
@@ -93,24 +95,21 @@ def import_pdb_type_into_ghidra(
...
+        if type_index.is_scalar():
+            return self._import_scalar_type(type_index)
 
         try:
-            type_pdb = self.extraction.compare.types.keys[type_index_lower]
+            type_pdb = self.extraction.compare.types.keys[type_index]
         except KeyError as e:
             raise TypeNotFoundError(
-                f"Failed to find referenced type '{type_index_lower}'"
+                f"Failed to find referenced type '{type_index:#x}'"
```
jonschz: TODO for myself: Check how this looks

reccmp/ghidra/importer/type_importer.py
```diff
@@ -149,16 +148,19 @@ def _scalar_type_to_cpp(self, scalar_type: str) -> str:
...
-            raise TypeNotFoundError(f"Type has unexpected format: {type_index_lower}")
+    def _import_scalar_type(self, type_key: CvdumpTypeKey) -> DataType:
+        if not type_key.is_scalar():
+            raise TypeNotFoundError(f"Type has unexpected format: {type_key:#x}")
 
-        scalar_cpp_type = self._scalar_type_to_cpp(match.group("typename"))
+        # Remove "T_" prefix and convert to lower-case
+        # to match previous processing.
+        type_name = cvinfo_type_name(type_key)[2:].lower()
+        scalar_cpp_type = self._scalar_type_to_cpp(type_name)
```
jonschz: This looks a bit off here. Should it be moved into `_scalar_type_cpp()` an be covered by tests there?

tests/test_cvdump_types.py
```diff
@@ -371,253 +379,265 @@ def types_empty_parser_fixture():
...
 def test_struct_format_string(parser: CvdumpTypesParser):
     """Generate the struct.unpack format string using the
     list of scalars with padding filled in."""
     # MxString, padded to 16 bytes.
-    assert parser.get_format_string("0x4db6") == "<LLLHBB"
+    assert parser.get_format_string(TK(0x4DB6)) == "<IIIHBB"
 
     # MxVariable, with two MxString members.
-    assert parser.get_format_string("0x22d5") == "<LLLLHBBLLLHBB"
+    assert parser.get_format_string(TK(0x22D5)) == "<IIIIHBBIIIHBB"
```
jonschz, line 514: fine (why does Python even distinguish between `int` and `long` in the first place?)

reccmp/cvdump/cvinfo.py
jonschz:
```suggestion
        # For types like `T_ULONG(0022)` and `???(047C)`
```
maybe?

reccmp/cvdump/cvinfo.py
```diff
@@ -0,0 +1,518 @@
...
+    CvInfoType(key=CvdumpTypeKey(0x017b),  name="T_PCHAR32",      fmt="H",    size=2,   pointer=CvdumpTypeKey(0x007b),  weird=True ), # 16 bit pointer to a 32-bit unicode char
+    CvInfoType(key=CvdumpTypeKey(0x027b),  name="T_PFCHAR32",     fmt="I",    size=4,   pointer=CvdumpTypeKey(0x007b),  weird=True ), # 16:16 far pointer to a 32-bit unicode char
+    CvInfoType(key=CvdumpTypeKey(0x037b),  name="T_PHCHAR32",     fmt="I",    size=4,   pointer=CvdumpTypeKey(0x007b),  weird=True ), # 16:16 huge pointer to a 32-bit unicode char
+    CvInfoType(key=CvdumpTypeKey(0x047b),  name="T_32PCHAR32",    fmt="I",    size=4,   pointer=CvdumpTypeKey(0x007b),  weird=True ), # 32 bit pointer to a 32-bit unicode char
+    CvInfoType(key=CvdumpTypeKey(0x057b),  name="T_32PFCHAR32",   fmt="6x",   size=6,   pointer=CvdumpTypeKey(0x007b),  weird=True ), # 16:32 pointer to a 32-bit unicode char
+    CvInfoType(key=CvdumpTypeKey(0x067b),  name="T_64PCHAR32",    fmt="Q",    size=8,   pointer=CvdumpTypeKey(0x007b),  weird=True ), # 64 bit pointer to a 32-bit unicode char
+
+
+#      8-bit unicode char
```
jonschz:
```suggestion
#      8-bit unicode char
#      Note that MSVC appears to use `T_RCHAR` and related types instead
```

reccmp/cvdump/cvinfo.py
```diff
@@ -0,0 +1,518 @@
...
+    CvInfoType(key=CvdumpTypeKey(0x0033),  name="T_BOOL64",       fmt="Q",    size=8,   pointer=None,                   weird=True ), # 64 bit boolean
+    CvInfoType(key=CvdumpTypeKey(0x0133),  name="T_PBOOL64",      fmt="H",    size=2,   pointer=CvdumpTypeKey(0x0033),  weird=True ), # 16 bit pointer to 64 bit boolean
+    CvInfoType(key=CvdumpTypeKey(0x0233),  name="T_PFBOOL64",     fmt="I",    size=4,   pointer=CvdumpTypeKey(0x0033),  weird=True ), # 16:16 far pointer to 64 bit boolean
+    CvInfoType(key=CvdumpTypeKey(0x0333),  name="T_PHBOOL64",     fmt="I",    size=4,   pointer=CvdumpTypeKey(0x0033),  weird=True ), # 16:16 huge pointer to 64 bit boolean
+    CvInfoType(key=CvdumpTypeKey(0x0433),  name="T_32PBOOL64",    fmt="I",    size=4,   pointer=CvdumpTypeKey(0x0033),  weird=True ), # 32 bit pointer to 64 bit boolean
+    CvInfoType(key=CvdumpTypeKey(0x0533),  name="T_32PFBOOL64",   fmt="6x",   size=6,   pointer=CvdumpTypeKey(0x0033),  weird=True ), # 16:32 pointer to 64 bit boolean
+    CvInfoType(key=CvdumpTypeKey(0x0633),  name="T_64PBOOL64",    fmt="Q",    size=8,   pointer=CvdumpTypeKey(0x0033),  weird=True ), # 64 bit pointer to 64 bit boolean
+
+
+#      ???
```
jonschz:
```suggestion
#      "???" (according to cvinfo.h)
#      These appear to be internal types.
```

reccmp/cvdump/cvinfo.py
```diff
@@ -0,0 +1,518 @@
...
+)
+
+
+_TYPE_ENUM_E = MappingProxyType({cv.key: cv for cv in _CVINFO_TYPES})
+
+
+def get_cvinfo(key: CvdumpTypeKey) -> CvInfoType:
+    assert key.is_scalar()
+    return _TYPE_ENUM_E.get(
+        key, CvInfoType(key=key, name="???", fmt="", size=0, pointer=None, weird=True)
```
jonschz: I'd rather `raise` here. Having a wrong type (e.g. wrong size, no format string) could lead to unexpected/undefined behaviour elsewhere in our code base. I briefly changed it and had no issues on LEGO1 or BETA10.

jonschz (approved):
Looks very good, this is a big improvement to the code quality!

reccmp/cvdump/cvinfo.py
disinvite: What about declaring the enum explicitly in the file?
```python
class CVInfoTypeEnum(CvdumpTypeKey, Enum):
    T_NOTYPE = CvdumpTypeKey(0x0000)
    T_ABS = CvdumpTypeKey(0x0001)
    T_SEGMENT = CvdumpTypeKey(0x0002)
    # ...
```
And then verify that they match with a test like:
```python
def verify_cvinfo_enum():
    for e in CVInfoTypeEnum:
        assert isinstance(e, CvdumpTypeKey)
        assert isinstance(e, int)
        assert get_cvinfo(e).name == e.key

    for t in _CVINFO_TYPES:
        assert CVInfoTypeEnum[t.name] == t.key
```
`mypy` alerts for invalid keys like `CVInfoTypeEnum.ABCD`.

reccmp/cvdump/cvinfo.py
disinvite: Yeah, that sounds good.

Since we're using `MappingProxyType` which is effectively an immutable `dict`, what do you think about exposing it directly in the module instead of wrapping with `get_cvinfo()`? Callers could trap `KeyError` and handle it or not. In `types.py` we can do our logging and then re-throw `CvdumpKeyError`.

reccmp/cvdump/cvinfo.py
disinvite: I don't think we'll know for sure until we see `T_BIT` in the wild. I considered adding a `special` attribute to `CvInfoType` for types like `T_PASCHAR` where the handling will not be so simple, but I took it out because we don't have any of that handling yet.

`LF_BITFIELD` is doable if we can tell how big the footprint (with padding) should be. #299

reccmp/cvdump/cvinfo.py
disinvite: Created #316 for general cleanup outside this PR

reccmp/ghidra/importer/type_importer.py
disinvite: Just added this. Let me know what you think.

reccmp/cvdump/cvinfo.py
jonschz: I got some warnings, but maybe I did something wrong. If you can make it work, that's better still than my solution

reccmp/cvdump/cvinfo.py
jonschz: Exposing directly sounds good as well

reccmp/ghidra/importer/type_importer.py
jonschz: I like it!

tests/test_cvdump_types.py
disinvite: This could be why:
> When using native size, the size of the packed value is platform-dependent.

https://docs.python.org/3/library/struct.html#format-characters

reccmp/cvdump/cvinfo.py
disinvite: I think I'd rather just drop it as a test since the signed-ness only informs how we unpack the value. I have some tests ready for `datacmp` that check the final string version of a variable with various types (including signed/unsigned) that use the same `bytes` value.

jonschz: fine with me. I just didn't like this separate list

PR #317, opened 2026-02-13
Variable comparator with tests

### LINKED ISSUE #303: Move variable comparison into its own module
Similar to the function comparator, we should have a module that takes in entities and returns the compared result. This logic is in the `datacmp` script right now and there aren't any unit tests.

disinvite:
Closes #303.

Moves the comparison logic for variables out of `tools/datacmp.py` and into a new variable comparator class, similar to the existing one for functions. This allows us to drag in specific dependencies from the main `Compare` object so they can be mocked in tests. The `is_pointer_match` function in `core.py` was only used by `datacmp`, so I moved it here.

`MockTypesDb` allows us to define complex types without reading cvdump text, as we've done in other tests.

I also set up a file for the existing `RawImage` class because it is now used in two test modules. This allows us to define the initialized and uninitialized data for an image  without having to create a complex mock on the `read()` function.

My main focus was testing primitive type variables in different scenarios. Since `get_scalars_gapless()` function reduces the complex type to a list of these scalar values, I felt that this wasn't that different from comparing a single value.


tests/raw_image.py
```diff
@@ -0,0 +1,36 @@
...
+import dataclasses
+from pathlib import Path
+from reccmp.formats import Image
+from reccmp.formats.exceptions import InvalidVirtualAddressError
+
+
+# pylint: disable=abstract-method
+@dataclasses.dataclass
+class RawImage(Image):
+    """For testing functions implemented in the base Image class."""
```
jonschz: This docstring isn't true anymore, and I'd rather focus on what it does rather than what it's used for

tests/test_variable_comparator.py
```diff
@@ -0,0 +1,456 @@
...
+    """Match on scalar variable with initialized data."""
+    create_matched_variable(db, 0, data_type=CVInfoTypeEnum.T_32PVOID)
+    # Entity at address 0x0004 required to match
+    with db.batch() as batch:
+        batch.set_recomp(4, name="Hello")
+        batch.match(4, 4)
+
+    orig = RawImage.from_memory(b"\x04\x00\x00\x00")
+    recomp = RawImage.from_memory(b"\x04\x00\x00\x00")
+    comparator = VariableComparator(db, types, orig, recomp)
```
jonschz, line 91: The test could be even more expressive by having different addresses for the addresses the pointers point to. Applies to the test below as well.

tests/raw_image.py
```diff
@@ -0,0 +1,36 @@
...
+
+    # Total size of the image.
+    # If it is more than the size of physical data, the remainder is uninitialized (all null).
```
jonschz: Read the comment below first, it may make this comment obsolete.

I'd move/copy that into a new docstring for `from_memory`, and emphasise that `size` may be larger than `data`.

tests/raw_image.py
```diff
@@ -0,0 +1,36 @@
...
+    size: int
+
+    @classmethod
+    def from_memory(cls, data: bytes = b"", *, size: int = 0) -> "RawImage":
+        if size is None:
+            maxsize = len(data)
+        else:
+            maxsize = max(size, len(data))
```
jonschz: What do you think about replacing `size` by `additional_uninitialized_bytes`, which must be `>= 0`, default `0`. This may also make the tests below more self-explanatory.

In case you want to keep `size`: I'd rather raise an error here if `size < len(data)`. This is most likely unintentional and the test author may be suprised by this behaviour.

tests/test_variable_comparator.py
```diff
@@ -0,0 +1,456 @@
...
+
+    c = comparator.compare_variable(get_match(db, 0))
+
+    assert c is not None
+    assert c.result == CompareResult.DIFF
+    assert c.compared[0].match is True
+    assert c.compared[1].match is False
+
+
+def test_compare_complex_with_padding(db: EntityDb):
```
jonschz: Nice, I didn't even know this was supported already!

Should we also test padding in the middle?

tests/test_variable_comparator.py
```diff
@@ -0,0 +1,456 @@
...
+    recomp = RawImage.from_memory(b"\x01\x02\x03\x04\x05\x06\x00\x00")
+    comparator = VariableComparator(db, types, orig, recomp)
+
+    c = comparator.compare_variable(get_match(db, 0))
+
+    assert c is not None
+    assert c.result == CompareResult.DIFF
+
+
+@pytest.mark.xfail(reason="GH #305")
```
jonschz, line 352: I like this pattern of linking xfails to GitHub issues. Maybe the issue should reference its xfails tests as well, so we don't forget to re-enable them before closing the ticket.

tests/test_variable_comparator.py
```diff
@@ -0,0 +1,456 @@
...
+    recomp = RawImage.from_memory(b"\x0a\x00\x00\x00", size=16)
+    comparator = VariableComparator(db, types, orig, recomp)
+
+    c = comparator.compare_variable(get_match(db, 0))
+
+    assert c is not None
+    assert c.result == CompareResult.MATCH
+
+
+def test_compare_complex_raw_missing_key(db: EntityDb, types: CvdumpTypesParser):
```
jonschz, line 405: should this have a mismatch test as well?

tests/raw_image.py
disinvite: I like it, thanks! I only had to change a couple things around in `test_image_raw.py` to get it working.

What do you think of `bss` as the parameter name? We're already using that abbreviation for uninitialized data in `variables.py`. Since it's required to be a kwarg I think it should be short if possible.

jonschz: `bss_size` maybe? `bss` is fine as well

PR #321, opened 2026-02-14
Fix `IndexError` raised for single operand `IMUL` instruction

disinvite:
A group of instructions like this breaks the recent `mov/imul` swap correction in #314:
```asm
mov ax, word ptr [bp + 0x38]
imul word ptr [bp + 0x3a]
```
This fix aborts the attempted fix if we get a single-operand `imul`.


jonschz (approved):
Good to merge, assuming you have checked that the test fails on the old code.

PR #323, opened 2026-02-14
Fix: Did not check all intermediate instructions when relocating

### LINKED ISSUE #322: Moving `and` instruction creates bad effective match
See dethrace-labs/dethrace#558

We assume incorrectly that `and eax, 0xff` can be moved between these two spots:
```diff
mov eax, dword ptr [ebp - 4] (powerup.c:898)
-and eax, 0xff
mov ecx, dword ptr [gReal_render_palette (DATA)]
mov eax, dword ptr [ecx + eax*4]
+and eax, 0xff
mov ecx, dword ptr [gRender_palette (DATA)]
```

disinvite:
Fixes #322. For each instruction we want to move, we check whether its registers are used in any of the intermediate instructions. The problem is that we didn't check all of them.

Each diff opcode has indices `i1, i2, j1, j2` for the two text blocks `A` and `B`. `A[i1 : i2]` and `B[j1 : j2]` are the ranges of lines affected by each alteration. The bug is that we used `j1` from the `insert` opcode in `B` as an index for `A`, when we should use `i1` instead.

For `CARM95` this only impacts the function from the original issue, which is what we want.

`LEGO1` had two new diffs. I fixed `LegoCarBuild::Escape` but not `SpheresIntersect`. The second function goes to 100% in entropy builds, though.

### DIFF
--- a/reccmp/compare/asm/fixes.py
+++ b/reccmp/compare/asm/fixes.py
@@ -418,10 +418,8 @@ def relocate_instructions(
     into recomp, according to the diff opcodes. Using this list, match up
     any pairs of instructions that we assume to be relocated and return
     the indices in recomp where this has occurred.
-    For now, we are checking only for an exact match on the instruction.
-    We are not checking whether the given instruction can be moved from
-    point A to B. (i.e. does this set a register that is used by the
-    instructions between A and B?)"""
+    This function has many limitations and could be improved. See: GH #324.
+    """
     deletes = {
         i for code, i1, i2, _, __ in codes for i in range(i1, i2) if code == "delete"
     }
@@ -445,8 +443,6 @@ def relocate_instructions(
         recomp_regs_used = set(find_regs_used(line))
         for i in deletes:
             # Check for exact match.
-            # TODO: This will grab the first instruction that matches.
-            # We should probably use the nearest index instead, if it matters
             if orig_asm[i] == line:
                 # To account for a move in either direction:
                 # the deleted line can precede or follow the inserted line.


jonschz:
Are there any other regressions on `isle`? If so, please don't forget to open a ticket listing all of them

reccmp/compare/asm/fixes.py
```diff
@@ -441,9 +448,11 @@ def relocate_instructions(
             # TODO: This will grab the first instruction that matches.
```
jonschz: Now that I notice it: Is this TODO still relevant?

disinvite:
> Are there any other regressions on `isle`? If so, please don't forget to open a ticket listing all of them

The only one is `SpheresIntersect` across `BETA10, CONFIG, LEGO1, ISLE`.

For posterity, the effective match fails now because we consider each of the swaps as individual lines:
```diff
mov esi, dword ptr [eax + 4]
+fld dword ptr [esi + 4]
+mov edx, dword ptr [ecx + 4]
+fsub dword ptr [edx + 4]
fld dword ptr [esi + 8]
-mov edx, dword ptr [ecx + 4]
fsub dword ptr [edx + 8]
-fld dword ptr [esi + 4]
-fsub dword ptr [edx + 4]
```
For each added (in recomp) line, we check lines from orig to see if its registers are changed:
```
Want to move `fld dword ptr [esi + 4]`
  fld dword ptr [esi + 8]
  mov edx, dword ptr [ecx + 4]
  fsub dword ptr [edx + 8]

Want to move `mov edx, dword ptr [ecx + 4]`
  fld dword ptr [esi + 8]

Want to move `fsub dword ptr [edx + 4]`
  fld dword ptr [esi + 8]
  mov edx, dword ptr [ecx + 4]    <-- Conflict, modifies edx
  fsub dword ptr [edx + 8]
  fld dword ptr [esi + 4]
```
Certainly there is room for improvement here, both in sophistication and guardrails to block false positives. For example, we don't check the FPU stack pointer, and we don't check EBP as modified by push and pop instructions.

reccmp/compare/asm/fixes.py
disinvite: Relevant in that this describes the current behavior, but it is far from the only problem. I put some ideas down in #324.

disinvite: Actually I just realized that since we use a `set()` for `deletes`, there is no guarantee of ordering, so this line is truly irrelevant 😬

PR #327, opened 2026-02-28
Set up Ghidra integration tests

jonschz:
At long last, there are some integration tests for Ghidra. The setup is still very basic, I am planning to build on it in future PRs.

For the binary I used a recompiled `ISLE.EXE`, though the test I wrote so far does not depend on the binary. We can still change that, it was the first thing to come to my mind.

I couldn't get the setup for the original binaries to work in GitHub actions, so I disabled the tests for now.

~~I also plan to do transactions and rollback in the fixture once I have tests that actually make meaningful modifications.~~ Edit: done

Furthermore, I refactored importing scalars, which made one exception type obsolete.

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -36,7 +36,7 @@ def pytest_addoption(parser: Parser):
     parser.addoption(
         REQUIRE_GHIDRA_OPTION,
         action="store_true",
-        help="Fail tests that depend on Ghidra it is not available.",
+        help="Fail tests that depend on Ghidra if it is not available.",
     )
 
 
@@ -137,9 +137,9 @@ def fixture_ghidra_program(ghidra_program: "Program") -> "Iterator[FlatProgramAP
     assert ghidra_program.getCurrentTransactionInfo() is None
 
     transaction = ghidra_program.openTransaction("reccmp-integration-test")
-    api = FlatProgramAPI(ghidra_program)
-
-    yield api
-
-    # Revert all side effects of the test that just ran
-    transaction.abort()
+    try:
+        api = FlatProgramAPI(ghidra_program)
+        yield api
+    finally:
+        # Revert all side effects of the test that just ran
+        transaction.abort()


reccmp/cvdump/cvinfo.py
```diff
@@ -46,7 +46,7 @@ class CvInfoType(NamedTuple):
 # The original description for each type is on the right, after the tuple.
 
 # fmt: off
-_CVINFO_TYPES = (
+_CVINFO_TYPES: tuple[CvInfoType, ...] = (
```
jonschz, line 49: drive-by: Improves IDE experience

reccmp/ghidra/importer/ghidra_helper.py
```diff
@@ -35,15 +34,15 @@ def get_scalar_ghidra_type(api: FlatProgramAPI, type_name: str) -> DataType:
     Note that this function may raise errors when a type by that name exists multiple times.
     Manual cleanup is needed in that case.
     """
+    category = (
+        api.getCurrentProgram().getDataTypeManager().getCategory(CategoryPath("/"))
```
jonschz: This only finds Ghidra's builtins, which removes the need for handling duplicates

jonschz:
> I guess we could do that. I wasn't thinking about the performance angle. It would all depend on how long it takes to analyze once, and that may be slow enough. 

I'll do a benchmark to have a basis for the decision. Will move the PR back to draft until then.

> The alternative would be to mock the communication with Ghidra's API and not use a real project at all. How feasible is that?

I specifically wanted a real Ghidra instance for the tests because there is a lot of code handling obscure edge cases, which are not obvious from the API at all.

jonschz:
I built a proof-of-concept and ran a few benchmarks:

- Ghidra startup (always required): ~ 3.5 seconds
- Loading an existing project: ~ 2.5 seconds
- Creating a new project, importing a binary (without analysis), and saving: ~ 7 seconds
- Analyzing `ISLE.EXE`: ~ 13 seconds

So a non-cached startup takes ~24 seconds, while a cached startup takes ~ 6 seconds. I'll go for the cache approach. I haven't yet looked at the Git cache download issue again, though.

jonschz:
I sort of got the cache to work. According to https://github.com/actions/cache/issues/1361#issuecomment-2611294058, the cache operation does not share caches across runners with different containers, so I had to set up a second cache for the Ghidra runs. Still needs some polish before it can be reviewed.

reccmp/ghidra/importer/type_conversion.py
```diff
@@ -1,25 +1,55 @@
...
+    CVInfoTypeEnum.T_USHORT: UnsignedShortDataType(),
+    CVInfoTypeEnum.T_ULONG: UnsignedLongDataType(),
+    CVInfoTypeEnum.T_UQUAD: UnsignedLongLongDataType(),
+    CVInfoTypeEnum.T_REAL32: FloatDataType(),
+    CVInfoTypeEnum.T_REAL64: DoubleDataType(),
+    CVInfoTypeEnum.T_RCHAR: CharDataType(),
+    CVInfoTypeEnum.T_WCHAR: WideCharDataType(),
+    CVInfoTypeEnum.T_INT4: IntegerDataType(),
+    CVInfoTypeEnum.T_UINT4: UnsignedIntegerDataType(),
+}
```
jonschz, line 44: I refactored these again, now they also work regardless of whether an analysis has been performed

reccmp/ghidra/ghidra_import.py
```diff
@@ -70,19 +69,13 @@ def main():
             if not target.ghidra_config.allow_hash_mismatch:
                 return 1
 
-        # Not exactly sure why this is necessary, but it can't hurt
-        GhidraScriptUtil.acquireBundleHostReference()
-
```
jonschz, line 75: Only required to run an analysis

requirements-tests.txt
```diff
@@ -3,4 +3,4 @@ black==25.1.0
 pylint==3.2.7
 mypy==1.16.1
 types-colorama>=0.4.6
-ghidra-stubs==11.4
+ghidra-stubs==12.0.2
```
jonschz, line 6: Running the tests locally requires Ghidra 12. The import still works with Ghidra 11.4

tests/conftest.py
```diff
@@ -1,19 +1,43 @@
...
     parser.addoption(
-        "--require-binfiles",
+        REQUIRE_BINFILES_OPTION,
         action="store_true",
         help="Fail tests that depend on binary samples if we cannot load them.",
     )
+    parser.addoption(
+        REQUIRE_GHIDRA_OPTION,
+        action="store_true",
+        help="Fail tests that depend on Ghidra it is not available.",
```
disinvite: _if_ it is not available 😀

tests/conftest.py
```diff
@@ -22,57 +46,100 @@ def check_hash(path: Path, hash_str: str) -> bool:
...
+    # we need to make sure that no other transactions is already open.
+    assert ghidra_program.getCurrentTransactionInfo() is None
+
+    transaction = ghidra_program.openTransaction("reccmp-integration-test")
+    api = FlatProgramAPI(ghidra_program)
+
+    yield api
+
+    # Revert all side effects of the test that just ran
+    transaction.abort()
```
disinvite: Out of curiosity, would the transaction still revert if the calling function (using the context of the ghidra program) threw an exception?

jonschz: I am not entirely sure, but it would be prudent to use try-finally. Good point

PR #331, opened 2026-03-03
Support multiple dirs in YML `source_root` parameter

### LINKED ISSUE #330: No option to target specific source code subdirectories
Consider a project with this directory structure:
```
project/
├── src
├── lib
└── irrelevant-files
```
Our `reccmp-project.yml` accepts a single directory for the `source-root` parameter. If you want both `src` and `lib`, your only option is to specify `.` (the project root), but this will drag in `irrelevant-files` too. If that directory has many source code files for us to parse, it could really reduce reccmp load time.

disinvite:
Fixes #330. The supported formats will now be:
```yml
targets:
  ISLE:
    filename: ISLE.EXE
    source-root: isle
```
or
```yml
targets:
  ISLE:
    filename: ISLE.EXE
    source-root:
      - isle
      - another/dir
```

This field has been and still is required for reccmp. However: we now provide a default empty tuple for `source_root`. This means that if the array is empty or the `source_root` key is missing from the YML, this will pass `pydantic` verification but fail in our code when you try to run with the target. (I added new tests that verify this.)

### DIFF
--- a/.github/workflows/format.yml
+++ b/.github/workflows/format.yml
@@ -12,7 +12,7 @@ jobs:
 
     - uses: actions/setup-python@v5
       with:
-        python-version: '3.12'
+        python-version: '3.14'
 
     - name: Install python libraries
       shell: bash
--- a/.pylintrc
+++ b/.pylintrc
@@ -99,10 +99,6 @@ recursive=no
 # source root.
 source-roots=
 
-# When enabled, pylint would attempt to guess common misconfiguration and emit
-# user-friendly hints instead of false-positive error messages.
-suggestion-mode=yes
-
 # Allow loading of arbitrary C extensions. Extensions are imported into the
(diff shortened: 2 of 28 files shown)


madebr:
> Looks like we broke compatibility with Python 3.9 through 3.11 back in #279, 

3.9 is EOL according to the [graph at python.org](https://www.python.org/downloads/).
Perhaps it makes sense to only support actively maintained python versions?
Our dependencies (pydantic et al) probably also have a similar policy.
It allows us to use "newer" Python features.

disinvite:
> 3.9 is EOL according to the [graph at python.org](https://www.python.org/downloads/).

It seems we do _not_ support 3.9. I don't where I got the idea that we still did, or if we ever did. We only test 3.10 through 3.14.

PR #332, opened 2026-03-06
Unify behavior of source code search

disinvite:
This started with fixing a problem observed in #331 and evolved from there.

### Fixing compatibility issues in `walk_source_dir`
`walk_source_dir` iterates the root source code directory for a decomp project. #279 changed this to use [`Path.walk()`](https://docs.python.org/3/library/pathlib.html#pathlib.Path.walk) instead of [`os.walk()`](https://docs.python.org/3/library/os.html#os.walk). The problem is that `Path.walk()` is only available on Python 3.12 and higher, but we still support a few versions earlier than that. This was missed because no unit tests targeted `walk_source_dir`. We didn't hear anything from users so hopefully the explanation is that no one is using `reccmp` with older Python versions.

This restores the older `os.walk()` implementation so we can keep compatibility with Python 3.10 and 3.11. We now have unit tests to verify the behavior of `walk_source_dir`.

### Returning source code paths in a consistent order
`decomplint` already supported multiple source code search paths at the command line, and we will add this option in #331. In the main `reccmp` script, there is the potential for data to be overwritten if we read metadata for the same address in multiple files. Therefore, we need to ensure:
1. Paths are deduplicated so we do not scan the same file twice
2. Paths are scanned in the same order across platforms

The second point is an issue because `pathlib.PosixPath` is sorted in case-sensitive order, but `pathlib.WindowsPath` is not. The paths are now always sorted in case-insensitive order so the output will be the same across platforms.

To bring this all together, we now have a `source_code_search` function. Tests are included.

### DIFF
--- a/reccmp/dir.py
+++ b/reccmp/dir.py
@@ -134,9 +134,9 @@ class PathResolver:
         return self._memo[path_str]
 
 
-def is_file_c_like(filename: str) -> bool:
-    (_, ext) = os.path.splitext(filename)
-    return ext.lower() in (
+def is_file_c_like(path: Path) -> bool:
+    """Tests the given path for typical C/C++ file extensions."""
+    return path.suffix.lower() in (
         ".c",
         ".h",
         ".cc",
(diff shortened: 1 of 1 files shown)


reccmp/dir.py
```diff
@@ -134,8 +134,9 @@ def resolve_cvdump(self, path_str: str) -> str:
         return self._memo[path_str]
 
 
-def is_file_c_like(filename: Path | str) -> bool:
-    return Path(filename).suffix.lower() in (
+def is_file_c_like(filename: str) -> bool:
```
jonschz: Minor point: I'd refactor the argument to `Path` instead of `str` and simplify the code. We need to convert to a `Path` anyway

reccmp/dir.py
```diff
@@ -144,18 +145,51 @@ def is_file_c_like(filename: Path | str) -> bool:
...
+def walk_source_dir(source: Path, *, recursive: bool = True) -> Iterator[Path]:
     """Generator to walk the given directory recursively and return
     any C++ files found."""
 
-    for subdir, _, files in source.walk():
+    for subdir, _, files in os.walk(source.absolute()):
         for file in files:
             if is_file_c_like(file):
-                yield subdir / file
+                yield Path(os.path.join(subdir, file))
```
jonschz:
```suggestion
    for subdir, _, files in os.walk(source.absolute()):
        for file in files:
            path = Path(os.path.join(subdir, file))
            if is_file_c_like(path):
                yield path
```
Maybe something like this?

Issue #325, opened 2026-02-20
Handle missing NE relocation types


disinvite:
I am working on a PR that will expand NE support to the point where we can actually use it with `reccmp`. Some non-critical things are missing, so I am creating this first so we have an issue to attach to the `# TODO` comments.

In each case, we would benefit from having a sample NE image that demonstrates the relocation. `SKI.EXE` is missing all of these.

These are not handled by Ghidra:
- [ ] `OSFIXUP` relocations
- [ ] Additive relocations

This is handled by Ghidra but I would like to test myself with an example:
- [ ] `LOBYTE` relocations

This is not in `SKI.EXE` but I have manually confirmed that this works using a different image. The TODO item is to add a test.
- [ ] `IMPORTNAME` relocations

PR #341, opened 2026-03-15
Improve support for NE images

disinvite:
This has been in the oven for way too long and I want to get some feedback on the approach.

For `NE` images, we need to apply relocations before we can diassemble imports or far calls. This requires a lot of code to handle each relocation type and walk the "chain" of offsets to find each spot to patch. Our handling is not complete because I could not easily find examples of some relocation types. See: #325. Ghidra is also lacking support for some of these relocations, so we are doing at least as good as them.

We have not had to worry about segments until now. Ghidra's approach is to use `0x1000` as the ID for the first segment, then shift by 8 for each segment that follows. This ensures that there is no overlap in the address space, so you can use the shortcut of combining the two values to create a 32-bit address. For example: `0x10001234` is offset `0x1234` in segment `0x1000`, the first segment.

Strictly speaking: we are not following the intended way of dealing with segment and offset as a 20-bit address. The main point is that we need to match what Ghidra does or our import script will not work out of the box. (I'm not sure how we're going to handle `MZ` because there is no easy way to identify all segments used in the image, but we'll cross that bridge when we come to it.)

The only thing missing now is connecting this to the main `Compare` function and relaxing the restriction that our tools only work with `PE` images. This will come later in a follow-up.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,13 @@
+# To get started with Dependabot version updates, you'll need to specify which
+# package ecosystems to update and where the package manifests are located.
+# Please see the documentation for all configuration options:
+# https://docs.github.com/code-security/dependabot/dependabot-version-updates/configuration-options-for-the-dependabot.yml-file
+
+version: 2
+updates:
+  - package-ecosystem: "pip" # See documentation for possible values
+    directory: "/" # Location of package manifests
+    schedule:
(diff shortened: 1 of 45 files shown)


jonschz (commented):
To be honest, I don't have that much personal investment into NE at the moment, and this is quite a lot to process and get into. @madebr what about you?

PR #345, opened 2026-03-27
Set up trusted publishing to PyPI

disinvite:
Preview: https://test.pypi.org/project/reccmp/

I was experimenting with calver. We will user semver instead.


madebr (approved):

setup.cfg
```diff
@@ -4,10 +4,20 @@ version = attr: reccmp.VERSION
 description = Python recompilation tools
 license = AGPL-3.0-or-later
 classifiers =
+    Development Status :: 4 - Beta
+    Intended Audience :: Developers
+    Operating System :: OS Independent
     Programming Language :: Python :: 3
-    Programming Language :: C
-    Programming Language :: C++
-    License :: OSI Approved :: GNU Affero General Public License v3 or later (AGPLv3+)
```
jonschz, line 10: ~~Do we not need a license anymore?~~

My bad, it's in line 5. Disregard this comment

.github/workflows/publish.yml
```diff
@@ -0,0 +1,74 @@
+name: Publish
+
+on:
+  push:
```
jonschz, line 4: General question: What is the workflow you intended here? Manually create a GitHub release and have the CI run on the tag being pushed? (Edit: I saw that you intend to create a release in this flow, so that's probably not what you had in mind). In that case, I'd set up a trigger like
```
on:
  release:
    types: [created]
```
Furthermore: If we stay on the `push` trigger, should we also restrict this to commits to the `master` branch that have a version tag? I don't think we should release from any branch other than `master` since they don't have branch protection.

Also, maybe we should create a `PUBLISHING.md` just to write down the workflow / procedure as a reference?

jonschz (changes requested):
Thanks for pushing this forward! A few points:

1. Relative links in the README at https://test.pypi.org/project/reccmp/ do not work. Not sure what the best course of action is - maybe a separate, shorter README for PyPI with a link to the full GitHub README?
2. ~~At the moment, you are the only person with access to the PyPI project, correct? Not sure if we want to change that (bus factor etc.). I haven't looked into this, but can you give other people access to the PyPI project without having to share your account credentials?~~ Edit: Addressed
3. ~~I think we should also set up dependabot. I'll look into it~~ -> #346  Edit: Addressed
4. ~~Have you tested this flow somewhere? I couldn't find any runs.~~ Edit: Addressed

I'll take another look once you've addressed these points. Feel free to request another review.

.github/workflows/publish.yml
```diff
@@ -0,0 +1,74 @@
...
+        uses: softprops/action-gh-release@v2
+        with:
+          tag_name: ${{ github.ref_name }}
+          generate_release_notes: true
+          working_directory: dist
+          files: |
+            *.whl
+            *.tar.gz
+
+  rollback:
```
jonschz, line 54: Should this step also [yank](https://docs.pypi.org/project-management/yanking/) a release?

.github/workflows/publish.yml
disinvite: In its current state:
1. Push the tag upstream.
2. Approve the build/publish step to run.

And the release is created from the tag. I wanted the build artifacts added to the release automatically so there's less potential for user error with a manual download/upload step. If there's a better way that begins with creating (draft) releases then that's probably better.

I tried adding `branch: [master]` to the YML but that ended up running the build action for _any_ push to `master`, not the intended combination of `v*` tag _and_ `master` push. If it's possible to add multiple conditions then we should have them.

.github/workflows/publish.yml
disinvite: Can't yank without providing a PyPI token; it's not possible through the trusted publishing action. It seems like the only need to do that would be if we fail while creating the release, so PyPI and GH are out of sync.

.github/workflows/publish.yml
jonschz: Fair enough. If you add a 3-sentence `PUBLISHING.md`, I'll consider this point to be resolved.

jonschz (approved):

disinvite:
New readme is in. You can see it here: https://test.pypi.org/project/reccmp/ (Ignore the weird version number.)

Is that everything we need to go live?

madebr:
Formatting looks good. Is **re**-compilation a good term to use here?
It's not statically recompiling binaries that we do here.
Is the `re-`prefix needed?

jonschz:
Good to go, as far as I am concerned. If we notice any problems, we can always upgrade to a new patch version.

PR #346, opened 2026-03-29
Set up dependabot

jonschz:


.github/dependabot.yml
```diff
@@ -0,0 +1,14 @@
...
+
+version: 2
+updates:
+  - package-ecosystem: "pip" # See documentation for possible values
+    directory: "/" # Location of package manifests
+    schedule:
+      interval: "weekly"
+    cooldown:
+      default-days: 5 # some protection against supply chain attacks
+    target-branch: "add-dependabot" # FIXME: Remove this line after testing is done
```
jonschz: Looks like the config file is only respected on the default branch, GitHub does not detect this config file outside of `master`. So I'll have to merge and fine-tune the config later.

PR #347, opened 2026-03-29
Handle `__real` symbols without embedded size

### LINKED ISSUE #328: Handle `__real@` symbols with no size digit
We have seen symbols like this in MSVC 6 to represent floats:
```
__real@4@3ff18000000000000000
__real@8@4015800bf60000000000
```
The interesting part for us is `@4@` or `@8@` to indicate  single or double precision. (i.e. how many bytes wide is this symbol?)

.NET 2002 uses symbols like this:
```
__real@3b95a025
__real@0000000000000000
```

Here there is no size indicator, but we can use the length of the part that follows the `@` (presumably the bytes used) to guess at the size.

disinvite:
Fixes #328.

### DIFF
--- a/reccmp/cvdump/analysis.py
+++ b/reccmp/cvdump/analysis.py
@@ -1,5 +1,6 @@
 """For collating the results from parsing cvdump.exe into a more directly useful format."""
 
+import string
 from dataclasses import dataclass
 from pathlib import PureWindowsPath
 from reccmp.types import EntityType
@@ -18,6 +19,7 @@ def get_size_from_float_symbol(symbol: str) -> int | None:
     """Get the size for the floating point constant from its symbol that resembles:
     1. __real@X@... where X being 4 or 8 indicates the length in bytes.
     2. __real@..... using the length of the hex string following the prefix.
+    See GH #328 for more context.
     """
     prefix, delim, suffix = symbol.partition("__real@")
 
@@ -25,6 +27,7 @@ def get_size_from_float_symbol(symbol: str) -> int | None:
     if prefix or not delim:
         return None
 
+    # Pattern #1.
     # We have only seen __real@4@ and __real@8@. This is probably overkill, but
     # we should protect against typos in symbols imported from user metadata.
     if "@" in suffix:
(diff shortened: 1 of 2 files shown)


reccmp/cvdump/analysis.py
```diff
@@ -14,6 +14,35 @@
...
+def get_size_from_float_symbol(symbol: str) -> int | None:
+    """Get the size for the floating point constant from its symbol that resembles:
+    1. __real@X@... where X being 4 or 8 indicates the length in bytes.
+    2. __real@..... using the length of the hex string following the prefix.
+    """
+    prefix, delim, suffix = symbol.partition("__real@")
+
+    # Make sure the string was split as we expect.
+    if prefix or not delim:
+        return None
```
jonschz, line 28: I'd log a warning here, and also in all other cases where we return `None` in this function

reccmp/cvdump/analysis.py
```diff
@@ -14,6 +14,35 @@
...
+            return 4
+
+        if suffix.startswith("8@"):
+            return 8
+
+        return None
+
+    size = len(suffix) // 2
+    if size > 0:
+        return size
```
jonschz: I suppose you have examples where you were able to confirm the actual size? It isn't obvious to me that
```
sizeof(__real@4@00000000000000000000) == 4
sizeof(__real@0000000000000000) == 8
```
In any case, I'd add a comment (a link to the GitHub issue would be sufficient) since that logic is highly unintuitive.

jonschz (approved):
Looks good!

reccmp/cvdump/analysis.py
disinvite: To avoid a warning, since it's not possible for user metadata to hit this code yet, what about this configuration?
```python
        elif self.decorated_name.startswith("__real@"):
            size = get_size_from_float_symbol(self.decorated_name)

            if size is not None:
                self.node_type = EntityType.FLOAT
                self.confirmed_size = size
```
since we're checking for a match on two specific patterns that begin with `__real@`. If there's no match, do nothing and keep going.

reccmp/cvdump/analysis.py
disinvite: Good idea. That issue will serve as our documentation for lack of anything official.

PR #353, opened 2026-03-30
Add support for files in `reccmp-project detect`

disinvite:
`reccmp-project detect` uses the project filename to locate orig and recomp files in the paths provided in the `--search-path` parameter. If either file on your local system doesn't match the project filename, the `detect` action will fail.

This enables the option to specify the path to the file instead of just its directory.

For detecting the original binary, we always check against the project SHA256 checksum.

For detecting the recompiled binary, there is no checksum, so we use whichever file is the first to exist and _also_ has a .pdb file with the same base name in the same directory.

I added tests in a few file `test_project_detect.py` since there is a lot of setup code.


tests/test_project_detect.py
```diff
@@ -0,0 +1,237 @@
...
+    BuildFile,
+    UserFile,
+)
+from reccmp.project.detect import (
+    detect_project,
+    DetectWhat,
+)
+from reccmp.formats import PEImage
+
+LEGO1_SHA256 = "14645225bbe81212e9bc1919cd8a692b81b8622abb6561280d99b0fc4151ce17"
```
jonschz: I think this value is already somewhere in the test suite. How about we centralise it?

jonschz (approved):

PR #355, opened 2026-04-03
Create `reccmp-cvdump` tool

disinvite:
I've asked people to run `cvdump.exe` to help diagnose problems a few times, so I think there is enough need for this now. This calls `wine` behind the scenes if needed so it is easier to use.

We only expose the flags that we actually use during parsing. If you run without any options, you get the entire output, same as `cvdump.exe`.

For convenience, I added a `-o / --out` option to dump to a file.


reccmp/tools/cvdump.py
```diff
@@ -0,0 +1,109 @@
...
+    if args.options:
+        cv.options.update(args.options)
+
+    if args.out_file:
+        with open(args.out_file, "w+", encoding="utf-8") as f:
+            with subprocess.Popen(cv.cmd_line(), stdout=f):
+                pass
+    else:
+        with subprocess.Popen(cv.cmd_line()):
+            pass
```
madebr, line 107: Why aren't we using [`subprocess.check_output`](https://docs.python.org/3/library/subprocess.html#subprocess.check_output)?
```suggestion
    with (open(args.out_file, "w+", encoding="utf-8") if args.out_file else sys.stdout) as f:
        with subprocess.Popen(cv.cmd_line(), stdout=f):
            pass
```
(needs an `import sys`)

reccmp/tools/cvdump.py
```diff
@@ -0,0 +1,109 @@
+import argparse
```
madebr, line 3: Needs a shebang (other tools also need it)

reccmp/tools/cvdump.py
disinvite: It was just copied from `cvdump/runner.py` that uses `TextIOWrapper` to split the cvdump sections.

Is `check_output` better than the condensed `Popen` from your suggestion?

madebr: I remember the subprocess documentation discouraging Popen in favor of the utility functions, but I cannot find it anymore.
The check functions raise an exception when the executable fails.

madebr (approved):

PR #358, opened 2026-04-06
Drop image-id specific functions

### LINKED ISSUE #300: Prefer `img, addr` API for DB access
We should prefer to do this:
```python
def get_entity(db: EntityDb, img: ImageId, addr: int) -> str:
    entity = db.get(img, addr)
```

over this:
```python
def get_entity(db: EntityDb, addr: int, is_orig: bool) -> str:
    entity = (
        db.get_by_orig(addr) if is_orig else isle_compare.get_by_recomp(addr)
    )
```
unless there is a good reason not to. (Maybe unit testing?)

This also extends to `EntityBatch` with its `set_orig()` and `set_recomp()` functions.

disinvite:
Fixes #300.

### DIFF
--- a/reccmp/compare/asm/replacement.py
+++ b/reccmp/compare/asm/replacement.py
@@ -1,7 +1,7 @@
 from functools import cache
 from typing import Callable, Protocol
-from reccmp.compare.db import ReccmpEntity
-from reccmp.types import EntityType
+from reccmp.compare.db import EntityDb, ReccmpEntity
+from reccmp.types import EntityType, ImageId
 
 
 class AddrLookupProtocol(Protocol):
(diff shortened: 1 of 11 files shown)


reccmp/compare/db.py
```diff
@@ -419,45 +416,26 @@ def get_one_match(self, addr: int) -> ReccmpMatch | None:
...
 
-    def get_by_orig(self, addr: int, *, exact: bool = True) -> ReccmpEntity | None:
-        """Return the ReccmpEntity at the given orig address.
+    def get(
+        self, img: ImageId, addr: int, *, exact: bool = True
+    ) -> ReccmpEntity | None:
+        """Return the ReccmpEntity at the given address.
         If there is no entry for the address and exact=True (default), return None.
-        Otherwise, return the entity at the preceding orig address if it exists.
+        Otherwise, return the entity at the preceding recomp address if it exists.
```
jonschz:
```suggestion
        Otherwise, return the entity at the preceding address if it exists.
```
or am I missing something here?

reccmp/compare/asm/replacement.py
```diff
@@ -1,7 +1,7 @@
 from functools import cache
 from typing import Callable, Protocol
-from reccmp.compare.db import ReccmpEntity
-from reccmp.types import EntityType
+from reccmp.compare.db import EntityDb, ReccmpEntity
+from reccmp.types import EntityType, ImageId
 
 
 class AddrLookupProtocol(Protocol):
```
jonschz: Is this protocol still being used somewhere else?

jonschz (approved):
Looks good! I assume you have checked for regressions on LEGO1 and ISLE.

reccmp/compare/db.py
disinvite: Missed a spot cleaning up the docstring after merging the two functions.

reccmp/compare/asm/replacement.py
disinvite: Not any more. I'll drop it.

disinvite:
> Looks good! I assume you have checked for regressions on LEGO1 and ISLE.

Yes, all clear in those and `BETA10`.

PR #360, opened 2026-04-10
Add `FOLDED` option for function markers

### LINKED ISSUE #254: Handle Identical Code Folding
MS's linker has a `/OPT:ICF` option which avoids adding identical data: `/OPT:ICF`

From the [msdn documentation](https://learn.microsoft.com/en-us/cpp/build/reference/opt-optimizations):
> CF[=iterations] | NOICF
>
> Use ICF[=iterations] to perform identical COMDAT folding. Redundant COMDATs can be removed from the linker output. The optional iterations parameter specifies the number of times to traverse the symbols for duplicates. The default number of iterations is 1. Additional iterations may locate more duplicates that are uncovered through folding in the previous iteration.
> 
> By default, /OPT:ICF is enabled by the linker unless /OPT:NOICF or [/DEBUG](https://learn.microsoft.com/en-us/cpp/build/reference/debug-generate-debug-info?view=msvc-170) is specified. To override this default and prevent COMDATs from being folded in the program, specify /OPT:NOICF.
> 
> In a debug build, you must explicitly specify /OPT:ICF to enable COMDAT folding. However, because /OPT:ICF can merge identical data or functions, it can change the function names that appear in stack traces. It can also make it impossible to set breakpoints in certain functions or to examine some data in the debugger, and can take you into unexpected functions when you single-step through your code. The behavior of the code is identical, but the debugger presentation can be very confusing. Therefore, we do not recommend that you use /OPT:ICF in debug builds unless the advantages of smaller code outweigh these disadvantages.

It's undefined which symbol name will finally be emitted to the exe/pdb.
So it's probably required to add a tag to every folded function, so reccmp knows it's folded and will verify each recompiled function against the same reference. Perhaps reccmp can add extra useful information to the report.
Ideally, a recompilation project uses the same linker flags as the original (including `/OPT:ICF`) and the functions will also get folded. But bugs happen...

disinvite:
Fixes #254.

### DIFF
--- a/tests/test_parser.py
+++ b/tests/test_parser.py
@@ -733,3 +733,22 @@ def test_mixed_case_module_name(parser):
     # Syntax warning for mixed-case module name.
     assert len(parser.alerts) == 1
     assert parser.alerts[0].code == ParserError.BAD_DECOMP_MARKER
+
+
+def test_folded_option(parser):
+    """Read FOLDED option from nameref and lineref annotations."""
+    parser.read("""\
+        // FUNCTION: HELLO 0x1234 FOLDED
+        int test() { return 5; }
+
+        // LIBRARY: HELLO 0x5555 FOLDED
+        // SomeLibraryFunction
+        """)
+
+    assert parser.functions[0].offset == 0x1234
+    assert parser.functions[0].is_folded is True
+
+    assert parser.functions[1].offset == 0x5555
+    assert parser.functions[1].is_folded is True
+
+    assert len(parser.alerts) == 0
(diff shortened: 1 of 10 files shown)


tests/test_parser.py
```diff
@@ -733,3 +733,22 @@ def test_mixed_case_module_name(parser):
...
+
+        // LIBRARY: HELLO 0x5555 FOLDED
+        // SomeLibraryFunction
+        """)
+
+    assert parser.functions[0].offset == 0x1234
+    assert parser.functions[0].is_folded is True
+
+    assert parser.functions[1].offset == 0x5555
+    assert parser.functions[1].is_folded is True
```
madebr: Testing wise, it should also  test `.is_folder` is False when `FOLDED` is not present.

I think it should also test a function with multiple annotations (for multiple targets) with FOLDED present and non-present. (Perhaps that can be added to an already test exercising this?)

foxtacles (approved):

madebr (approved):

PR #363, opened 2026-04-11
Store entity size for each address space

disinvite:
Writes to the `size` keyword for an entity will now update `orig_size` or `recomp_size` behind the scenes, depending on the `ImageId`. This means it is now (finally) possible for a function to have its orig size set independently from the recomp size read from the PDB.

The `any_size()` method for the `ReccmpEntity` and `ReccmpMatch` objects will always return an `int`, defaulting to zero, with the recomp size preferred. This matches the existing behavior of the (now deleted) `size` property. You can specify that you want orig size by providing the image id. 

The `size()` method checks only the size for the `ImageId` you specify, returning `None` if it is unset.


tests/test_function_comparator.py
```diff
@@ -557,3 +558,87 @@ def test_source_reference_without_line_annotation(
...
+    assert diffreport.diff.codes == [("equal", 0, 1, 0, 1)]
+    assert diffreport.diff.orig_inst == [("0x0", "nop ")]
+    assert diffreport.diff.recomp_inst == [("0x0", "nop ")]
+
+
+def test_compare_with_distinct_size(
+    db: EntityDb, lines_db: LinesDb, report: ReccmpReportProtocol
+):
+    """In this example, the recomp function is not implemented enough to match.
+    However: if we only read 1 byte (recomp_size) from the orig binary, it will appear to match.
```
jonschz, line 603:
```suggestion
    However: if we only read 1 byte (recomp_size) from the orig binary, it would appear to match.
```

jonschz (approved):
Very nice!

PR #364, opened 2026-04-11
Add option to use custom encoding for strings

### LINKED ISSUE #229: Encoding utf-8 may be wrong for old windows files

some100:
Fixes #229, probably. Adds a per-target option to specify encoding of source files in the project YAML.

### DIFF
--- a/reccmp/compare/core.py
+++ b/reccmp/compare/core.py
@@ -76,7 +76,8 @@ class Compare:
     recomp_bin: Image
     report: ReccmpReportProtocol
     target_id: str
-    encoding: str
+    src_encoding: str
+    bin_encoding: str
     types: CvdumpTypesParser
     function_comparator: FunctionComparator
     data_sources: list[TextFile]
@@ -89,7 +90,7 @@ class Compare:
         recomp_bin: Image,
         pdb_file: CvdumpAnalysis,
         target_id: str,
-        encoding: str = "utf-8",
+        encoding: str | None = None,
         code_files: list[TextFile] | None = None,
         data_sources: list[TextFile] | None = None,
     ):
@@ -97,7 +98,8 @@ class Compare:
         self.recomp_bin = recomp_bin
         self.cvdump_analysis = pdb_file
         self.target_id = target_id
-        self.encoding = encoding
+        self.src_encoding = encoding or "utf-8"
+        self.bin_encoding = encoding or "latin1"
 
         if isinstance(code_files, list):
             self.code_files = code_files
@@ -142,7 +144,7 @@ class Compare:
             self.orig_bin,
             self.target_id,
             self._db,
-            self.encoding,
+            self.src_encoding,
             self.report,
         )
 
@@ -165,12 +167,12 @@ class Compare:
             create_imports(self._db, img_id, binfile)
             create_import_thunks(self._db, img_id, binfile)
             create_analysis_floats(self._db, img_id, binfile)
-            create_analysis_strings(self._db, img_id, binfile, self.encoding)
+            create_analysis_strings(self._db, img_id, binfile, self.bin_encoding)
             create_seh_entities(self._db, img_id, binfile)
             create_thunks(self._db, img_id, binfile)
             create_analysis_vtordisps(self._db, img_id, binfile)
             complete_partial_floats(self._db, img_id, binfile)
-            complete_partial_strings(self._db, img_id, binfile, self.encoding)
+            complete_partial_strings(self._db, img_id, binfile, self.bin_encoding)
 
         match_imports(self._db)
         match_exports(self._db, self.orig_bin, self.recomp_bin)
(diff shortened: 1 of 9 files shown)


disinvite:
Hello! I like the approach, but I'm not sure `latin1` is a good default value for _reading_ the files.

I used `latin1` when searching for strings so that we would not exclude extended ASCII text. (IIRC it was a string with the copyright symbol that I wanted to capture.) But we should default to `utf-8` for reading _unless_ another encoding is specified in the YML with this new change.

reccmp/tools/decomplint.py
```diff
@@ -120,18 +127,24 @@ def main():
...
         for path in args.paths:
             if path.is_dir() or path.is_file():
                 search_paths.append(path)
             else:
                 logger.error("Invalid path: %s", path)
-
-    # Use set() so we check each file only once.
-    reduced_file_list = source_code_search(search_paths)
-
-    codefiles = list(TextFile.from_files(reduced_file_list, allow_error=True))
```
disinvite, line 134: Did it not work with the `encoding` keyword set here?

some100: I moved that section since I wanted each target to be read with their custom specified encoding. Otherwise if there were multiple targets with differing encodings then it'd only be able to use one over the other.

disinvite:
I think we should have a test for reading the string from the code file (i.e. the changes to `ingest.py`).

Does this new test for `test_compare_ingest_code.py` look correct? If so, feel free to add it. You will have to add `from unittest.mock import Mock` to the imports at the top of the file.

```python
def test_read_gb2312_string(db: EntityDb, lines_db: LinesDb):
    """Make sure we read the full length of the string in GB 2312 encoding and create the entity. GH #364"""
    string_text = "你吃饭了吗"
    string_bytes = string_text.encode("gb2312") + b"\x00"

    # Can't use RawImage here; it is missing some functions from PEImage.
    orig_bin = Mock(spec=[])
    orig_bin.read = lambda _, size: string_bytes[:size]
    orig_bin.imagebase = 0
    orig_bin.is_valid_vaddr = Mock(return_value=True)

    files = (
        TextFile(
            PurePath("test.cpp"),
            dedent(f"""\
                // STRING: TEST 0x1000
                const char* test = "{string_text}";
                """),
        ),
    )

    load_markers(files, lines_db, orig_bin, "TEST", db) # <-- need to add encoding param here

    entity = db.get(ImageId.ORIG, 0x1000)
    assert entity is not None
    assert entity.get("type") == EntityType.STRING
    assert entity.get("size") == len(string_bytes)
    assert entity.name.encode() == f'"{string_text}"'.encode("unicode_escape")
```

disinvite:
Sorry, more notes. 😀

I think the behavior ought to be:
- If `encoding` is not specified for the target, assume `utf-8` when reading via `TextFile` but assume `latin1` when reading/decoding non-widechar strings from the binary.
- If `encoding` is specified, use it for both reading files and reading/decoding from the binary.

This (new) test would fail if we assume `utf-8` when reading/decoding:
```python
def test_complete_partial_strings_extended_ascii(db: EntityDb):
    """Should assume extended ASCII for non-widechar strings unless directed otherwise."""
    text = "8½"

    # UTF-8 and Latin1 only overlap up to 0x7f.
    assert len(text.encode("utf-8")) != len(text.encode("latin1"))

    binfile = Mock(spec=[])
    binfile.read_string = Mock(return_value=text.encode("latin1"))

    with db.batch() as batch:
        batch.set(ImageId.ORIG, 100, type=EntityType.STRING)

    complete_partial_strings(db, ImageId.ORIG, binfile)

    e = db.get(ImageId.ORIG, 100)
    assert e is not None
    assert e.size == len(text) + 1
```

reccmp/project/detect.py
```diff
@@ -182,6 +185,9 @@ class RecCmpTarget:
     # SHA-256 checksum of the original binary.
     sha256: str
 
+    # Encoding of the source code files for this target.
+    encoding: str | None
```
some100, line 189: there is probably a nicer way to do this without using a none union here (maybe by having src_encoding and bin_encoding earlier on reccmptarget and reccmppartialtarget instead of just a single encoding field)

disinvite: Looks good to me! We can tweak it later if need be

reccmp/compare/core.py
```diff
@@ -139,6 +144,7 @@ def run(self):
             self.orig_bin,
             self.target_id,
             self._db,
+            self.src_encoding,
```
disinvite: Should this be `bin_encoding` instead? Since it impacts how we measure how many bytes to read from the binary.

reccmp/compare/ingest.py
```diff
@@ -140,13 +140,14 @@ def load_cvdump_lines(
...
 
-# pylint: disable=too-many-positional-arguments
+# pylint: disable=too-many-positional-arguments, too-many-arguments
 def load_markers(
     code_files: Sequence[TextFile],
     lines_db: LinesDb,
     orig_bin: PEImage,
     target_id: str,
     db: EntityDb,
+    encoding: str = "utf-8",
```
disinvite: For the same reason as above, I think the default here should be `latin1`.

disinvite (approved):

Issue #320, opened 2026-02-13
Handle `FORWARD REF` leaves with no `UDT`


disinvite:
We handle forward references for `LF_STRUCTURE` or `LF_CLASS` leaves by checking if the `field list type` is `0x0000`. We don't parse the `FORWARD REF` attribute at all. If not equal to the leaf id, the `UDT` value tells what other leaf is referenced. This typically looks like:
```
0x2e94 : Length = 38, Leaf = 0x1504 LF_CLASS
	# members = 0,  field list type 0x0000, FORWARD REF, 
	Derivation list type 0x0000, VT shape type 0x0000
	Size = 0, class name = MxStreamProvider, UDT(0x0000b203)
```

However, some forward refs have no `UDT` at all:
```
0x2562 : Length = 30, Leaf = 0x1505 LF_STRUCTURE
        # members = 0,  field list type 0x0000, FORWARD REF,
        Derivation list type 0x0000, VT shape type 0x0000
        Size = 0, class name = HWND__
```

We have somehow avoided this problem up until now. It must be that these types are not referenced anywhere that we have been looking in either `reccmp` or `datacmp`.

disinvite:
I couldn't find an example in any of the PDBs I was working on when this opened. So it is just `HWND__`. If that's not enough cause for us to do anything beyond what's already there in the Ghidra scripts, we can close this.

PR #366, opened 2026-04-12
Add more Ghidra import tests

### LINKED ISSUE #380: Enums with negative values are not handled correctly
From `BETA10`:

```
0x5695 : Length = 50, Leaf = 0x1203 LF_FIELDLIST
	list[0] = LF_ENUMERATE, public, value = (LF_CHAR) -1(0xFF), name = 'c_unknownminusone'
	list[1] = LF_ENUMERATE, public, value = 8, name = 'c_unknown8'

0x5696 : Length = 42, Leaf = 0x1507 LF_ENUM
	# members = 2,  type = T_INT4(0074) field list type 0x5695
NESTED, 	enum name = LegoCarBuild::Unknown0xf8, UDT(0x00005696)
```

The `c_unknownminusone` is not imported correctly into Ghidra. Other parts of the code might also be affected. I will provide an xfail test in #366 .

jonschz:
I'd be interested to hear opinions on the decompiled function assertions. Overall I think they are a good idea, since that ultimately is the Ghidra output we care most about. On the other hand, they are at risk of breaking on Ghidra updates. Thoughts would be appreciated.

I will submit further tests in separate PRs. I'd like to get the Ghidra import test coverage as high as reasonably possible. The virtual inheritance code turned out to be particularly difficult, so I'll submit a PR in the future.


reccmp/cvdump/types.py
```diff
@@ -356,8 +356,15 @@ def get(self, type_key: CvdumpTypeKey) -> TypeInfo:
         if obj is None:
             raise CvdumpKeyError(type_key)
 
-        if obj.get("type") == "LF_POINTER":
-            return self.get(CVInfoTypeEnum.T_32PVOID)
+        # TODO: Validate these changes, test coverage
```
jonschz: Address TODO

reccmp/cvdump/types.py
```diff
@@ -356,8 +356,15 @@ def get(self, type_key: CvdumpTypeKey) -> TypeInfo:
         if obj is None:
             raise CvdumpKeyError(type_key)
 
-        if obj.get("type") == "LF_POINTER":
-            return self.get(CVInfoTypeEnum.T_32PVOID)
```
jonschz, line 360: I don't know if that was a limitation we always had or a bug we introduced at some point, but it did lead to some global constants (e.g. arrays) not being imported correctly.

reccmp/ghidra/importer/ghidra_helper.py
```diff
@@ -148,7 +148,7 @@ def set_ghidra_label(api: FlatProgramAPI, address: int, label_with_namespace: st
             and existing_label_name == name
         ):
             logger.debug(
-                "Label '%s' at 0x%s already exists", label_with_namespace, address_hex
+                "Label '%s' at %s already exists", label_with_namespace, address_hex
```
jonschz, line 151: This used to render as `0x0xdeadbeef`

reccmp/ghidra/importer/pdb_extraction.py
```diff
@@ -31,7 +31,6 @@ class CppRegisterSymbol(CppStackOrRegisterSymbol):
 
 @dataclass
 class FunctionSignature:
-    original_function_symbol: SymbolsEntry
```
jonschz, line 34: This was dead code: It was written but never read

tests/test_ghidra_integration_function_imports.py
```diff
@@ -0,0 +1,218 @@
...
+            FieldListItem(offset=0, name="m_name", type=CVInfoTypeEnum.T_32PCHAR),
+            FieldListItem(offset=4, name="m_type", type=CVInfoTypeEnum.T_INT4),
+        ],
+    )
+    compare.types.keys[anim_actor_entry_key] = CvdumpParsedType(
+        type="LF_CLASS",
+        name="LegoAnimActorEntry",
+        field_list_type=anim_actor_entry_field_list_key,
+        size=8,
+    )
```
jonschz: I am looking into setting up the PDB data in string from and running the parser here. This is quite tedious to set up from existing examples.

reccmp/ghidra/importer/type_importer.py
```diff
@@ -137,7 +136,10 @@ def _import_scalar_type(self, type_key: CvdumpTypeKey) -> DataType:
         cvtype = CvdumpTypeMap[type_key]
 
         if cvtype.pointer is None:
-            return get_scalar_ghidra_type(type_key)
+            # TODO: Check if this type addition needs to be done for more types, or if we can centralise it
```
jonschz: Address TODO

tests/test_ghidra_integration.py
jonschz, line 1: move detection failed because a lot was changed

tests/ghidra_integration_test_setup.py
```diff
@@ -41,6 +41,17 @@ def ghidra_integration_test_program(
...
+        # Gets rid of spurious stack traces from a misunderstanding between pytest and jpype.
+        # See https://jpype.readthedocs.io/en/latest/userguide.html#errors-reported-by-python-fault-handler
+        import faulthandler
+
+        faulthandler.enable()
+        faulthandler.disable()
+    # pylint: disable-next=broad-exception-caught # This is fine to fail, we don't need to handle it
+    except Exception:
+        pass
+
```
jonschz, line 61: As discussed on Matrix with @disinvite

reccmp/cvdump/types.py
jonschz: Decided to pull this out into a separate PR

reccmp/cvdump/types.py
jonschz: Decided to pull this out into a separate PR

jonschz:
@disinvite there is one more TODO to address, but other than that, this could already benefit from a first review. I have discovered a few more problems, but I will submit separate tickets / PRs for that.

reccmp/ghidra/importer/type_importer.py
disinvite: What problem led to adding `add_data_type_or_reuse_existing`? The `HWND_` thing from #320? (so, covered by `test_forward_ref_to_pre_existing_type`?)

disinvite:
This all looks great. I especially like how you're patching `isle.exe` in memory so we can test on arbitrary instructions.

The decompiler approach seems fine to me. We're only testing cases where Ghidra should have all the information it needs to produce good output, and specifically where `reccmp` has provided this information. I would say the type importer is as important as the function importer in this regard because of how it influences struct/array lookups. (For me, this is where the decompiler is most helpful.)

reccmp/ghidra/importer/type_importer.py
jonschz: Curiously, this caused trouble in the function import test `test_record_array_access`.

According to my testing, there is a difference in Ghidra whether you instantiate a scalar type or whether you acquire the same type from the type database. I.e. 
```py
not_from_database = LongDataType()
from_database = api.getCurrentProgram()
        .getDataTypeManager()
        .addDataType(LongDataType(), DataTypeConflictHandler.KEEP_HANDLER)
assert not_from_database != from_database
```
In the test, this caused an error
```
ERROR    reccmp.ghidra.importer.function_importer:function_importer.py:359 Type mismatch for parameter: [int param0@Stack[0x4]:4] in Ghidra, CppStackSymbol(name='p_index', data_type=<CVInfoTypeEnum.T_INT4: 116>, stack_offset=4) in PDB
```
which then causes `p_index` to get a mismatch on its name.

I'll need to experiment and think a little more if there in which cases we _don't_ want to use `add_data_type_or_reuse_existing`. Will probably know more once all the type imports have test coverage for collisions.

jonschz: I have found a more specific, type-import-only test case that shows the necessity of this change. Good enough for me.

tests/test_ghidra_integration_type_imports.py
```diff
@@ -0,0 +1,371 @@
...
+    )
+    _assert_legoanimactorentry(imported_pointer.dataType)
+
+    second_import = type_helper.type_importer.import_pdb_type_into_ghidra(
+        legoanimactor_pointer_key
+    )
+    assert second_import == imported_pointer
+
+
+def test_pointer_to_scalar(type_helper: GhidraTypeTestHelper):
```
jonschz, line 253: This new test also fails without the `add_data_type_or_reuse_existing`

jonschz:
Fixes #380

reccmp/ghidra/importer/importer.py
```diff
@@ -91,10 +91,6 @@ def _do_execute_import(
 ):
     pdb_functions = extraction.get_function_list()
 
-    if api is None:
-        logger.info("Completed the dry run outside Ghidra.")
-        return
-
```
jonschz, line 97: Removed obsolete code

jonschz:
This should now be all I want to add to this PR. I fixed one problem and moved another one to a separate PR + issue.

disinvite (approved):

PR #371, opened 2026-04-15
Hide 'failed to match' messages behind `--debug`

### LINKED ISSUE #342: Limit `Failed to match` messages to code annotations
The functions that try to match orig and recomp entities will report error messages like `Failed to match function at 0x1234 with name 'Hello'` if we try to find an orig entity with the given name but came back empty. This made sense when it was only possible to provide metadata for orig entities via the code annotations.

If you are providing metadata via CSV or (in the future) from a PDB or MAP file that belongs to the original program, we should not report a failed match. The message isn't useful because the orig metadata would be more comprehensive than the recompiled code until the completion of the decomp.

disinvite:
Fixes #342.


disinvite:
Outside of the Ghidra scripts, we only use `logger.debug` when parsing `SYMBOLS` from `cvdump`.

jonschz (approved):

PR #375, opened 2026-04-17
Condensed `README` for PyPI

disinvite:
Blocker for #345. Relative links to other doc pages don't work on PyPI, so we need something short that explains the project and redirect users here to read more.

Too short or just right?


jonschz (approved):
This is a decent start. We can always add more if we feel the need to do so later. Adding too much here increases the risk of forgetting to maintain it.

madebr (approved):

PR #376, opened 2026-04-17
cmake: update ghidra project through build system target

madebr:
After providing CMake with the required information, a CMake build target is creates that will update your Ghidra project.

Example for the LEGO Racers decompilation project:

Configure with:
```sh
cmake .. \
  -DRECCMP_LEGORACERS_GHIDRA_LOCAL_PROJECT_PATH=/path/to/lego_racers.gpr
  -DRECCMP_LEGORACERS_GHIDRA_FILE=/LEGORacers.exe
```

The following command will then update /LEGORacers.exe in your Ghidra project:
```
cmake --build . --target reccmp-import-ghidra-LEGORACERS
```

If you target multiple binaries, then running the following command will update all of them:
```
cmake --build . --target reccmp-import-ghidra
```
(ninja parallelization is inhibited using pools)


disinvite:
Is this intended for CI? The import script needs to lock the Ghidra project, so it won't work if you are building locally with Ghidra open.

madebr:
This is for local development, allowing a dev to update their ghidra project from CMake without needing an extra script.

disinvite:
Ah! I missed that it was its own build target, not something that is run every time.

disinvite (approved):

PR #377, opened 2026-04-17
project: add "reccmp-project update" for updating the project

madebr:
By running , reccmp will (try to) update the project files to latest. Currently, only thing it does is copy the latest version of reccmp.cmake.

https://github.com/isledecomp/reccmp/pull/376 is trying to update `cmake/reccmp.cmake`.

Running `reccmp-project update` will copy the latest files.

Thoughts and/or alternative approaches?

Will we ever need to add a migration path whenever need to add something that is backwards incompatible?
Should I rename this command to `reccmp-project migrate` instead?
Yes because it modifies the project.
No because this is not a real migration.

### DIFF
--- a/reccmp/project/common.py
+++ b/reccmp/project/common.py
@@ -1,3 +1,17 @@
+import textwrap
+
 RECCMP_PROJECT_CONFIG = "reccmp-project.yml"
 RECCMP_USER_CONFIG = "reccmp-user.yml"
 RECCMP_BUILD_CONFIG = "reccmp-build.yml"
+
+
+def helper_create_project(target_name: str, filename: str, sha256: str) -> str:
+    """Creates YML for a project file with one target using the given parameters."""
+    return textwrap.dedent(f"""\
+        targets:
+          {target_name}:
+            filename: {filename}
+            source-root: sources
+            hash:
+              sha256: {sha256}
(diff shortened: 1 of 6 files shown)


jonschz:
Is it realistic to assume that our generated cmake file will be able to address the needs of most projects? I have a feeling that most projects will have custom logic in their cmake files anyway, and any attempt to migrate that on our side will break something on theirs.

Furthermore, I fear that people will likely forget about running this new command, or not know that it exists in the first place. I'd rather advertise new features in release notes, and then they can always re-generate a fresh project and merge.

Or am I missing your point here?

madebr:
The point of the cmake script is not to provide a general fix, but a simple way to get reccmp running without messing a second time with paths.

In #376 I updated the CMake script. Improvements would've remained inaccessible to users without this command.

disinvite:
Ignoring the CMake change for a moment... we will need this when it's time for a new schema version. (We have enough users now that it would be a big help to us for this to be automated.)

I don't know if any users have added comments to their YML file but AFAIK those would be dropped using our pydantic import/export process.

We have to consider the case where the project maintainer updates the schema version on the repo, but an individual user's files (user.yml and build.yml) are on an earlier version.

PR #378, opened 2026-04-17
Fix typos found through codespell

madebr:


disinvite (approved):

jonschz:
Is that something we could use on our CI as well?

PR #381, opened 2026-04-18
Exclude default values from YML

disinvite:
We have many optional items in the `reccmp-project.yml`, mostly lists of items that default to empty lists. The file produced by `reccmp-project create` looks like:
```yaml
targets:
  JMAN:
    filename: JMAN.EXE
    source_root: .
    hash:
      sha256: 243f62328af86447376d1d94a2458bc96e64bc9b0e70c27b1470a4c21a40f2a0
    data_sources: []
    encoding:
    ghidra:
      ignore_types: []
      ignore_functions: []
      name_substitutions: []
      allow_hash_mismatch: false
    report:
      ignore_functions: []
```

If we exclude items that are set to their default value, we get only:
```yaml
targets:
  JMAN:
    filename: JMAN.EXE
    source_root: .
    hash:
      sha256: 243f62328af86447376d1d94a2458bc96e64bc9b0e70c27b1470a4c21a40f2a0
```
One argument against doing this is our (admittedly) [minimal documentation](https://github.com/isledecomp/reccmp/blob/master/docs/project_files.md) for the project schema. Displaying all the possible options could be considered helpful to users.

### DIFF
--- a/reccmp/project/config.py
+++ b/reccmp/project/config.py
@@ -2,7 +2,7 @@
 
 from pathlib import Path
 from dataclasses import dataclass
-
+from io import StringIO, TextIOBase
 from pydantic import AliasChoices, BaseModel, Field
 import ruamel.yaml
 from .yml_extensions import PathSequence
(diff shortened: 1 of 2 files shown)


jonschz:
For now I think exporting all properties isn't the worst idea since it makes working with the config easier. I'd see this differently if we had schemas (#105).

disinvite:
Let's table this for now.

PR #382, opened 2026-04-18
ci: add codespell check

madebr:
This adds a codespell ci actions.

[Example run where I added a few typos](https://github.com/madebr/reccmp/actions/runs/24613503988/job/71971825337#step:8:17)

### DIFF
--- a/reccmp/ghidra/ghidra_import.py
+++ b/reccmp/ghidra/ghidra_import.py
@@ -221,7 +221,9 @@ def shared_repository_program(args: argparse.Namespace):
                     create_keep_file,
                 )
 
-                dom_file.checkin(checkin_handler, TaskMonitor.DUMMY)
+                dom_file.checkin(  # codespell:ignore checkin
+                    checkin_handler, TaskMonitor.DUMMY
+                )
 
                 logger.debug("DomainFile checked in successfully.")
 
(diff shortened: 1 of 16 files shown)


reccmp/ghidra/ghidra_import.py
```diff
@@ -221,7 +221,9 @@ def shared_repository_program(args: argparse.Namespace):
                     create_keep_file,
                 )
 
-                dom_file.checkin(checkin_handler, TaskMonitor.DUMMY)
+                dom_file.checkin(  # codespell:ignore checkin
```
jonschz, line 224: Does it complain about `.checkin` or the `checkin_handler`? The latter can be renamed

jonschz (approved):
Nice!

reccmp/ghidra/ghidra_import.py
madebr: It's about `checkin`.
This is the warning it emits:
```
./reccmp/ghidra/ghidra_import.py:224: checkin ==> checking, check in
```
It says nothing about the initialization of that variable `checkin_handler = ...` in that method.

The util is not infallible. (I introduced a few typo's which it did not detect)
But useful nonetheless.

tests/test_compare_ingest_csv.py
```diff
@@ -115,6 +115,7 @@ def test_load_csv_overwrite(db: EntityDb):
 
 def test_load_csv_with_errors(db: EntityDb):
     """Should skip lines with a syntax error and create entities for the rest."""
+    # codespell:ignore-begin
```
madebr, line 118: @jonschz in #366, you'll need to wrap cvdump output with these annotations. The util complains about `Parms`.

Issue #336, opened 2026-03-11
Code annotations incompatible with mixed-case target


disinvite:
If the target in your `reccmp-project.yml` uses lower-case chars, everything will work except parsing code annotations. The problem is that the code parser converts the module name to upper case, and so it cannot match the input target.

Historically, reccmp target names have been upper-case, starting with `LEGO1` and `ISLE`, but we don't document this as a requirement. Using a mixed-case module in your annotation will raise the `bad_recomp_marker` warning in `decomplint`, but (IIRC) my intention was for processing to go through regardless.

We could patch this easily by converting the input target name to upper case, but this implies that targets have case-insensitive uniqueness in the YML. (i.e. Another thing to validate.)

#230 is another potential solution (apart from its own merits as an independent enhancement).

```python
@pytest.mark.xfail(reason="TODO")
def test_mixed_case_module_name(parser):
    parser.read(
        dedent(
            """\
        // FUNCTION: Hello 0x1234
        void interesting_function() {
        }
        """
        )
    )

    # Preserve module case
    assert len(parser.functions) == 1
    assert parser.functions[0].module == "Hello"

    # Should alert to this
    assert len(parser.alerts) == 1
    assert parser.alerts[0].code == ParserError.BAD_DECOMP_MARKER
```

disinvite:
One problem is that we have many items that are all intertwined.

We have two possible inputs:
- **A**: Target name from `reccmp-project.yml`.
- **B**: Base name of the original binary.

Whichever one you provide (via the `--target` or `--paths` option), you get this internal value:

- **C**: Query for decomp annotation parser.

(#230 proposes to allow you to specify **C** explicitly.)

**C** has to match **X**, the identifier used in each code annotation. My intention was for **C** to match **X** case-insensitively. This flexibility is not actually there, hence this issue.

##

I agree that **A** should require case-sensitive match. This is analogous to a cmake/makefile target.

**B** may not match anything relevant to the project because the user can name their local copy of the file anything they want. We also do not call `Path.resolve()` before extracting the name. This means that on Windows (where paths are case-insensitive) we will use whatever value was provided. (This is probably worth its own issue 😶)

I'm not sure what to do about **X** that will feel correct to users. My instinct would be to consider this a typo, not two different targets built out of the same file:
```cpp
// STUB: Isle 0x1000
void func_1000() {
}

// STUB: ISLE 0x2000
void func_2000() {
}
```

[...]

PR #393, opened 2026-04-25
Fix: file names in SVG don't use original case

foxtacles:


disinvite:
We may be stuck with a lower-case comparison for any files already out in the field. This gets into a whole other conversation about the case-sensitivity of the host OS (overlap with #336).

Maybe we should centralize the check with this added to `ReccmpStatusReport`:
```python
    def has_same_source(self, other: "ReccmpStatusReport") -> bool:
        """Were both reports derived from the same reccmp target?"""
        return self.filename.lower() == other.filename.lower()
```
and then patch the sites in `aggregate.py` and `utils.py`.

And this for `test_report.py`:
```python
def test_same_source():
    """Reports contain the filename of the original binary.
    If these filenames match case-insensitively, the reports 
    came from the same reccmp target."""
    report_lower = ReccmpStatusReport(filename="test.exe")
    report_upper = ReccmpStatusReport(filename="TEST.EXE")
    report_mixed = ReccmpStatusReport(filename="Test.Exe")

    reports = (report_lower, report_upper, report_mixed)

    # Check all case-variant pairs, including identity
    for x, y in zip(reports, reports):
        assert x.has_same_source(y)

    # Check with different filename
    report_hello = ReccmpStatusReport(filename="hello.dll")
    for x in reports:
        assert x.has_same_source(report_hello) is False
        assert report_hello.has_same_source(x) is False
```

disinvite (approved):

PR #394, opened 2026-04-25
Just use Python dicts for the entity db

disinvite:
Closing the era of the SQLite database with JSON that began back in #25. At the time, I was thinking about:
1. The need to support as-yet-unknown entity attributes.
2. The code complexity required to get good performance from pure-Python.
3. The potential need for thread-safety and atomicity.

None of these have really been a factor in the 18 months since. We don't yet have multi-threaded data import or multi-user support (daemon job) that would demand a locking system. (And it's not like SQL is the only game in town for that.) The list of entity attributes have been somewhat stable for a while. I would say code readability is far worse because of the gymnastics required to get good performance from _SQLite_. (Never materialize `ReccmpEntity` objects and do as much work as possible inside the database at the C layer.)

On top of it all... SQLite isn't faster! The difference is less pronounced with cProfile, but the clock-on-the-wall time is faster. (At least that's the case on my somewhat outdated Windows box with Python 3.14).

There are certainly more optimization steps possible. We should get off the `kvstore` dictionary and just use named properties (with types) on `ReccmpEntity` unless this proves to be a lot slower for some reason. It might not hurt to have an index on `EntityType` since this is the most commonly used attribute in "queries".

### DIFF
--- a/reccmp/compare/db.py
+++ b/reccmp/compare/db.py
@@ -293,7 +293,9 @@ class EntityDb:
                 else:
                     entities[addr] = ReccmpEntity(None, addr, values)
             else:
-                entities[addr]._kvstore.update(values) # pylint: disable=protected-access
+                entities[addr]._kvstore.update(
+                    values
+                )  # pylint: disable=protected-access
 
         self._update_addr_index(image, new_addrs)
 
--- a/reccmp/compare/db.py
+++ b/reccmp/compare/db.py
@@ -287,15 +287,14 @@ class EntityDb:
         for addr, values in rows:
             new_addrs.add(addr)
 
+            # pylint: disable=protected-access
             if addr not in entities:
                 if image == ImageId.ORIG:
                     entities[addr] = ReccmpEntity(addr, None, values)
                 else:
                     entities[addr] = ReccmpEntity(None, addr, values)
             else:
-                entities[addr]._kvstore.update(
-                    values
-                )  # pylint: disable=protected-access
+                entities[addr]._kvstore.update(values)
 
         self._update_addr_index(image, new_addrs)
 
--- a/reccmp/compare/db.py
+++ b/reccmp/compare/db.py
@@ -2,67 +2,11 @@
 addresses/symbols that we want to compare between the original and recompiled binaries.
 """
 
-import sqlite3
+import bisect
 import logging
-import json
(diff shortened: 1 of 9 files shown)


disinvite:
Does this still need refinement before we can merge it?

**Edit**: I just checked this in to do some testing on another machine.

We could combine the internal structures by `ImageId`. You get less code and the performance appears to be the same if you pull out the item you want to update:

```python
def _update_addr_index(self, img: ImageId, addrs: set[int]):
    """Update the ordered list of addresses in this address space."""
    extent = self._addr_set[img]
    order = self._addr_order[img]
    order.extend(addrs - extent)
    order.sort()
    extent |= addrs

def bulk_insert(self, image: ImageId, rows: Iterable[tuple[int, dict[str, Any]]]):
    assert image in (ImageId.ORIG, ImageId.RECOMP), "Invalid image id"
    new_addrs = set()
    entities = self._entities[image]

    for addr, values in rows:
        new_addrs.add(addr)

        if addr not in entities:
            if image == ImageId.ORIG:
                entities[addr] = ReccmpEntity(addr, None, values)
            else:
                entities[addr] = ReccmpEntity(None, addr, values)
        else:
            entities[addr]._kvstore.update(values) # pylint: disable=protected-access

    self._update_addr_index(image, new_addrs)
```

madebr:
I just ran `hyperfine "reccmp-reccmp --target LEGORACERS" --warmup 2` using reccmp from current master and this pr and see a performance improvement.

master:
```
Benchmark 1: reccmp-reccmp --target LEGORACERS
  Time (mean ± σ):      2.088 s ±  0.045 s    [User: 1.577 s, System: 0.126 s]
  Range (min … max):    2.043 s …  2.201 s    10 runs
```

this pr:
```
Benchmark 1: reccmp-reccmp --target LEGORACERS
  Time (mean ± σ):      1.882 s ±  0.022 s    [User: 1.372 s, System: 0.124 s]
  Range (min … max):    1.858 s …  1.926 s    10 runs
```

jonschz (approved):
Sound like a good decision overall. I didn't read every line of code, but if there are no known regressions I'd keep it like this.

disinvite:
I'm going to do a final regression test later tonight then merge this.

PR #401, opened 2026-05-01
Fix datacmp crash in `unpack`

jonschz:
I haven't looked into why this happens, but for now this certainly an improvement. You can trigger the bug on the current `racers` build using `reccmp-datacmp --target LEGORACERS`.

### DIFF
--- a/reccmp/compare/variables.py
+++ b/reccmp/compare/variables.py
@@ -3,7 +3,7 @@ import logging
 from dataclasses import dataclass
 from enum import Enum
 from typing import NamedTuple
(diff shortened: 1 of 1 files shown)


disinvite:
Which variable caused the crash? (Or what was the last successful one?) I can't reproduce here.

jonschz:
Odd, now I am also unable to reproduce the issue, though it was stable for me when it happened. Still, I think this error handling is nice to have in case the problem appears again.

disinvite (approved):

PR #403, opened 2026-05-02
Derive more data from the finalized `ReccmpStatusReport`

disinvite:
Until #307, it was not possible to access the full generated asm for a function outside of the function comparator. The debug feature that dumps all assembly in both binaries to two files can now be moved up to `asmcmp.py`. We no longer need a `debug` option in `Compare` and `FunctionComparator` to enable the debug dump, so this is removed.

As of #371, `--debug` is now used to display matching errors and other `logging.DEBUG` messages. Dumping the asm files is now enabled via the `--dump` option on `asmcmp.py`. A user may want to see matching errors but not create a lot of unneeded files.

The main loop on compared entities in `asmcmp.py` does a lot of work inline:
- Create the `ReccmpStatusReport`
- Prints the oneline diff summary of all entities
- Count aligned functions
- Compute raw accuracy and effective accuracy

Since all of these can be derived from the report data, I created new utility functions to do this. I moved the SVG progress function from #392 so they are now all neighbors in `report.py`. There are a few quick-n-dirty tests to demonstrate the behavior.

To make this possible I had to make a slight change to `ReccmpComparedEntity` that should be seamless. We had been storing `100.0` as the accuracy for effective matches, so the raw accuracy was lost. However, because we have always recorded `is_effective_match`, it's possible to use the raw accuracy instead and compute effective accuracy whenever it's needed.

### DIFF
--- a/reccmp/tools/asmcmp.py
+++ b/reccmp/tools/asmcmp.py
@@ -2,7 +2,6 @@
 
 from collections.abc import Sequence
 from datetime import datetime
-from typing import Callable
 import argparse
 import logging
 import os
@@ -24,6 +23,8 @@ from reccmp.compare.report import (
     ReccmpComparedEntity,
     deserialize_reccmp_report,
     serialize_reccmp_report,
+    report_function_alignment,
+    report_function_accuracy,
 )
 from reccmp.types import EntityType
 from reccmp.project.logging import argparse_add_logging_args, argparse_parse_logging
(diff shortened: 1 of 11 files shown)


reccmp/tools/asmcmp.py
disinvite: Scope creep already bit this PR, but maybe a function like `ReccmpComparedEntity.from_match()` is warranted? Mainly because the behavior is duplicated in `test_compare_output.py` and it should be in sync.

PR #405, opened 2026-05-04
Generate assembly and diff for stubs

disinvite:
Marking a function with `// STUB` designates it as incomplete. We handle stubs differently in these ways:
1. When importing to Ghidra, do not alter the function's parameters because they will probably change.
2. Do not include the stub's (low or zero percent) accuracy in the total accuracy report.
3. Do not generate assembly or diff the stub function.

This removes behavior 3. You can see the diff in when running `reccmp-reccmp` with `--verbose` and in the HTML view. The accuracy is still reported as `stub` without a specific percentage.

Draft while we consider the impact. One potential issue: if there is ever a crash while generating asm or doing the diff for a function, a user could work around it by adding `STUB`. With this change, the only recourse is to add it to the `ignored_functions` list in `reccmp-project.yml`.


jonschz:
Haven't thought about it for too long, but the idea sounds good to me. A few thoughts:
- Even if we don't count towards the total accuracy, we could still show the diff in the command output and HTML.
- I don't think that adding a `STUB` to work around a bug in the decompilation is a necessary feature. As far as I am concerned, we can just drop that and rely on bug reports to address such issues.
  - However, if you see this differently, I believe we shouldn't reuse the `ignore_functions` from the Ghidra config. We should either add a new config value (my preference) or move the existing one out of the `ghidra` config block (which I don't really like since it feels like mixing two different concepts - the Ghidra ignore block is intended to be used for functions that have a wrong, missing, or otherwise unusable signature in the PDB)

disinvite:
> Haven't thought about it for too long, but the idea sounds good to me. A few thoughts: ...

There's another `ignore_functions` list under the `report` key that we would use in the event of any problem.

I like to have a way for users to fix these things in the field so we're not under pressure to fast-track a patch.

reccmp/assets/components/listingTable.js
```diff
@@ -32,14 +32,17 @@ function createDiffRow(obj, showRecomp) {
...
-  } else if (diff.length === 0) {
-    contents = document.createElement('div');
-    contents.className = 'no-diff';
-    contents.textContent = 'Identical function - no diff';
+  if (diff.length === 0) {
+    // Display an appropriate message if there is no diff data.
+    if ('stub' in obj) {
+      contents = document.createElement('div');
+      contents.className = 'no-diff';
+      contents.textContent = 'Stub. No diff.';
```
jonschz: Maybe `Stub, but seems to match (no diff)`?

jonschz (approved):
Nice! This PR somehow flew under my radar.

reccmp/assets/components/listingTable.js
disinvite: This is now "'Identical function - no diff" or "No diff available" depending on the match percentage. I can't think of a way for the latter message to appear from normal operation (maybe with `reccmp-aggregate`?) but we're covered either way. I want to change this eventually because there isn't much value in opening the drawer with nothing in it.

jonschz (approved):

PR #406, opened 2026-05-04
Calculate max size for entities in one pass

disinvite:
Entities like functions, vtables, or variables may have different size in orig and recomp. Typically, we only have the size of the recomp entity because the PDB has the best information. Until #363, it was not possible to set an entity's size for only the orig address space; the matched entity used a single `size` attribute.

If we don't have the orig size, we can get an upper bound by measuring the distance from the entity to the next one in the address space, excluding entities like `OFFSET` or `LINE` that have no footprint.

It is faster and more straightforward to do this measurement in one pass instead of delaying until we need the information. We no longer need the `get_next_orig_addr` function so this is dropped.

This also adds the section ranges to the database. I considered using sentinel entities for this but they would collide with a real entity at the start of each section.


reccmp/compare/core.py
```diff
@@ -154,29 +156,45 @@ def run(self):
...
+            (ImageId.RECOMP, self.recomp_bin),
+        ):
+            # Some float consts may appear to be strings.
+            # Detect floats first because we can identify them with more confidence
+            # and this eliminates them from consideration as strings.
+            create_analysis_floats(self._db, img_id, binfile)
+            create_analysis_strings(self._db, img_id, binfile, self.bin_encoding)
+            complete_partial_floats(self._db, img_id, binfile)
+            complete_partial_strings(self._db, img_id, binfile, self.bin_encoding)
+
```
disinvite, line 204: Had to shuffle the order here because array offsets created by `match_array_elements` were colliding with the partial strings created by `create_analysis_strings`. Specifically: the string entity would have an incorrect `size` value. (As distinct from `max_size`.)

disinvite:
Finally done tinkering with this one. The `get_max_size()` function is intended for #432 to replace `get_next_orig_addr()`. No diffs in any of the decomp projects I checked.

madebr (approved):
I see times going from 73s to 13s for bigger projects, and no diff in the output of `reccmp-reccmp`.

reccmp/types.py
```diff
@@ -21,6 +22,30 @@ class EntityType(IntEnum):
     LABEL = 13
     OFFSET = 14
 
+    @classmethod
```
jonschz, line 25: What is the reason that you used a classmethod instead of a constant? Just curious

tests/test_compare_db.py
```diff
@@ -402,58 +402,107 @@ def test_generic_get_invalid_id(db: EntityDb):
...
+        batch.set(image_id, 100, type=EntityType.FUNCTION)
+        batch.set(image_id, 200)
+
+    # Ignored entity at addr=200. Calculated against the end of the section.
+    assert db.get_max_size(image_id, 100) == 200
 
-    assert db.get_next_orig_addr(0) == 100
-    assert db.get_next_orig_addr(100) == 200
-    assert db.get_next_orig_addr(200) is None
+    # Behavior differs from `set_max_size` because we do not require a specific
```
jonschz, line 445: I do not understand this comment

tests/test_compare_db.py
```diff
@@ -402,58 +402,107 @@ def test_generic_get_invalid_id(db: EntityDb):
...
+    # type for the entity or for the entity to exist at all.
+    assert db.get_max_size(image_id, 200) == 100
+    assert db.get_max_size(image_id, 250) == 50
 
 
-def test_get_next_orig_addr_no_type(db: EntityDb):
-    """Skip entities without a type."""
+@pytest.mark.parametrize("image_id", ImageId)
+def test_get_max_size_skip_non_compared_type(db: EntityDb, image_id: ImageId):
+    """When measuring max size, skip entities that are not a 'compared type'."""
```
jonschz:
```suggestion
    """When measuring max size, skip entities that are not a 'solid type'."""
```

tests/test_compare_db.py
```diff
@@ -402,58 +402,107 @@ def test_generic_get_invalid_id(db: EntityDb):
...
-    assert db.get_next_orig_addr(100) == 200
-    assert db.get_next_orig_addr(200) is None
+    # Behavior differs from `set_max_size` because we do not require a specific
+    # type for the entity or for the entity to exist at all.
+    assert db.get_max_size(image_id, 200) == 100
+    assert db.get_max_size(image_id, 250) == 50
 
 
-def test_get_next_orig_addr_no_type(db: EntityDb):
-    """Skip entities without a type."""
```
jonschz, line 434: Are the notypes dropped entirely? If so, cool! Otherwise, Id'd add one to the test with the lines

reccmp/compare/core.py
jonschz: Maybe add a comment why the order is significant?

jonschz (approved):
Very nice!

reccmp/types.py
disinvite: That was done just to avoid another import in the calling modules. Adding the set as a constant to the enum class seems like it would interfere with the individual enum values

tests/test_compare_db.py
disinvite: get_max_size reports the distance from the address to the next solid entity or section boundary. If there is an entity at the input address or not, the output is the same.
set_max_size is a batch operation on entities that exist, and here we do care about the entity’s type

The tests for get_max_size were copied from the tests for set_max_size so this is just calling attention to their differences

reccmp/compare/core.py
disinvite: Which order? The overall order of the tasks under run()?

tests/test_compare_db.py
jonschz: ahh, now I get it. Thanks, I didn't review it all in one go and didn't notice the similarities

reccmp/compare/core.py
jonschz: Yes, like the dependency you mentioned (first arrays, then strings?)

PR #413, opened 2026-05-14
Colorize the level name of log messages

madebr:
This change will cause the level name of log messages to be colorized.

`ERROR` in the following message will now be in red.
```
[ERROR] Failed to find function symbol with filename and line: /home/maarten/programming/racers/GolDP/src/boundingvolume0x64.cpp:28. If this issue persists after a recompile, the compiler has probably inlined this function.
```

color is limited to the level name as to not screw up colors in the messages.


disinvite (approved):
Nice! The only thing missing is that the `--no-color` argument (in various tool scripts) doesn't remove the colors. Any way to connect that to your recent change in #404?

madebr:
Can you re-review? I made color a global property, so you don't need to pass around a `plain` or `no_color` value.

reccmp/project/logging.py
```diff
@@ -14,4 +30,7 @@ def argparse_add_logging_args(parser: argparse.ArgumentParser):
 
 
 def argparse_parse_logging(args: argparse.Namespace):
+    if hasattr(args, "no_color"):
+        reccmp.color.enable_color(not args.no_color)
+        preconfigure_logging()
```
disinvite: We call `preconfigure_logging()` twice: once here and once in `main()` before `parse_args()`. If we run with `--no-color`, the ANSI chars are already embedded in the level string by the first calls to `logging.setLevelName`. They aren't removed by the second set of calls.

Would it work if we move `preconfigure_logging` outside the `if` and then not call it in each `main()` method?
```python
def argparse_parse_logging(args: argparse.Namespace):
    if hasattr(args, "no_color"):
        reccmp.color.enable_color(not args.no_color)
    preconfigure_logging()
```

PR #416, opened 2026-05-15
Overhaul `reccmp-decomplint` usage

### LINKED ISSUE #127: QOL features for `decomplint`

### LINKED ISSUE #400: Running reccmp-decomplint without path shows lots of false warnings
Runing `reccmp-decomplint --module LEGORACERS` (without paths) shows lots of `warning: duplicate_offset` warnings, even though these addresses are unique.

disinvite:
`decomplint` predates the idea of a `reccmp` project. You ran it by providing a list of paths to check and a "module", which is the identifier/namespace for the annotation. We now call these "targets".

First change in this PR: The difference between "module" and "target" is so small, I decided to just merge the command-line options for this script only. We can formally deprecate `--module` later if that makes sense, but I found at least one CI script using it.

There are two classes of errors you get from `decomplint`:
- **Syntax errors** reported by the parser. (For example: formatting on the annotation did not match the "strict" format.)
- **Target errors** reported by the linter. (For example: this address is out of order.)

With that distinction, here are the four ways to run `reccmp-decomplint`:

1. **With no arguments:** for each target in the project file, collect the paths and report all syntax errors and target errors scoped to the paths in each target.
2. **`--target` or `--module` option only**: same as 1 but for only a single target from the project file.
3. **List of paths:** syntax errors in those files only.
4. **List of paths plus `--target` or `--module`:** syntax errors from the given paths, target errors for the target provided.

Fixes #400. Fixes #127.


reccmp/parser/error.py
jonschz: Now that we are making changes here, how about addressing this as a drive-by? `ParserAlertCode` or `AlertCode` sounds fine to me

reccmp/parser/linter.py
```diff
@@ -1,62 +1,70 @@
...
+
+    return _filter
+
+
+def file_is_header(filename: PurePath) -> bool:
+    return filename.suffix.lower() in (".h", ".hpp")
+
+
+@dataclass
+class PerModuleData:
```
jonschz: Nitpick: How about `ModuleLinterData` or `PerModuleLinterData`? `PerModuleData` sounds too generic to me.

reccmp/parser/linter.py
```diff
@@ -1,62 +1,70 @@
...
+    addresses: set[int] = field(default_factory=set)
+    strings: dict[int, str] = field(default_factory=dict)
```
jonschz: How about a short docstring what the dict maps from and to, and what `strings` means here?

reccmp/parser/linter.py
```diff
@@ -1,62 +1,70 @@
...
 
             is_string = isinstance(marker, ParserString)
             is_folded = isinstance(marker, ParserFunction) and marker.is_folded
 
-            value = (marker.module, marker.offset)
-            if value in self._offsets_used:
+            if marker.module not in self._modules:
+                self._modules[marker.module] = PerModuleData()
+
+            module_data = self._modules[marker.module]
```
jonschz: Can maybe be done more elegantly using `module_data = self._modules.setdefault(marker.module, PerModuleData())`

reccmp/parser/linter.py
```diff
@@ -87,13 +96,12 @@ def _check_function_order(self):
         3. Functions marked FOLDED are ignored. The ordering is not well understood
         and you may not have much (any?) control over which footprint is used.
         """
-
-        if self._module is None:
+        if file_is_header(result.path):
```
jonschz: Why wasn't this necessary before?

reccmp/tools/decomplint.py
```diff
@@ -56,18 +62,22 @@ def parse_args() -> argparse.Namespace:
     parser.add_argument(
         "--version", action="version", version=f"%(prog)s {reccmp.VERSION}"
     )
+    # Combine --target and --module because they have the same goal:
+    # focusing on specific annotations to run order and uniqueness checks.
+    # If specific paths are provided, target defines which annotations to verify.
+    # If no paths are provided, use the path list for that target in the reccmp project.
```
jonschz: This comment is a little hard to read if you don't know the history behind it. I'd say something along the lines of "--module is a legacy option kept in for backwards compatibility. New projects should use --target".

reccmp/tools/decomplint.py
```diff
@@ -56,18 +62,22 @@ def parse_args() -> argparse.Namespace:
...
+    parser.add_argument(
+        "--target",
+        "--module",
+        metavar="<target-id>",
+        help="Run targeted checks on annotations for the given target.",
```
jonschz:
```suggestion
        help="Run checks on annotations for the given target only.",
```
maybe?

reccmp/tools/decomplint.py
```diff
@@ -77,7 +87,6 @@ def parse_args() -> argparse.Namespace:
     )
     parser.add_argument(
         "--encoding",
-        default="utf-8",
```
jonschz, line 80: Why was this default removed? Does that have any breaking consequences, especially without `required=False`?

Edit: I now saw where the default is actually set. I also realised that `required` seems to default to `False. Still, I'd prefer to set the default here in the parser and also add `(default 'utf-8')` to the description.

Edit2: I'd also mention the fallback to the target's encoding configured in the project files. Coming to think of it, do we even need this argument at all?

jonschz (commented):
Looks good! I haven't looked at `reccmp/tools/decomplint.py` in detail yet, here's the first batch of review comments.

reccmp/tools/decomplint.py
```diff
@@ -90,66 +99,123 @@ def parse_args() -> argparse.Namespace:
...
-            codefiles.extend(
-                TextFile.from_files(
-                    reduced_file_list,
-                    allow_error=True,
-                    encoding=target.encoding or "utf-8",
-                )
+    try:
+        lint_targets = decomplint_parse_args(args)
+    except RecCmpProjectException as e:
+        logger.error("%s", e.args[0])
```
jonschz:
```suggestion
        logger.error("%s", e.args[0], exc_info=e)
```

reccmp/tools/decomplint.py
```diff
@@ -90,66 +99,123 @@ def parse_args() -> argparse.Namespace:
...
+    paths: tuple[Path, ...]
+    module: str | None
+    encoding: str
+
+    def __init__(
+        self, paths: tuple[Path, ...], module: str | None, encoding: str | None
+    ):
+        self.paths = paths
+        self.module = module
+        self.encoding = encoding or "utf-8"
```
jonschz: If we move the default back up, this can be turned into a dataclass and the `__init__` can be dropped

reccmp/tools/decomplint.py
```diff
@@ -90,66 +99,123 @@ def parse_args() -> argparse.Namespace:
...
-                logger.error("Invalid path: %s", path)
-        reduced_file_list = source_code_search(search_paths)
-        codefiles = list(
-            TextFile.from_files(
-                reduced_file_list, allow_error=True, encoding=args.encoding
+
+        except UnicodeDecodeError:
+            total_alerts.setdefault(path, []).append(
+                ParserAlert(
+                    code=ParserError.UNICODE_DECODE_ERROR, line_number=-1, line=encoding
```
jonschz: Is `line=encoding` intended?

reccmp/tools/decomplint.py
```diff
@@ -90,66 +99,123 @@ def parse_args() -> argparse.Namespace:
...
-    warning_count, error_count = process_files(codefiles, module=args.module)
+    # Syntax errors from the parser: read once.
+    for (path, _), result in all_files.items():
+        total_alerts.setdefault(path, []).extend(result.alerts)
+
+    # Lint each grouping of files from each linter target.
+    for target in lint_targets:
+        linter = DecompLinter()
+        for path in target.paths:
+            if (path, target.encoding) in all_files:
```
jonschz: Is this if-check ever allowed to fail? If so, would an `assert` be more appropriate?

jonschz (approved):
I ran into one issue while trying this out and was wondering if this is intended behaviour.

Setup: I introduced a wrong order error in `LEGORacers/src/frostpetal0x34.cpp`.

* `reccmp-decomplint` (without arguments) finds the problem
* `reccmp-decomplint --target LEGORACERS` finds the problem
* `reccmp-decomplint --target GOLDP` does not find the problem (expected)
* `reccmp-decomplint --target LEGORACERS LEGORacers/src/frostpetal0x34.cpp` finds the problem
* `reccmp-decomplint LEGORacers/src/frostpetal0x34.cpp` does **not** find the problem, because no module is set.

The last one feels inconsistent with the zero-argument call to me: The zero-argument call sets all modules but the path-only call sets none. I'd find it more intuitive if either the path-only call would check all modules, or the zero-argument would check none.

Looks very good otherwise! Feel free to merge after addressing the comments.

disinvite:
@jonschz Many thanks for the review, all great points that I will address today or tomorrow.

reccmp/parser/linter.py
disinvite: Refactored out

reccmp/parser/linter.py
disinvite: The caller already checked if the file was a header, just moving it inside the function instead.

disinvite:
> I ran into one issue while trying this out and was wondering if this is intended behaviour.

This should now be fixed, but I blew up the files doing it so you should probably take a second look. 🙃

The individual checks are now pure functions. There are four, two that should be run once on every file being checked (across all targets for the no-args case):
- order of addresses for functions in non-headers (`check_function_order`)
- name-ref function annotations appearing in non-headers (`check_byname_allowed`)

These two need to be run for each collection of files in each target. To prevent duplicates, capture only the alerts that belong to the target of interest:
- do string annotations on the same address have the same text? (`check_string_text`)
- is each address used only once in the collection of files? (`check_offset_uniqueness`), excluding strings and folded functions

All the filtering happens in `decomplint.py` which should hopefully make the tests easier to follow. We don't need the statefulness provided by a `DecompLinter` object. We were also testing the `full_reset()` function that was never used.

reccmp/tools/decomplint.py
disinvite: I changed this around a little bit, but the check would fail if we cannot open the file (in the specified encoding) for some reason. The non-fatal reasons are `FileNotFoundError` or `UnicodeDecodeError`.

reccmp/tools/decomplint.py
disinvite: Yes, a small hack to display the failed encoding without adding another field. I added a comment saying why it's there.

disinvite:
Regarding the `--encoding` parameter: Should we allow the user to override the values set in `reccmp-project.yml`? Or should `--encoding` only work for the list of paths option? @jonschz @madebr

tests/test_parser_statechange.py
```diff
@@ -3,7 +3,7 @@
     ReaderState as _rs,
     DecompParser,
 )
-from reccmp.parser.error import ParserError as _pe
+from reccmp.parser.error import AlertCode as _pe
```
jonschz: Maybe rename or drop the alias? `_pe` doesn't fit anymore

reccmp/tools/decomplint.py
jonschz: This is resolved but neither applied nor rejected with a comment?

reccmp/tools/decomplint.py
```diff
@@ -169,52 +174,68 @@ def main():
...
+        ]
+
+        scoped_alerts = lint_file_collections(parsed_files, module=target.module)
+        all_alerts.extend(scoped_alerts)
 
     error_count = 0
     warning_count = 0
 
+    # Group alerts by path
+    total_alerts: dict[PurePath, list[ParserAlert]] = {}
```
jonschz: Maybe `filtered_alerts_by_path`? I couldn't tell the difference between `all_alerts` and `total_alerts` from their name

jonschz (approved):
Look good, I like the more functional approach.

Regarding `--encoding`: I'd go so far and say we don't need that argument anymore since we can specify an encoding in the project file. I'd deprecate the argument and print a warning that it will be removed in the future. I'm open to other opinions, though.

reccmp/tools/decomplint.py
disinvite: Guess I forgot to write something on that first one. I think the intent is to show the error message embedded in the `RecCmpProjectException` and not show the stack. It's the same in the other `tools/` scripts.

reccmp/parser/linter.py
disinvite: Refactored out

reccmp/tools/decomplint.py
disinvite: We hide the stack in all other scripts and just show the exception message before exiting.

disinvite:
To keep things simple I reverted to the `--encoding` behavior established in #364: `--encoding` applies to the `<paths>` option only. For `--target` with no `<paths>`, use the project encoding, or UTF-8 if there is none.

If there's a need for `--encoding` to override what's in the project file, we can add it later.

jonschz (approved):

PR #420, opened 2026-05-18
Fix IndexError in asm patch helpers when recomp shorter than orig

maci0:
## Summary

The `patch_compare_jmp`, `patch_mov_compare_jmp`, and `patch_fld_fmul`
helpers in `reccmp/compare/asm/fixes.py` find a target instruction
(`cmp`/`test`/`fld`) in the `orig` slice and then index `recomp` at the
same position without first checking that `recomp` is long enough.
When the recompiled disassembly window is shorter than the original
window — which happens in real comparisons when a function diverges
from the original near its end — this raises `IndexError` and aborts
the entire reccmp run.

Stack trace observed against an isledecomp/reccmp 0.1.1 wheel:

```
File ".../reccmp/compare/asm/fixes.py", line 217, in patch_compare_jmp
    not recomp[cmp_index].startswith(cmp_instruction)
        ~~~~~~^^^^^^^^^^^
IndexError: list index out of range
```

## Changes

- Add `cmp_index >= len(recomp) - 1` / `fld_index >= len(recomp) - 1`
  guards to the three affected helpers, matching the style already
  used in `patch_mov_commutative` (which has `inst_index >= len(recomp)`).
- Add three direct unit tests in `tests/test_fixes.py` that exercise
  each helper with a `recomp` slice shorter than `orig`. These tests
  fail with `IndexError` against current `master` and pass with the
  fix applied.

## Test plan

- [x] `pytest tests/test_fixes.py` passes (24 passed, 2 xfailed).
- [x] Reverting `reccmp/compare/asm/fixes.py` and re-running the new
      tests reproduces the original `IndexError` (3 failures).
- [x] End-to-end `reccmp-reccmp --target SERVER` run that previously
      aborted now completes against a real PE target.

### DIFF
--- a/tests/test_fixes.py
+++ b/tests/test_fixes.py
@@ -1,6 +1,11 @@
 import difflib
 import pytest
-from reccmp.compare.asm.fixes import find_effective_match
+from reccmp.compare.asm.fixes import (
+    find_effective_match,
+    patch_compare_jmp,
+    patch_fld_fmul,
+    patch_mov_compare_jmp,
+)
 
 
 def test_fix_cmp_jmp():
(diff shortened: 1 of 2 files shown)


tests/test_fixes.py
```diff
@@ -443,3 +448,28 @@ def test_and_swap_not_allowed():
...
     assert is_effective is False
+
+
+def test_patch_compare_jmp_recomp_shorter_than_orig():
+    """Regression: cmp_index found in orig must be bounds-checked against recomp
+    before indexing recomp[cmp_index]. Previously raised IndexError."""
+    orig = ["mov eax, 1", "mov ebx, 2", "cmp eax, ebx", "jg 0x1"]
+    recomp = ["mov eax, 1", "mov ebx, 2"]
+    assert patch_compare_jmp(orig, recomp, "cmp") == set()
+    assert patch_compare_jmp(orig, recomp, "test") == set()
```
disinvite, line 464: This should be its own test since there are no `test` instructions in the sample.

tests/test_fixes.py
```diff
@@ -443,3 +448,28 @@ def test_and_swap_not_allowed():
...
+
+
+def test_patch_mov_compare_jmp_recomp_shorter_than_orig():
+    orig = [
+        "mov eax, dword ptr [ebp-4]",
+        "cmp dword ptr [ebp-8], eax",
```
disinvite: To keep things consistent with the `capstone` asm rendering style used in the real data, there should be a space before and after the minus sign in these and other instructions. e.g.: `mov eax, dword ptr [ebp - 4]`.

disinvite (approved):
Looks good, thanks! Just a couple things to touch up in the tests and we are good to merge.

maci0:
updated

PR #425, opened 2026-05-20
Remove setuptools patch for capstone issue

disinvite:


jonschz:
Feel free to push this directly to #351

disinvite:
I haven't advanced this yet because I'm not sure how to properly verify that this is no longer needed. We currently test: `pip install .` from source, but we don't run a test build or run unit tests _using_ that build. Should we?

madebr:
The test workflow runs `pip install -r requirements.txt -r requirements-tests.txt` prior to running the unit tests `pytest . --binfiles=binfiles --require-binfiles`.

Is your concern about testing an installed reccmp vs testing reccmp from the source tree?
If so, I think this fine.

Though, for SDL releases we make sure to test all release (source and binary) artifacts.
So aAdding an extra step to the install workflow to test an installed reccmp would be useful (ignoring reccmp from the source tree). There's imho no need to test all python versions there.
(This is an extra feature, not belonging in this pr)

PR #432, opened 2026-05-24
Analyze and match CRT startup functions

### LINKED ISSUE #426: Handle dynamic initialization
C++ code like
```cpp
class FloatConstant {
public:
	FloatConstant(LegoFloat p_value) { m_value = p_value; }

	LegoFloat m_value;
};

const FloatConstant g_floatConst4096(4096.0f);
```
can cause the constructor to be run during [dynamic initialization](https://en.cppreference.com/cpp/language/initialization), i.e. between the entry point and `main()`. This is currently not covered by reccmp. Some unstructured thoughts on this issue:

- [ ] In our projects (`isle` and `racers`), the entry point functions have a false negative match - they report 100 % even though some of their calls (including those to the initializers) are not checked. This may have to do with relocations, which have less strict checks on offsets than other parts of the code.
- [ ] The initializers have MSVC function names of the form `$E<decimal number>`, which are not stable over time. Therefore, we will need to analyse the recompiled instructions for a full match.
- [x] `// GLOBAL` annotations don't work together with constructor invocations, see also https://github.com/isledecomp/racers/pull/326/changes#r3290550303
  - see #427

disinvite:
Fixes #426.

If the start and end labels for each of the four [CRT startup arrays](https://learn.microsoft.com/en-us/cpp/c-runtime-library/crt-initialization?view=msvc-170) are in the database, we will create function entities for each. For example, C++ initializers use the arrays `___xc_a` and `___xc_z`. The array begins with `0`, which we skip.

To match the functions between arrays, we need to figure out which variable is being initialized/destroyed. To do this, we read a snippet of code (64 bytes) and disassemble until we reach a `ret` instruction. From the disassembly, collect the addresses referenced in the function. Of those, we create a list of already matched entities identified as variables (`EntityType.DATA`). This is the function's "fingerprint".

For each fingerprint list that is not empty, match functions that are in the same CRT array in each binary.

### DIFF
--- a/reccmp/analysis/crt_startup.py
+++ b/reccmp/analysis/crt_startup.py
@@ -104,7 +104,7 @@ class UsedAddressCollector:
             if inst.mnemonic in JUMP_MNEMONICS:
                 continue
 
-            if inst.mnemonic in ("mov", "lea"):
+            if inst.mnemonic in ("mov", "fstp"):
                 dst_operand, _, src_operand = inst.op_str.partition(", ")
                 self._append_addrs(dst_operand, True)
                 self._append_addrs(src_operand, False)
--- a/reccmp/analysis/crt_startup.py
+++ b/reccmp/analysis/crt_startup.py
@@ -46,6 +46,9 @@ def get_crt_function_name(type_: CrtStartupArrayType) -> str:
     return _CRT_FUNCTION_NAMES[type_]
 
 
+UsedAddress = tuple[int, bool]
+
+
 @dataclass
 class CrtStartupArray:
     """Result from analyzing functions in a CRT startup array.
@@ -53,7 +56,7 @@ class CrtStartupArray:
     For example: addresses of C++ initializer functions are between
     the labels ___xc_a and ___xc_z."""
 
-    functions: dict[int, tuple[int, ...]]
+    functions: dict[int, tuple[UsedAddress, ...]]
     """Maps function address -> (sorted) list of matched entities used in
     the function, normalized to orig address space. These fingerprints are
     used to match initializer functions in orig and recomp."""
@@ -67,7 +70,7 @@ ADDR_REGEX = re.compile(r"0x[0-9a-f]{6,8}")
 
 
 class UsedAddressCollector:
-    seen_addrs: list[int]
+    seen_addrs: list[UsedAddress]
     """List of addrs that would be replaced by a name or placeholder."""
 
     is_entity: Callable[[int], bool]
@@ -76,6 +79,12 @@ class UsedAddressCollector:
     def __init__(self, is_entity: Callable[[int], bool]) -> None:
         self.is_entity = is_entity
         self.seen_addrs = []
+
+    def _append_addrs(self, text: str, is_write: bool):
+        for hex_str in ADDR_REGEX.findall(text):
+            addr = int(hex_str, 16)
+            if self.is_entity(addr):
+                self.seen_addrs.append((addr, is_write))
 
     def analyze(self, data: Buffer, start_addr: int):
         ig = InstructGen(bytes(data), start_addr, True)
@@ -95,10 +104,12 @@ class UsedAddressCollector:
             if inst.mnemonic in JUMP_MNEMONICS:
                 continue
 
-            for hex_str in ADDR_REGEX.findall(inst.op_str):
-                addr = int(hex_str, 16)
-                if self.is_entity(addr):
-                    self.seen_addrs.append(addr)
+            if inst.mnemonic in ("mov", "lea"):
+                dst_operand, _, src_operand = inst.op_str.partition(", ")
+                self._append_addrs(dst_operand, True)
+                self._append_addrs(src_operand, False)
+            else:
+                self._append_addrs(inst.op_str, False)
 
 
 def get_function_sample_size(db: EntityDb, image_id: ImageId, addr: int) -> int:
@@ -123,7 +134,7 @@ def get_function_sample_size(db: EntityDb, image_id: ImageId, addr: int) -> int:
 
 def get_function_fingerprint(
     db: EntityDb, image_id: ImageId, binfile: PEImage, addr: int
-) -> tuple[int, ...]:
+) -> tuple[UsedAddress, ...]:
     size = get_function_sample_size(db, image_id, addr)
     raw = binfile.read(addr, size)
 
@@ -134,14 +145,14 @@ def get_function_fingerprint(
     collector.analyze(raw, addr)
 
     normalized_addrs = []
-    for ca in collector.seen_addrs:
-        ent = db.get(image_id, ca)
+    for sample_addr, is_write in collector.seen_addrs:
+        ent = db.get(image_id, sample_addr)
         # Only matched entities are candidates for the fingerprint
         # because we have an address in both address spaces.
         if ent and ent.matched and ent.get("type") == EntityType.DATA:
             normalized_addr = ent.addr(ImageId.ORIG)
             assert isinstance(normalized_addr, int)
-            normalized_addrs.append(normalized_addr)
+            normalized_addrs.append((normalized_addr, is_write))
 
     return tuple(normalized_addrs)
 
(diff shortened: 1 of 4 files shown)


reccmp/compare/mutate.py
```diff
@@ -162,3 +164,8 @@ def unique_names_for_overloaded_functions(db: EntityDb):
                 batch.set(ImageId.ORIG, func.orig_addr, computed_name=new_name)
             elif func.recomp_addr is not None:
                 batch.set(ImageId.RECOMP, func.recomp_addr, computed_name=new_name)
+
+
+def match_crt_startup(db: EntityDb, orig_bin: PEImage, recomp_bin: PEImage):
+    # Alias to keep this module's scope small
+    _match_crt_startup(db, orig_bin, recomp_bin)
```
jonschz: This alias feels a bit unexpected to me. Do you have any architectural pattern in mind here why you don't want to make `_match_crt_startup` public and use it directly?

reccmp/analysis/crt_startup.py
```diff
@@ -0,0 +1,290 @@
...
+)
+
+
+_CRT_STARTUP_ARRAY_BOUNDARIES = {
+    CrtStartupArrayType.C_INIT: ("___xi_a", "___xi_z"),
+    CrtStartupArrayType.CPP_INIT: ("___xc_a", "___xc_z"),
+    CrtStartupArrayType.C_PRE_TERM: ("___xp_a", "___xp_z"),
+    CrtStartupArrayType.C_TERM: ("___xt_a", "___xt_z"),
+}
+
```
jonschz: minor tweak:

```suggestion
_CRT_STARTUP_ARRAY_BOUNDARIES = {
    CrtStartupArrayType.C_INIT: ("___xi_a", "___xi_z"),
    CrtStartupArrayType.CPP_INIT: ("___xc_a", "___xc_z"),
    CrtStartupArrayType.C_PRE_TERM: ("___xp_a", "___xp_z"),
    CrtStartupArrayType.C_TERM: ("___xt_a", "___xt_z"),
}

_CRT_STARTUP_ARRAY_LABELS = [
    label for pair in _CRT_STARTUP_ARRAY_BOUNDARIES.values() for label in pair
]

```
The benefit is a single source of truth

reccmp/analysis/crt_startup.py
```diff
@@ -0,0 +1,290 @@
...
+                break
+
+
+def get_function_fingerprint(
+    db: EntityDb, image_id: ImageId, binfile: PEImage, addr: int
+) -> tuple[int, ...]:
+    # 64 bytes chosen arbitrarily.
+    # These functions are typically short, and we only need
+    # to read enough to create the fingerprint.
+    raw = binfile.read(addr, 64)
```
jonschz: How about a lookup until the next known symbol? I think these initializers can be arbitrarily long because they sometimes inline constructors.

reccmp/analysis/crt_startup.py
```diff
@@ -0,0 +1,290 @@
...
+        instructions = (
+            inst
+            for section in ig.sections
+            for inst in section.contents
+            if section.type == SectionType.CODE
+        )
+
+        for inst in instructions:
+            assert isinstance(inst, DisasmLiteInst)
+            if "0x" in inst.op_str and (
```
jonschz: I think we may have to filter for write instructions here. An initializer may read from one dynamically initialized variable and write to another one (example: `LEGORACERS 0x0044a740`, reads from `g_floatConst4096` and writes to `g_token0x0fconstant`).

reccmp/analysis/crt_startup.py
```diff
@@ -0,0 +1,290 @@
...
+    return CrtStartupArray(fingerprints, thunks)
+
+
+def analyze_crt_startup(
+    db: EntityDb, image_id: ImageId, binfile: PEImage
+) -> Iterator[tuple[CrtStartupArrayType, CrtStartupArray | None]]:
+    """For the CRT startup arrays in the given binary, if the start and end labels are known,
+    analyze the functions in each array."""
+    labels = find_crt_startup_labels(db, image_id)
+    for array_type, (label_start, label_end) in _CRT_STARTUP_ARRAY_BOUNDARIES.items():
```
jonschz, line 240: Bonus feature request (doesn't have to come in this PR): If some labels are found in recomp but not in orig, a warning should be emitted. Otherwise, `reccmp` will just silently skip this comparison. The warning could also link to Microsoft's documentation.

reccmp/analysis/crt_startup.py
```diff
@@ -0,0 +1,290 @@
...
+                # and collect the addresses.
+                self.sanitize(inst)
+
+            # The functions we are looking at should not have complex logic
+            # that creates multiple exits.
+            if inst.mnemonic == "ret":
+                break
+
+
+def get_function_fingerprint(
```
jonschz, line 133: I'd add a docstring explaning what the fingerprint is

reccmp/analysis/crt_startup.py
```diff
@@ -0,0 +1,290 @@
...
+    for ca in collector.seen_addrs:
+        ent = db.get(image_id, ca)
+        # Only matched entities are candidates for the fingerprint
+        # because we have an address in both address spaces.
+        if ent and ent.matched and ent.get("type") == EntityType.DATA:
+            normalized_addr = ent.addr(ImageId.ORIG)
+            assert isinstance(normalized_addr, int)
+            normalized_addrs.append(normalized_addr)
+
+    return tuple(normalized_addrs)
```
jonschz, line 159: I'd log a warning if we find
- 0 matches for a recomp function
- more than 1 match for an orig or recomp function
since either case indicates that our analysis cannot be trusted (0 for orig is fine since the variable might not yet be matched). With the current logic (that counts both reads and writes), my statistics is as follows in RACERS:

orig: 22 x 0 fingerprints, 4 x 1 fingerprint, 3 x 2 fingerprints
recomp: 0 x 0 fingerprints, 4 x 1 fingerprint, 3 x 2 fingerprints

reccmp/analysis/crt_startup.py
```diff
@@ -0,0 +1,290 @@
...
+def unwrap_jump(binfile: PEImage, addr: int) -> tuple[bool, int]:
+    """If there is a 5-byte JMP or CALL instruction at the given address,
+    follow it by calculating the destination address.
+    Returns either (True, jmp_destination) or (False, starting_addr)."""
+    jmp = binfile.read(addr, 5)
+    # Check for CALL (0xE8) or JMP (0xE9) opcodes.
+    if jmp[0] in (0xE8, 0xE9):
+        (offset,) = struct.unpack("<i", jmp[1:])
+        # Follow the jump only if it is small.
+        if abs(offset) <= 16:
```
jonschz: This feels a bit arbitrary. What is the reason here?

reccmp/analysis/crt_startup.py
```diff
@@ -0,0 +1,290 @@
...
+        (fp, addr) for addr, fp in orig_array.functions.items() if fp is not None
+    )
+    invert_recomp = dict(
+        (fp, addr) for addr, fp in recomp_array.functions.items() if fp is not None
+    )
+
+    matches = []
+
+    for fingerprint, orig_addr in invert_orig.items():
+        if fingerprint in invert_recomp:
```
jonschz: I'd log a warning in the `else` case. Also, should we check for fingerprints that are in recomp but not in orig? It may indicate that the recomp has an initializer that isn't supposed to be there.

reccmp/analysis/crt_startup.py
```diff
@@ -0,0 +1,290 @@
...
+        was_thunk, real_addr = unwrap_jump(binfile, xc_addr)
+        fp = get_function_fingerprint(db, image_id, binfile, real_addr)
+        fingerprints[real_addr] = fp
+        if was_thunk:
+            thunks[real_addr] = xc_addr
+
+    return CrtStartupArray(fingerprints, thunks)
+
+
+def analyze_crt_startup(
```
jonschz: Nitpick: I'd go for a name like `detect_crt_startup_arrays`. `Analyse` sounds to me like it should act via side effect and not return anything. Applies to other `analyze_` functions around here as well, at least to a degree.

tests/test_analysis_crt_startup.py
```diff
@@ -0,0 +1,122 @@
...
+    result = analyze_crt_startup_functions(db, ImageId.ORIG, binfile, XCA_XCZ_RANGE)
+    assert result.functions[0x1001A6D0] == (0x10102B28,)
+
+
+def test_xca_fingerprints_avoid_crash(binfile: PEImage):
+    db = EntityDb()
+    # Misaligned end address will cause struct.iter_unpack to raise struct.error.
+    modified_range = range(XCA_XCZ_RANGE.start, XCA_XCZ_RANGE.stop - 1)
+
+    try:
```
jonschz, line 125: I'd remove this `try-except`. If you really want to keep it, change it to `except Exception:` so it allows e.g. `KeyboardInterrupt` to pass

jonschz (commented):
This is very nice, very much looking forward to this being merged! I'm especially happy that this works without adding a new annotation type.

reccmp/analysis/crt_startup.py
```diff
@@ -0,0 +1,290 @@
...
+    """Maps function address -> (sorted) list of matched entities used in
+    the function, normalized to orig address space. These fingerprints are
+    used to match initializer functions in orig and recomp."""
+
+    thunks: dict[int, int]
+    """Maps thunked initializer function to the thunk address.
+    The thunk is what actually appeared in the ___xc_a/z list."""
+
+
+class UsedAddressCollector(ParseAsm):
```
jonschz: I've struggled with forming an opinion here (you may have seen the comment I accidentally submitted :D ). 

I think there isn't enough shared logic between `ParseAsm` and this for inheritance to be a good fit. The code will be sufficiently different that I think a slight amount of duplication is a better solution here. This will especially be true if we filter for write statements.

reccmp/analysis/crt_startup.py
disinvite: There was at least one function in ~~either `LEGORACERS` or `GOLDP`~~ where the pattern looked like
```asm
mov edi, address
call constructor
```
**Edit** It was the one I used in the unit test: `LEGO1`, `0x1001a6d0`.

reccmp/analysis/crt_startup.py
disinvite: Just by observation. In `LEGO1` the thunking was consistent: always a `jmp` and always 16 bytes offset from the start of the function.

I don't have it up in front of me, but I think it was `GOLDP` that had `call` instructions acting as thunks, but also init functions that began with a `call` that was _not_ intended as a thunk. So it's a question of what functions (and how many) to read and establish the identifying characteristics.

reccmp/analysis/crt_startup.py
disinvite: Yeah, I changed the names for these a few times; nothing seemed to fit perfectly.

tests/test_analysis_crt_startup.py
disinvite: Yeah, not sure why I didn't just use `struct.error` here since that's the one we're trying to avoid.

reccmp/analysis/crt_startup.py
disinvite: I think we could cover this by showing the CRT arrays (and their diff) in the report as if they were vtables. I'm not sure how much it would take to add this, but if it requires too many changes then I'll delay it until a follow-up.

reccmp/analysis/crt_startup.py
disinvite: Yeah, we could do that. The exact size may already be known (via PDB or CSV), so we should use it if it's there.

reccmp/analysis/crt_startup.py
disinvite: This was surely a hack to get us up and running. To start with, since we're already limiting the fingerprint output to entities that exist, we don't really need the "address test" function.

reccmp/analysis/crt_startup.py
jonschz: Maybe we could filter for write instructions in case we find multiple addresses? That heuristic could be pretty robust

reccmp/analysis/crt_startup.py
jonschz: Yeah, and even if not, reading too much would only cost a bit of performance at worst

reccmp/analysis/crt_startup.py
```diff
@@ -103,10 +104,12 @@ def analyze(self, data: Buffer, start_addr: int):
             if inst.mnemonic in JUMP_MNEMONICS:
                 continue
 
-            for hex_str in ADDR_REGEX.findall(inst.op_str):
-                addr = int(hex_str, 16)
-                if self.is_entity(addr):
-                    self.seen_addrs.append(addr)
+            if inst.mnemonic in ("mov", "lea"):
```
jonschz: I think there's also an FSTP in RACERS

Edit: Indeed there is, see `LEGORACERS 0x0044a740`.

reccmp/analysis/crt_startup.py
jonschz: Fair enough. I suggest we move it into a constant and add a docstring below with a brief explanation of the heuristic involved. Maybe something like
```py
INITIALIZER_THUNK_MAX_JUMP_OFFSET = 16
"""Some MSVC dynamic initializers are thunked, and we have observed that the offset is usually equal to 16. Tweak this value if necessary."""
```

reccmp/analysis/crt_startup.py
disinvite: `lea` may also be misplaced here

reccmp/analysis/crt_startup.py
disinvite: I think the second point is covered by simply not matching unless it can be done uniquely. For the first point, since we know the exact size of the arrays in both orig and recomp, if you have extra functions in recomp, the diff will be visible once we display the table in the report (#440).

jonschz: Fine with me 👍

PR #435, opened 2026-05-28
Allow hex numbers in `size` field in CSV parser

InvisibleUp:
I've added a super simple patch that allows the CSV parser to additionally parse sizes as `0x`-prefixed hex numbers. I've also documented the `size` field, since that was missing.

### DIFF
--- a/docs/csv.md
+++ b/docs/csv.md
@@ -129,5 +129,6 @@ These fields can appear in the CSV file:
 | `name` | Name for the entity (e.g. the fully-qualified function name) | text |
 | `symbol` | Linker name | text |
 | `type` | Entity type | one of: `function, template, synthetic, library, stub, global, string, widechar, float, vtable` |
+| `size` | Size of the entity in bytes | Decimal number, or hex number with `0x` prefix |
 
 All other fields are ignored.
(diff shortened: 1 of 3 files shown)


jonschz:
@disinvite I remember that we previously had a debate on decimal vs. hex in CSV. Do you remember the outcome?

disinvite:
> @disinvite I remember that we previously had a debate on decimal vs. hex in CSV. Do you remember the outcome?

Yes, it was about the `address` field. My position at the time (and the current behavior) is that `address` is always hex and so the `0x` prefix is not required. That's probably something we should change.

@InvisibleUp I think something like this would be better than using `base=0`:
```python
def decimal_or_hex(value: str) -> int:
    # Allow 0x or 0X prefix
    is_hex = value.lower().startswith("0x")
    base = 16 if is_hex else 10
    try:
        return int(value, base=base)
    except ValueError as ex:
        # TODO: Should be its own exception class
        raise ReccmpCsvParserError(value) from ex
```
and
```python
        if key == "size":
            output["size"] = decimal_or_hex(value)
```
We should also have a few tests that show the new parsing behavior and where exceptions are thrown. I can add the ones I have if you want, up to you.

InvisibleUp:
Ah, right, you'd probably want to allow leading zeroes too. I believe `output["size"] = int(value.lstrip('0'), 0)` would also work. (I suppose that would also allow octal or binary formats, which is a bit silly, but harmless imo.)

Ultimately I just want to use hex numbers for the length field, since that's a lot easier to get out of IDA than decimal. However you do that is up to you.

PR #437, opened 2026-05-30
Parse LF_BITFIELD leaf type

### LINKED ISSUE #436: Matching a global variable that contains an array of structs with bitfields throws a fatal error
Reproducer:
```c++
struct Cls
{
    unsigned char f0: 1;
};

// GLOBAL: AA 0x10008e08
Cls objects[4];

__declspec(dllexport) Cls* get_obj(int i);

// FUNCTION: AA 0x10001010
Cls* get_obj(int i)
 {
  return &objects[i];
}
```

project: [global-with-array-of-bitfield.zip](https://github.com/user-attachments/files/28411447/global-with-array-of-bitfield.zip)

madebr:
Fixes #436

### DIFF
--- a/tests/test_compare_mutate_match_array.py
+++ b/tests/test_compare_mutate_match_array.py
@@ -562,3 +562,82 @@ def test_match_array_array_of_union_structs(db: EntityDb, types_db: CvdumpTypesP
     assert not any(
         db.get(ImageId.RECOMP, addr) for addr in (101, 102, 103, 105, 106, 107)
     )
+
+
+def test_match_array_of_struct_bitfield(db: EntityDb, types_db: CvdumpTypesParser):
+    """Verify comparing an array of struct bitfields uses the underlying type of the bitfield"""
+    types_db.keys[TK(0x1000)] = {
+        "type": "LF_ARRAY",
+        "array_type": TK(0x1001),
+        "size": 4 * 8,
+    }
+    types_db.keys[TK(0x1001)] = {
+        "type": "LF_STRUCTURE",
+        "field_list_type": TK(0x1002),
+        "size": 8,
+    }
+    types_db.keys[TK(0x1002)] = {
+        "type": "LF_FIELDLIST",
+        "members": [
+            FieldListItem(offset=0, name="v0", type=CVInfoTypeEnum.T_UINT4),
+            FieldListItem(offset=4, name="bit0", type=TK(0x1003)),
+            FieldListItem(offset=4, name="bit1", type=TK(0x1004)),
+        ],
+    }
+    types_db.keys[TK(0x1003)] = {
+        "type": "LF_BITFIELD",
+        "bit_start": 0,
+        "bit_count": 1,
+        "bit_type": CVInfoTypeEnum.T_UCHAR,
+    }
+    types_db.keys[TK(0x1004)] = {
+        "type": "LF_BITFIELD",
+        "bit_start": 1,
+        "bit_count": 1,
+        "bit_type": CVInfoTypeEnum.T_UCHAR,
+    }
+
+    with db.batch() as batch:
+        batch.set(
+            ImageId.RECOMP,
+            0x100,
+            name="test",
+            type=EntityType.DATA,
+            data_type=0x1000,
+            size=32,
+        )
+        batch.match(0x100, 0x100)
+
+    match_array_elements(db, types_db)
+
+    e = db.get(ImageId.ORIG, 0x100)
+    assert e is not None
+    assert e.name == "test[0].v0"
+    assert e.get("type") == EntityType.DATA
+    assert e.any_size() == 32
+
+    e = db.get(ImageId.RECOMP, 0x118)
+    assert e is not None
+    assert e.name == "test[3].v0"
+    assert e.get("type") == EntityType.OFFSET
+    assert e.any_size() == 4
+
+    e = db.get(ImageId.RECOMP, 0x11C)
+    assert e is not None
+    assert e.name == "test[3].bit0"
+    assert e.get("type") == EntityType.OFFSET
+    assert e.any_size() == 1
+
+    assert not any(
+        db.get(ImageId.ORIG, addr) for addr in range(0x100, 0x120) if addr % 4 != 0
+    )
+    assert all(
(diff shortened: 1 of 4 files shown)


tests/test_compare_mutate_match_array.py
```diff
@@ -562,3 +562,83 @@ def test_match_array_array_of_union_structs(db: EntityDb, types_db: CvdumpTypesP
...
+    assert e.get("type") == EntityType.OFFSET
+    assert e.any_size() == 4
+
+    e = db.get(ImageId.RECOMP, 0x11C)
+    assert e is not None
+    assert e.name == "test[3].bit0"
+    assert e.get("type") == EntityType.OFFSET
+    assert e.any_size() == 1
+
+    # FIXME: add identical test for ImageId::ORIG
```
madebr: I've added 2 fixme's here.
The matcher does not add items for the int and bitfields of the original image, as it does for the recompiled one.

tests/test_compare_mutate_match_array.py
```diff
@@ -562,3 +562,83 @@ def test_match_array_array_of_union_structs(db: EntityDb, types_db: CvdumpTypesP
...
+    assert e.name == "test[0].v0"
+    assert e.get("type") == EntityType.DATA
+    assert e.any_size() == 4
+
+    # FIXME: should also access bit0 of original?
+    # e = db.get(ImageId.ORIG, 0x104)
+    # assert e is not None
+    # assert e.name == "test[0].f0"
+    # assert e.get("type") == EntityType.OFFSET
+    # assert e.any_size() == 8
```
disinvite: This sets name to `test[0].bit0` and size to 1, which seems fine?

tests/test_compare_mutate_match_array.py
```diff
@@ -562,3 +562,83 @@ def test_match_array_array_of_union_structs(db: EntityDb, types_db: CvdumpTypesP
...
+    }
+
+    with db.batch() as batch:
+        batch.set(
+            ImageId.RECOMP,
+            0x100,
+            name="test",
+            type=EntityType.DATA,
+            data_type=0x1000,
+            size=4,
```
disinvite: Should be 32 to be more realistic to the array size. In practice, it doesn't matter because `match_array` ignores the entity size and uses the data type's size.

tests/test_cvdump_types.py
```diff
@@ -1195,3 +1195,23 @@ def test_unknown_primitive_type(empty_parser: CvdumpTypesParser):
...
+0x1002 : Length = 10, Leaf = 0x1205 LF_BITFIELD
+        bits = 1, starting position = 0, Type = T_UCHAR(0020)
+
+0x1003 : Length = 10, Leaf = 0x1205 LF_BITFIELD
+        bits = 3, starting position = 6, Type = T_UINT4(0075)
+"""
+
+
+def test_bitfields(empty_parser: CvdumpTypesParser):
+    """Make sure we can read an LF_FIELDLIST of an enum with a negative value (GH #380)"""
```
disinvite: Docstring copied from an earlier function

tests/test_compare_mutate_match_array.py
madebr: I see `assert e is not None` failing. There are no entities for the original image (apart from one at offset 0x0).

tests/test_compare_mutate_match_array.py
madebr: This change causes the assertion at line 617 to fail:
```
    e = db.get(ImageId.ORIG, 0x100)
    assert e is not None
    assert e.name == "test[0].v0"
    assert e.get("type") == EntityType.DATA
    assert e.any_size() == 4   # <<== This one fails
```

Is this related to my question about not creating offset entities for the original image?

tests/test_compare_mutate_match_array.py
```diff
@@ -562,3 +562,83 @@ def test_match_array_array_of_union_structs(db: EntityDb, types_db: CvdumpTypesP
...
+        )
+        batch.match(0x100, 0x100)
+
+    match_array_elements(db, types_db)
+
+    e = db.get(ImageId.ORIG, 0x100)
+    assert e is not None
+    assert e.name == "test[0].v0"
+    assert e.get("type") == EntityType.DATA
+    assert e.any_size() == 4
```
disinvite: I think the failure is from this line: it should also be 32 to match the full size of the array variable on line 607.

madebr: Indeed, but 2 lines above it says `assert e.name == "test[0].v0"`, and `v0` is 4 bytes.

I applied your suggestion.

madebr: There being only one entity in the original binary, explains my FIXME's further down this test.

disinvite: > Indeed, but 2 lines above it says `assert e.name == "test[0].v0"`, and `v0` is 4 bytes.
> 
> I applied your suggestion.

Yes, this is a side effect of having only one entity per address. The main address for the array variable represents the total size of the array. The newly created entities (with type `EntityType.OFFSET`) get assigned a size for whatever is at that array element or struct member, according to the datatype used. All entities, including the main entity, are renamed with the struct/array offset name, up to two levels of depth.

That's why I wasn't sure what there is to fix about these lines:
```python
# FIXME: should also access bit0 of original?
e = db.get(ImageId.ORIG, 0x104)
assert e is not None
assert e.name == "test[0].bit0"
assert e.get("type") == EntityType.OFFSET
assert e.any_size() == 1
```
because they're working the way I would expect.

madebr: Having one entity is fine. But I have a question about the name: `"test[0].v0"`. Shouldn't it be `test` then?

disinvite: Sure, but the same name would be substituted in both of these situations:
```cpp
g_variable[0].first_offset = 5;

memcpy(&g_other, &g_variable, sizeof(g_variable));
```
`g_variable` and `g_variable[0].first_offset` are both "correct" but not perfect. Maybe we could do something by inspecting the type of instruction used.

madebr: I suppose this is nice to have, but not a deal breaker.

disinvite (approved):

PR #441, opened 2026-06-03
roadmap: output numbers as hexadecimals

madebr:
`reccmp-roadmap --target GOLDP --csv symbols.csv` is currently outputing addresses in decimal format.

This pr changes the default to hexadecimal (with `0x` prefix),
~and adds a `--decimal` cli argument for decimals (=current behavior)~.

This pr also contains a small change to not hardcode the csv header row.
[`namedtuple._fields` is a documented member](https://docs.python.org/3/library/collections.html#collections.somenamedtuple._fields).

### DIFF
--- a/reccmp/tools/roadmap.py
+++ b/reccmp/tools/roadmap.py
@@ -6,12 +6,12 @@ report how "far off" the recomp symbol is from its proper place
 in the original binary.
 """
 
-import os
 import argparse
+import bisect
 import logging
+import os
 from pathlib import Path
 import statistics
-import bisect
 from typing import Iterator, NamedTuple
 import reccmp
 from reccmp.formats import PEImage
@@ -35,6 +35,8 @@ logger = logging.getLogger(__name__)
 
 def or_blank(value) -> str:
     """Helper for dealing with potential None values in text output."""
+    if isinstance(value, int):
+        return hex(value)
     return "" if value is None else str(value)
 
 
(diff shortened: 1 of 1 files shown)


jonschz:
Do we even need the option to switch to decimal? @disinvite what do you think? I'd personally go for ~~decimal~~ hexacedimal only

madebr:
> Do we even need the option to switch to decimal? @disinvite what do you think? I'd personally go for decimal only

Hexadecimal or decimal? (Because decimal means no change)

I'm also in favor of removing the argument.

jonschz:
Ouch, typo. Fixed.

disinvite:
> Do we even need the option to switch to decimal? @disinvite what do you think? I'd personally go for ~decimal~ hexacedimal only

For addresses, yes. Hex by default, and no compelling argument for decimal. If you wanted to import to Excel/OpenOffice/Google Sheets for some reason, the values would still be sortable as a hex string.

For other numeric values like entity size and (specific to `roadmap`) the difference between orig and recomp addresses, I think decimal is better. But... I don't feel strongly either way.

jonschz:
Personally, I am more worried about feature creep than having the output in the perfect form. Let's go for addresses in hex only and leave the rest as decimal for now, and no additional argument to influence the output format.

madebr:
Everything is now hexadecimal, including sizes.
The isle and racers decomps also encodes its sizes as hexadecimals.

jonschz (approved):

disinvite (approved):

madebr:
I did a late commit message change (hexadecimals are used everywhere, and the option is gone)

madebr:
Oh, the repo does not do rebases or merge commits, only squashes. Then this was not really required :)

PR #442, opened 2026-06-05
Avoid divide by zero in asmcmp summary

madebr:
Fixes this error:
```
  File "/home/maarten/.local/bin/reccmp-reccmp", line 6, in <module>
    sys.exit(main())
             ~~~~^^
  File "/home/maarten/projects/reccmp/reccmp/tools/asmcmp.py", line 307, in main
    effective_accuracy = total_effective_accuracy / implemented_funcs * 100
                         ~~~~~~~~~~~~~~~~~~~~~~~~~^~~~~~~~~~~~~~~~~~~
ZeroDivisionError: division by zero
```

### DIFF
--- a/reccmp/tools/asmcmp.py
+++ b/reccmp/tools/asmcmp.py
@@ -13,6 +13,7 @@ from reccmp.utils import (
     print_combined_diff,
     diff_json,
     percent_string,
+    safe_denominator,
     write_html_report,
 )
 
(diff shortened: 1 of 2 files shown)


reccmp/tools/asmcmp.py
```diff
@@ -302,33 +308,37 @@ def main() -> int:
         # Use the alternate value if it exceeds the number of annotated functions
         function_count = max(function_count, int(args.total))
 
-    if function_count > 0:
-        implemented = implemented_funcs / function_count * 100
-        effective_accuracy = total_effective_accuracy / implemented_funcs * 100
-        # actual_accuracy = total_accuracy / implemented_funcs * 100
-        progress = total_effective_accuracy / function_count * 100
-        alignment_percentage = functions_aligned_count / function_count * 100
+    implemented = implemented_funcs / function_count * 100
```
disinvite: May as well add it here too, although practically speaking we should always have the `entry` function

disinvite (approved):

reccmp/tools/asmcmp.py
```diff
@@ -302,33 +308,37 @@ def main() -> int:
...
-        effective_accuracy = total_effective_accuracy / implemented_funcs * 100
-        # actual_accuracy = total_accuracy / implemented_funcs * 100
-        progress = total_effective_accuracy / function_count * 100
-        alignment_percentage = functions_aligned_count / function_count * 100
+    implemented = implemented_funcs / safe_denominator(function_count) * 100
 
+    effective_accuracy = (
+        total_effective_accuracy / safe_denominator(implemented_funcs) * 100
+    )
+    # actual_accuracy = total_accuracy / safe_denominator(mplemented_funcs) * 100
```
jonschz: typo. Also, for how long has this been commented out? Can we just remove it?

reccmp/tools/asmcmp.py
```diff
@@ -302,33 +308,37 @@ def main() -> int:
...
-                total_effective_accuracy,
-            )
+    if args.svg is not None:
+        gen_svg(
+            args.svg,
+            os.path.basename(target.original_path),
+            args.svg_icon,
+            implemented_funcs,
+            function_count,
+            total_effective_accuracy,
```
jonschz, line 334: good catch

reccmp/tools/asmcmp.py
disinvite: Only since #409 I think. I should've just removed it then.

reccmp/tools/asmcmp.py
madebr: Good point. `gen_svg` also throws.

reccmp/tools/asmcmp.py
madebr: I removed this comment.

reccmp/tools/asmcmp.py
madebr: `-S image.svg` is now "safe".

reccmp/tools/asmcmp.py
madebr: This has been added.

Issue #123, opened 2025-05-27
Custom annotation format


disinvite:
Reccmp supports code annotations in this format:

```
// [type]: [module] [addr] [extra?]
```

`extra` is the only optional element. We also enforce style rules, so users of `decomplint` are alerted to departures from this format:

```
// [type-upper]: [module-upper] [0x+addr-lower] [extra?]
```

where exactly one space separates each element.

This is reasonable but probably not what everyone wants. I've been kicking around the idea of supporting a [strptime-like format](https://docs.python.org/3/library/datetime.html#format-codes) so your project could do annotations your way. It shouldn't be too difficult to build the regular expression from the codes, so the real work is deciding on:

- Which options to support?
- How do you express those options as codes?
- Is addr the only required element?

jonschz:
To add to this, at least one user (@ifarbod) is interested in aliases for the different annotations (e. g. `FUNC` or `fn` instead of `FUNCTION`, see https://github.com/isledecomp/reccmp/issues/85#issuecomment-2877115218). 

It's the classic "convention vs configuration" argument. Just to play devil's advocate, there are some formats people *could* ask for that I wouldn't consider supporting, like annotations in block arguments, or annotations located at the end of the annotated line instead of being located in the previous line.

Personally, I'd consider supporting aliases (not sure if configurable or a few hard-coded ones) and wait regarding the flexible format until someone comes along and asks for it. Since it has to be one line and all parts are mandatory, the options are quite limited. Also, it has to be sufficiently unique/rigid in order to tell it apart from other comments. Case sensitivity settings is something else we may want to consider supporting. Curious to hear some more opinions before we make a decision.

ifarbod:
Here's a more relaxed and friendly version for your GitHub issue response:

---

I'm actually using reccmp in a medium-sized project alongside another assembly differ tool - running both helps me catch mistakes more efficiently.  

Right now, I'm using the standard format (lowercase `func` and `var` + the unchanged `stub`, `synthetic` , `vtable` and `library`), and from my experience, reliably parsing comments is tricky (maybe something like `libclang` could help here?).  

For `reccmp-project.yml`, it’d be great to support extra aliases like:  

```yaml
case_sensitive: false
...
FUNCTION: [FN, FUN, FUNC]  
GLOBAL: VAR  
```

ifarbod:
I'm actually using reccmp in a medium-sized project alongside another assembly differ tool - running both helps me catch mistakes more efficiently.  

[...]

PR #443, opened 2026-06-06
Code annotation aliases

disinvite:
Adds support for aliases of annotation types, effectively providing custom annotations.
e.g.
```cpp
// func: test 0x123456
```
will work with
```yml
targets:
  test:
    # ...
    marker-aliases:
      func: FUNCTION
```

Aliases must follow these rules:
1. They must point directly to a built-in type. Double aliases are not evaluated.
2. They must be unique (case-insensitive) among the set. Only the first is used if there are duplicates. You cannot have `func` and `Func` alias two different internal types.
3. They must not overwrite a built-in annotation type.

`decomplint` will alert to any problems.

I took the opportunity to rename some linter errors to be more useful:
- `bad_decomp_marker` is now `not_strict_format` to suggest that we read the annotation without a problem, but it does not follow the style guide we established on `LEGO1`. (We probably should give users the option to disable this check altogether since the syntax parameters are not customizable. See #123 for some discussion on how we could do this.)
- The Bill and Ted inspired `bogus_marker` is now `unknown_annotation` so it's more clear that we did not do anything with the value. This is a warning and not an error because there is the potential for regular comments to be mistaken for reccmp markers. Without an indicator character (e.g. javadoc `/**`) I don't think we can spot these with 100% accuracy.

And finally, I changed some parameters in `ParserAlert` to be more broadly useful.

I don't love the prop drilling of the aliases, but I don't want to delay this further while I figure that out. The list of parameters in `Compare.__init__()` is getting unmanageable; this is also something I want to address in a future update.

### DIFF
--- a/reccmp/compare/core.py
+++ b/reccmp/compare/core.py
@@ -5,6 +5,7 @@ from typing import Iterable, Iterator
 from typing_extensions import Self
 from reccmp.project.detect import RecCmpTarget
 from reccmp.compare.diff import EntityCompareResult, RawDiffOutput
+from reccmp.parser.marker import ProjectAliases, normalize_project_aliases
 from reccmp.dir import source_code_search
 from reccmp.compare.functions import FunctionComparator
 from reccmp.formats import (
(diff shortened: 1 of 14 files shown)


madebr (approved):
Works great in a dummy project!

PR #444, opened 2026-06-06
Include stubs in total function count

disinvite:
More tweaking of the progress display 🙃

We included unmatched functions to the total function count in #407, but this does not include stubs. This means that adding a function as a stub will decrease the total function count and increase the progress percentage in error.

### DIFF
--- a/reccmp/tools/asmcmp.py
+++ b/reccmp/tools/asmcmp.py
@@ -253,11 +253,16 @@ def main() -> int:
 
         report.add_match(match)
 
+    # Includes stubs
+    function_count = sum(
+        1 for ent in report.entities.values() if ent.type == EntityType.FUNCTION
+    )
+
     # Count how many functions have the same virtual address in orig and recomp.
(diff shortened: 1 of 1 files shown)


reccmp/tools/asmcmp.py
```diff
@@ -292,8 +297,6 @@ def main() -> int:
     if args.html is not None:
         write_html_report(args.html, report)
 
-    implemented_funcs = function_count
-
     # Add known but unmatched functions to our count
     function_count += compare.count_unmatched_functions()
```
jonschz, line 301: I'd move this function up to line 259 (to be more functional - this mutation does not serve any purpose in my eyes) and expand a little on the comment. To be honest, I'm not sure what counts as an entity above vs. an "unmatched" function down here. Looks good otherwise!

jonschz (approved):

disinvite:
Back to draft for now. This may not go far enough in giving users control over which function entities are considered for the count of "how many are left to implement and match". I have another approach I will put up shortly.

PR #447, opened 2026-06-08
More direct entity counting for decomp progress

disinvite:
Improvement on #444.

The problem to solve is counting the number of functions in the decomp. You can directly set the value with `--total`. If it is not set, we need to count every annotated function from the orig address space, matched or not, stubbed or not. If `--nolib` is set, exclude library functions. We also exclude function names from the `report.exclude_functions` list.

Draft because I'm not sure where to draw the line between `asmcmp.py` and `core.py`. But this is the gist of what I think we need.

### DIFF
--- a/reccmp/compare/core.py
+++ b/reccmp/compare/core.py
@@ -1,7 +1,7 @@
 import logging
 import difflib
 import struct
-from typing import Iterable, Iterator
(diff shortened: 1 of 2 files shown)


disinvite:
Replaced by #486

PR #449, opened 2026-06-09
Support target icons for SVG and HTML reports

disinvite:
This adds an `icon` parameter to the project YML file so you can provide a path (relative to project root) of an image to use in the SVG progress display.

e.g.
```yml
targets:
  ISLE:
    filename: ISLE.EXE
    source-root: .
    hash:
    report:
      icon: assets/isle.png
  LEGO1:
    filename: LEGO1.DLL
    source-root: LEGO1
    hash:
      sha256: 14645225bbe81212e9bc1919cd8a692b81b8622abb6561280d99b0fc4151ce17
    report:
      icon: assets/lego1.png
```

While we're here, we now use that image as the favicon for the HTML report.

The `--svg-icon` argument will override the project setting.

`reccmp-aggregate` does not load the project file, so you need to keep using `--svg-icon` to get the image to appear.

### DIFF
--- a/reccmp/utils.py
+++ b/reccmp/utils.py
@@ -52,11 +52,16 @@ def read_js_file(filename: str) -> str:
     return file_header + "".join(lines)
 
 
-def base64_blob(path: Path) -> str:
+def get_base64_icon(path: Path) -> str:
+    if path.suffix.lower() != ".png":
+        logging.getLogger().error("Icon must be PNG image: %s", path)
+        return ""
+
     try:
         with path.open("rb") as f:
             return base64.b64encode(f.read()).decode("utf-8")
     except FileNotFoundError:
+        logging.getLogger().error("Could not open icon: %s", path)
         return ""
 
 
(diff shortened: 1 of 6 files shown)


madebr (approved):
Works great!

I did not know reccmp-reccmp ignored unknown keys in `reccmp-project.yml` so adding this to the yml does not have to wait for a next reccmp release.

reccmp/utils.py
```diff
@@ -114,9 +124,13 @@ def write_html_report(html_file: str, report: ReccmpStatusReport):
     # Convert the report to a JSON string to insert in the HTML template.
     report_str = serialize_reccmp_report(report, diff_included=True)
 
+    favicon = ""
+    if icon and icon.suffix == ".png":
+        favicon = base64_blob(icon)
```
jonschz: I'd add a warning log that non-png images are ignored

jonschz (approved):
Trusting @madebr's test here, code looks good

disinvite:
@jonschz We were wondering if `icon` should go under the `report` key instead of being at the target root. What do you think?

jonschz:
Both is fine with me, I'd prefer under the report key.

PR #450, opened 2026-06-12
`--markdown` output for CI

disinvite:
Sample output below. Use more bold/italics? Emojis?

## New (3) ##
0x10001100 - GolCommonDrawState::~GolCommonDrawState (new -> 100.00%)
0x10004470 - GoldDune0x38::~GoldDune0x38 (new -> 100.00%)
0x1001f1d0 - GolBmpFile::~GolBmpFile (new -> 100.00%)

## Increased (3) ##
0x1000a2c0 - GolD3DRenderDevice::FUN_1000a2c0 (98.26% -> 98.55%)
0x100184c0 - GolCommonDrawState::\`scalar deleting destructor' (90.91% -> 100.00%)
0x1001ccf0 - GolSurfaceFormat::MapRGBA (88.20% -> 93.26%)

**Implemented**: 68.33% *(997 / 1459)*
**Accuracy:** 98.37%
**Progress:** 67.22%

### DIFF
--- a/reccmp/tools/asmcmp.py
+++ b/reccmp/tools/asmcmp.py
@@ -167,6 +167,7 @@ def parse_args() -> argparse.Namespace:
         action="store_true",
         help="Exclude LIBRARY annotations from the analysis",
     )
+    parser.add_argument("--markdown", action="store_true")
     argparse_add_logging_args(parser)
 
     args = parser.parse_args()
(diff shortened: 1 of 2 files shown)


madebr:
How does it look in a table? Escape the function name with backtics for a courier font?

A little color to attract attention would make sense imho.
Perhaps sparingly add green/red/yellow emojis?

madebr:
I think it would also be useful to report how many functions were added and how much the accuracy changed (it can increase if functions are implemented, it can decrease if stubs are added).
For example [this diff](https://github.com/madebr/rec2/pull/9#issuecomment-4694679638) only shows the state after but it would be nice to see in a glance what the totals were before.

PR #451, opened 2026-06-12
Use JSDoc for type checking web UI

disinvite:
This didn't require many real code changes. The biggest one is using custom `Event` classes instead of `CustomEvent` to simplify the expansion to the default `HTMLElementEventMap`.

### DIFF
--- a/webui/package.json
+++ b/webui/package.json
@@ -11,6 +11,7 @@
   "devDependencies": {
     "@biomejs/biome": "2.0.6",
     "@playwright/test": "^1.60.0",
-    "@types/node": "^24.0.8"
+    "@types/node": "^24.0.8",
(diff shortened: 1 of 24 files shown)


webui/package.json
```diff
@@ -11,6 +11,7 @@
   "devDependencies": {
     "@biomejs/biome": "2.0.6",
     "@playwright/test": "^1.60.0",
-    "@types/node": "^24.0.8"
+    "@types/node": "^24.0.8",
+    "typescript": "^6.0.3"
```
jonschz: I personally prefer to pin those unless the package is a library

jonschz (approved):
Very nice! I didn't double-check the behaviour, but that should be covered by the Playwright tests anyway

PR #452, opened 2026-06-12
Handle `loop` instructions as jumps

disinvite:
We expect to see the jump displacement shown as the operand, not the raw address.


madebr (approved):
```
0x518700: BRPM_convert 100% match.

✨ OK! ✨
```

Thank you!

PR #459, opened 2026-06-25
Replace `match_array_elements` with offset name lookup

### LINKED ISSUE #308: `datacmp` should be more lenient with pointer comparison
Another one from [`CARM95`](https://github.com/dethrace-labs/dethrace):
```cpp
typedef struct tProgram_state {         // size: 0x356c
    // ...
    tCar_spec current_car;              // @0xb0
    // ...
} tProgram_state;

// GLOBAL: CARM95 0x00546930
tProgram_state gProgram_state;

// GLOBAL: CARM95 0x0050d3a0
tCar_spec* gCar_to_view = &gProgram_state.current_car;
```
In `datacmp` we have a diff, but nothing is wrong:
```
gCar_to_view (0x50d3a0) ... DIFF
    + 0x00                                  Unknown pointer 0x5469e0 : Unknown pointer 0x573c20
```
There is no entity at `0x5469e0` for the `current_car` offset of `gProgram_state`.

We could extend `match_array_elements` to create offset entities for structs, not just arrays.

We could also have `datacmp` try different strategies when matching pointers. Why not open the entity at this address and compare the offsets? (See also #305)

When resolved:
- [ ] Remove `xfail` test added in #317.

### LINKED ISSUE #456: Large arrays of global data slows down reccmp
In my rec2 project, running `reccmp-reccmp --target CARMA2_HW -v 0x004e9940` currently takes around 10s.
When I apply the following patch to reduce the size of a big global data array, it reduces to 3s.

```patch
--- a/src/carma2/newcommon/16-graphics1.c
+++ b/src/carma2/newcommon/16-graphics1.c
@@ -45,7 +45,7 @@ tShadow_level gShadow_level = eShadow_us_only;
 int gRender_indent;
 
 // GLOBAL: CARMA2_HW 0x006baa40
-tU8 gTemporary_physics_render_buffer[300000];
+tU8 gTemporary_physics_render_buffer[1];
 
 // MungeClipPlane
 
```

disinvite:
Fixes #308.
Fixes #456.

`match_array_elements` creates new `OFFSET` entities for any array variable. We create new entities at two levels of precision, so a 2d array or array of structs would have its members created.

The function impacts `reccmp` output in three ways. The main goal was to allow variables like this to match in `datacmp`:

```cpp
int g_values[4] = {0, 1, 2, 3};
int* g_value = &value[1];
```

For example: `g_skeletonKickPhases` from `LEGO1`.

This also adds array index and member names to the asm disassembly. This had the side effect of adding a member name or `[0]` to the parent entity at offset 0.

A _third_, more obscure benefit is that `OFFSET` entities block the creation of `FLOAT` or `STRING` entities for values that resemble floats or strings.

---

`match_array_elements` is a performance hog for large arrays (as reported on Discord) so it's time to drop it. The replacement code:
1. When matching pointers in `datacmp`, use `db.get(exact=False)` to get the parent entity and compare the offset.
2. In asm replacement, get the offset name whenever it's needed instead of creating entities.
3. To block junk strings from being created, the new `occupied()` function for the database checks whether the given address would be contained by an existing entity with a size. (`used()` checks the exact address only.)

This will eventually need a second refactor pass after merging but I didn't want to blow up the name replacement module for this. (Will create a GH issue for the `# TODOs` when it's time to merge.)

**Addendum:** This also solves a similar performance issue with arrays in the types database. The `types_db.get()` function no longer builds a `list` with each array index to imitate a struct; the caller is now expected to handle this.


madebr:
For no change in the report, this is the performance difference for rec2:

| reccmp branch | `char[300000]` | `char[1]` |
| - | - | - |
| master | 7.98s | 4.02s |
| #459 | 4.77s | 3.73s |

So this is definitively an improvement.
I have not measured what is the behavior of multiple big global arrays.
Would adding another similarly sized global array add an extra second?

disinvite:
> Would adding another similarly sized global array add an extra second?

Yes. We also have to fix `mock_array_members()`  in the types database (`cvdump/types.py`). It creates a `list` with each array offset to imitate a struct so that all the various type patterns can use the same API. We probably have to stop doing this and make arrays a special case. I'll see how big of a change that would be.

reccmp/compare/db.py
```diff
@@ -496,9 +493,25 @@ def get_lines_in_recomp_range(
                 yield match
 
     def used(self, img: ImageId, addr: int) -> bool:
+        """Is there an entity at this address?"""
         assert img in (ImageId.ORIG, ImageId.RECOMP), "Invalid image id"
         return addr in self._addr_set[img]
 
+    def occupied(self, img: ImageId, addr: int) -> bool:
```
jonschz: I like the semantics. Maybe go a little more verbose with the wording? I couldn't tell which is which without looking it up. Maybe

```
used -> entity_exists_at_exact_address
occupied -> address_overlaps_some_entity
```

or maybe copy the semantics of `self.get()`, make both one function with a mandatory kv-argument `exact`? Like
```py
def entity_exists_at_address(self, img: ImageId, addr: int, *, exact: bool) -> bool:
```

reccmp/compare/variables.py
```diff
@@ -193,15 +193,34 @@ def create_comparison_item(
...
+        base_addr = entity.addr(img)
+        assert isinstance(base_addr, int)
+
+        offset = addr - base_addr
+        if offset == 0:
+            name = entity.match_name()
+        else:
+            type_key = entity.get("data_type")
+            if type_key:
+                suffix = types.get_name_for_offset(CvdumpTypeKey(type_key), offset)
```
jonschz, line 217: I didn't fire it up in my IDE, but is there a good reason why `entity.get("data_type")` doesn't already return a `CvdumpTypeKey`?

reccmp/cvdump/types.py
```diff
@@ -500,6 +546,57 @@ def get_scalars_gapless(self, type_key: CvdumpTypeKey) -> list[ScalarType]:
 
         return output
 
+    def get_name_for_offset(self, type_key: CvdumpTypeKey, offset: int) -> str:
+        names = []
+
+        # Limited to arrays for now.
+        # The goal is to imitate the effect of match_array_elements.
```
jonschz:
```suggestion
    def get_name_for_offset(self, type_key: CvdumpTypeKey, offset: int) -> str:
        """Limited to arrays for now."""
        names = []

```
I'd drop this comment since `match_array_elements` is a dead reference after this PR

tests/test_function_comparator.py
```diff
@@ -59,7 +60,10 @@ def compare_functions(
     recomp_bin.is_relocated_addr = is_relocated_addr or Mock(return_value=False)
     recomp_bin.is_debug = Mock(return_value=False)
 
-    comp = FunctionComparator(db, lines_db, orig_bin, recomp_bin, report)
+    # TODO: Remove types dep?
```
jonschz: Tracker for TODO comments. We should make a decision before merging

reccmp/cvdump/types.py
```diff
@@ -500,6 +546,57 @@ def get_scalars_gapless(self, type_key: CvdumpTypeKey) -> list[ScalarType]:
...
+        ok_to_run = False
+        if type_key in self.keys:
+            type_dict = self.keys[type_key]
+            ok_to_run = type_dict.get("type") == "LF_ARRAY"
+
+        if not ok_to_run:
+            if offset > 0:
+                return f"+{offset}"
+
+            return ""
```
jonschz:
```suggestion
        if type_key in self.keys:
            type_dict = self.keys[type_key]
            if type_dict.get("type") != "LF_ARRAY":
                return f"+{offset}" if offset > 0 else ""
```
minor stylistic point.

Also, it looks like we already have a lot here. I am not sure myself, but will supporting more than arrays be a lot of effort at this point?

tests/test_compare_mutate_match_array.py
jonschz, line 1: I feel like we got rid of a lot of tests here. Do we have the same level of test coverage as before on the array datacmp capabilities?

jonschz (commented):
I like the direction you have taken here. Could there be any regressions on `asmcmp`? Not in terms of code that matches on both sides, but e.g. a decrease in diff readability if a read from somewhere deep in an array mismatches. Not sure myself (and even if there is a problem, I'm not sure if it's worth fixing), just raising a potential point here.

reccmp/cvdump/types.py
disinvite: No effort at all. I only blocked it because it will create more asmcmp diffs that we would have to verify. (At present, the only diffs should be for `g_variable[0]` instances changing to just `g_variable`.)

tests/test_compare_mutate_match_array.py
disinvite: Maybe we should restore the file but replace calls to `match_array_elements` with calls to get the name at the same addresses where we are now checking for entities.

tests/test_function_comparator.py
disinvite: It requires combining some dependencies, so it could end up being a larger refactor than this PR alone. Maybe `Compare` should provide an `is_valid_addr` and `name_for_address` API we can pass down to the `ParseAsm` disassembler thing?

I'll point out that the reason for the `create_valid_addr_lookup` and `create_name_lookup` framework is that it wasn't possible to parameterize which address space you wanted. These were effectively acting as `functools.partial` when we still had the `db.get_by_orig()` functions. Now that we have `db.get(image_id, ...)`, `entity.addr(image_id)` and so on, these are not really necessary.

disinvite: Also... the tests in the function comparator seem to not require name lookups aside from some of the `LINE` entities. So if it was a pure function you could just have a dict with the exact address replacement you want.

reccmp/compare/variables.py
disinvite: It's still using `_kvstore: dict[str, Any]` internally, left over from the SQLite version. I guess a `TypedDict` would be a good bridge to something else.

reccmp/cvdump/types.py
jonschz: Great! So you're planning a follow-up PR that enables the full capabilities soon?

tests/test_function_comparator.py
jonschz: Sounds reasonable. I'd either make the changes here or add an issue with follow-up tasks, so we don't have TODO comments in the commit without a ticket number

reccmp/compare/variables.py
jonschz: Doesn't have to come in this PR, just something I noticed

disinvite: #460

tests/test_compare_mutate_match_array.py
```diff
@@ -22,9 +20,25 @@ def fixture_types() -> CvdumpTypesParser:
...
 
 
-def test_match_array_with_nothing():
-    """Should not fail if there is nothing in either database."""
-    match_array_elements(EntityDb(), CvdumpTypesParser())
+def name_for_address(
+    db: EntityDb, types_db: CvdumpTypesParser, image_id: ImageId, addr: int
+) -> str:
+    entity = db.get(image_id, addr, exact=False)
+    assert entity is not None
```
disinvite, line 26: The only problem here is that if we don't add the `create_name_lookup` refactor now, we're testing something that doesn't exactly replicate the real code. I can live with that for now to not hold up the performance gains from this PR. What do you guys think? @jonschz @madebr

madebr: The performance hit is not too bad for rec2. I can live with whatever is the easiest for you.

jonschz: +1. I would at most add a comment.

reccmp/compare/db.py
disinvite: What about: 
- `exists` for checking the address only (replacing `used`)
- `intersects` for checking the address plus size range (replacing `occupied`)?

I'm starting to think `exact` on `get()` doesn't perfectly explain what is intended, but we can discuss that later.

jonschz: I like those! Feel free to make changes to `get(exact = ...)` as well as you see fit, either here or in a separate PR

disinvite:
Still need to do some final manual tests (diffing generated asm in various projects) and then I'll merge this one.

disinvite:
## Asm changes to be aware of ##

- The root address for array variables is now rendered as `g_array` instead of the qualified variants: `g_array[0]`, `g_array[0][0]`, `g_array[0].member`.
- 2d array offsets now rendered as `[0][1]` instead of `[0].[1]`
- If your binary has no relocation table, some `<OFFSET>` values may change back to raw addresses. (By fixing #464). You can solve this by adding the annotation for the variable.
- As a result of the above fix, `<OFFSET_X>` numbers may have changed inside each function.

Issue #440, opened 2026-06-02
Follow-ups to #432 (CRT startup functions)


disinvite:
Things we're not doing now in #432, but will do:

- From @jonschz: If some labels are found in recomp but not in orig, a warning should be emitted. Otherwise, reccmp will just silently skip this comparison. The warning could also link to Microsoft's documentation.

- Display the CRT arrays as if they were vtables. This isn't conceptually difficult but it requires a new entity type and special handling in the report data structure.

- Try to determine which variable is being initialized, then add its name to the function name. Since we now put addresses-being-written-to in their own bucket when matching the functions, we could get the name of the smallest address in that list and use it.

- If we can't match a CRT function, the reason why may not be readily apparent. (Our current technique is probably fine for short functions.) It would be nice if there was a way to inspect unmatched functions to see what asm replacement has taken place. (This gets into a broader topic of estimating a function's size.)

jonschz:
This recomp is currently missed by the reccmp PR:

<img width="921" height="250" alt="Image" src="https://github.com/user-attachments/assets/2a483d6f-feba-424e-a469-6b90f2b1d99e" />

Wishlist:
- Support globals initialised by a constructor call (see the screenshot)
- Would it make sense to name the matched functions after the global they initialise? E.g. `$CRT_CPP_Initializer::g_unk0x12341234`. That would probably increase the order stability and go hand in hand with the vtable-like comparison (which may e.g. produce 100 % matching variables, but show them to be initialised in the wrong order)

jonschz:
Less important, but still for the record: Some statics set up a destructor, e.g. `GOLDP 0x100043a0`, which is not detected correctly at the moment:
```diff
0x100043a0 : mov ecx, g_unk0x10062568 (DATA)
0x100043a5 : call GolImgFile::GolImgFile (FUNCTION)
-0x100043aa : push <OFFSET3>
+           : push $E4(1) (FUNCTION)
0x100043af : call _atexit (FUNCTION)
0x100043b4 : pop ecx
0x100043b5 : ret 
```

disinvite:
I have something stashed that matches pushed address to the `atexit` function if it is matched.

New problem: performance is very bad if you have a lot of CRT functions. 10.5.exe from GTA2 has thousands. If there isn't an obvious bottleneck, we should add an option to disable CRT analysis/matching.

PR #467, opened 2026-06-29
Create and match CRT functions in two steps

disinvite:
We want to do this so we can call `set_max_size` in between the two actions and establish the upper bound for how many bytes to read when disassembling.

[`105`](https://github.com/CriminalRETeam/gta2_re) has thousands of CRT functions and performance is not great since merging #432 😬

This was thrown together quickly while working on #459 and I'm sure there are more things we could improve (items from #440 perhaps)

### DIFF
--- a/reccmp/analysis/crt_startup.py
+++ b/reccmp/analysis/crt_startup.py
@@ -2,6 +2,7 @@ import enum
 import re
 import struct
 from dataclasses import dataclass
+from functools import partial
 from typing import Callable, Iterator
 from typing_extensions import Buffer
 from reccmp.compare.asm.const import JUMP_MNEMONICS
@@ -140,10 +141,7 @@ def get_function_fingerprint(
     size = get_function_sample_size(db, image_id, addr)
     raw = binfile.read(addr, size)
 
-    def entity_exists(test_addr: int) -> bool:
-        return db.get(image_id, test_addr, exact=True) is not None
-
-    collector = UsedAddressCollector(entity_exists)
+    collector = UsedAddressCollector(partial(db.used, image_id))
     collector.analyze(raw, addr)
 
     normalized_addrs = []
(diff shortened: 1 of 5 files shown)


reccmp/analysis/crt_startup.py
```diff
@@ -248,6 +246,25 @@ def detect_crt_startup_arrays(
             yield (array_type, None)
 
 
+def read_crt_arrays(
+    db: EntityDb, image_id: ImageId, binfile: PEImage
+) -> Iterator[tuple[str, list[int]]]:
```
jonschz: Either use a `NamedTuple` or a docstring explaining the return type

reccmp/compare/analyze.py
```diff
@@ -38,6 +41,23 @@ def match_entry(db: EntityDb, orig_bin: PEImage, recomp_bin: PEImage):
         batch.match(orig_bin.entry, recomp_bin.entry)
 
 
+def create_crt_functions(db: EntityDb, image_id: ImageId, binfile: PEImage):
+    crt_arrays = tuple(read_crt_arrays(db, image_id, binfile))
+
+    with db.batch() as batch:
+        for base_name, array in crt_arrays:
```
jonschz:
```suggestion
        for crt_function_base_name, crt_array in crt_arrays:
```
maybe? There are also some other instances where it might be possible to improve variable names. Just a suggestion

reccmp/compare/mutate.py
jonschz, line 93:
```suggestion
def match_crt_startup(db: EntityDb, orig_bin: PEImage, recomp_bin: PEImage):
    """Should be called after `set_max_size` was called on both images. Otherwise, a large chunk of code will be disassembled for each CRT startup function, hurting performance."""
```
Maybe something along these lines?

jonschz (approved):
Looks good! I'll take your word for the performance boost. Didn't find any regressions on GOLDP and LEGORACERS.

disinvite:
I managed to streamline this a lot by making `CrtStartupArray` the center of the analysis, returned by `detect_crt_startup_arrays` in both the entity creation and matching steps. Calculating the function fingerprint is done only when matching, after we have a max size for the entities involved.

We could call `detect_crt_startup_arrays` once and store the result somewhere, but there isn't an obvious place in the `EntityDb` that will not make things slower. Maybe we need more purpose-built storage for things like the CRT arrays, unwind functions, etc. and then limit the number of fields that go in the main database.

Through more observation, I noticed that CRT functions have one of three patterns:
```asm
; 1. Thunked constructor/setter function
jmp +0x10

; 2. Thunked constructor and atexit-setter
call +0x10
jmp <some other 16-byte aligned location> ; depending on the size of the function being called above

; 3. Non-thunk, any other pattern
...
```

so I made the `unwrap_jump` check more restrictive. Matching the functions that call `atexit` will be very simple to do in a follow-up, more simple than my intended fix.

No regressions in `LEGORACERS` or `GOLDP`. Performance in `105` is much improved.

PR #468, opened 2026-06-30
Don't create entities for substrings

disinvite:
Fixing a regression I noticed after merging #459. You can see this in `LEGO1` with the strings `"eatdn"` and `"A_Bitmap"`.

```
Decreased (3):
0x10040060 - Act3Cop::ParseAction (100.00% -> 99.25%)
0x10077cc0 - RegistrationBook::ReadyWorld (100.00% -> 98.54%)
0x100826f0 - HistoryBook::ReadyWorld (76.33% -> 75.91%)
```

`Act3Cop::ParseAction` is still reduced, but this is "correct" because we don't replace addresses for `cmp reg, value` instructions unless the value is a known entity.

### DIFF
--- a/tests/test_compare_analyze.py
+++ b/tests/test_compare_analyze.py
@@ -116,6 +116,30 @@ def test_create_analysis_strings_do_not_create_substrings(db: EntityDb):
     assert db.get(ImageId.ORIG, 102) is None
 
 
+def test_create_analysis_strings_abutting_strings(db: EntityDb):
+    """Demonstrating that our check for substrings allows strings that are
+    next to each other, separated by the null-terminator."""
+    binfile = Mock(spec=[])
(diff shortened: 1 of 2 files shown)


tests/test_compare_analyze.py
```diff
@@ -97,6 +97,25 @@ def test_create_analysis_strings_do_not_replace_overlap(db: EntityDb):
     assert e is None
 
 
+def test_create_analysis_strings_do_not_create_substrings(db: EntityDb):
+    """iter_string() uses the relocation table to search for strings.
+    Bounds checking in a loop may point to an offset of the string,
```
jonschz, line 102: Maybe add a case that back-to-back strings behave as expected, to protect against off-by-one errors

jonschz (approved):
Looks good!

PR #469, opened 2026-07-08
Bump `actions/cache` version to v6

disinvite:
I've been having sporadic failures on my fork caused by cache miss for the binfile dependencies. Not just to download them the first time, but also merging them together between steps. This appears to solve it.


madebr:
Perhaps add github-actions to dependabot?
https://docs.github.com/en/code-security/how-tos/secure-your-supply-chain/secure-your-dependencies/auto-update-actions

Issue #175, opened 2025-07-20
Replace `capstone` with `iced`


disinvite:
I ran across the [`iced`](https://github.com/icedland/iced) library for disassembling x86 and it looks to be an improvement over `capstone`.

We use the `disasm_lite` function for best performance from `capstone`, which only returns strings for the mnemonic and (combined) operands of each instruction. We have to use regular expressions to pull out the information we want and substitute the entity name where there is a pointer.

You get a lot more detail from `iced` by default. For example:

```python
if inst.op_count == 1:
    if inst.mnemonic == Mnemonic.CALL and inst.op0_kind == OpKind.NEAR_BRANCH32:
        yield ("call", inst.near_branch_target, None)
    
    elif inst.mnemonic == Mnemonic.JMP and inst.op0_kind == OpKind.NEAR_BRANCH32:
        yield ("jump", inst.near_branch_target, None)
        break

    elif inst.mnemonic == Mnemonic.JMP and inst.op0_kind == OpKind.MEMORY:
        if inst.memory_displ_size == 4:
            yield ("jump_table", inst.memory_displacement, None)
            break

    elif inst.mnemonic in IcedMnemonicJumps and inst.op0_kind == OpKind.NEAR_BRANCH32:
            yield ("jump", inst.near_branch_target, None)
```

You can get the same sort of thing from `capstone` if you enable the `md.detail` option and use the regular `disasm` function, but this is slower.

I tested out `iced` in the instruction generator (`instgen.py`) and the performance is better _until_ we need to format the string output to imitate `capstone`. At that point we break even, but if we can get more flexibility and readability from switching then it may be worth it.

Also: `iced` has a [SymbolResolver](https://docs.rs/iced-x86/1.21.0/iced_x86/index.html#disassemble-with-a-symbol-resolver) that is probably better than the regular expressions we use in the asm sanitize step.

PR #471, opened 2026-07-08
Use raw tuple from `capstone` instead of wrapping

disinvite:
Capstone's `disasm_lite()` function returns a tuple with the instruction details. We wrap it in a `namedtuple` to be self-documenting. It's faster to not do that. `LEGO1` is 500-700ms faster on my Windows box.

I've been looking at `iced_x86` again (#175) as a potential replacement for `capstone`. This will be helpful to establish the best performance with the current setup.

Except for some unit tests that still use the `DisasmLiteInst` wrapper, the instruction tuple is only passed around internally between the users of `instgen.py`: asm rendition in `asm/parse.py` and CRT function analysis.

### DIFF
--- a/tests/test_instgen.py
+++ b/tests/test_instgen.py
@@ -1,5 +1,4 @@
 from reccmp.compare.asm.instgen import InstructGen, SectionType
-from reccmp.compare.asm.types import DisasmLiteInst
 
 
 def test_ret():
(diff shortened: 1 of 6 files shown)


tests/test_instgen.py
```diff
@@ -100,8 +104,10 @@ def test_beta_case():
     assert ig.sections[0].type == ig.sections[2].type == SectionType.CODE
 
     # Make sure we captured the instruction immediately after
-    assert isinstance(ig.sections[2].contents[0], DisasmLiteInst)
-    assert ig.sections[2].contents[0].mnemonic == "mov"
+    section = ig.sections[2]
+    assert section.type == SectionType.CODE
+    instructions = [DisasmLiteInst(*inst) for inst in section.contents]
```
jonschz: Are we still using the `DisasmLiteInstr` in any production (i.e. non-test) code? If not, I'd remove it from the tests as well. Looks good otherwise!

jonschz (approved):

PR #477, opened 2026-07-10
Use native replacement for `difflib.SequenceMatcher`

disinvite:
@madebr and I were looking into this last month. [`cydifflib`](https://github.com/rapidfuzz/CyDifflib) is another free performance boost, although the magnitude varies with the number of asm instructions in your decomp.

While we are here, pin all versions for our Python dependencies. Many of them are effectively finished projects and there would never be a reason to upgrade, so make sure this does not happen without our knowing.

### DIFF
--- a/reccmp/compare/pinned_sequences.py
+++ b/reccmp/compare/pinned_sequences.py
@@ -1,8 +1,13 @@
-from difflib import SequenceMatcher
 from itertools import pairwise
 import itertools
-from typing import Iterable, Iterator, Sequence
+from typing import Iterable, Iterator, Sequence, TYPE_CHECKING
 from reccmp.difflib import DiffOpcode, get_grouped_opcodes
+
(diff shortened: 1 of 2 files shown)


reccmp/compare/pinned_sequences.py
```diff
@@ -1,9 +1,14 @@
...
 import itertools
-from typing import Iterable, Iterator, Sequence
+from typing import Iterable, Iterator, Sequence, TYPE_CHECKING
 from reccmp.difflib import DiffOpcode, get_grouped_opcodes
 
+# cydifflib has no type stubs, but we expect it to match the builtin difflib API exactly.
+if TYPE_CHECKING:
+    from difflib import SequenceMatcher
+else:
+    from cydifflib import SequenceMatcher
```
jonschz, line 11:
```suggestion
    # better performance than the stock `difflib`
    from cydifflib import SequenceMatcher
```

jonschz (approved):
Nice!

PR #484, opened 2026-07-14
Lazy parsing for cvdump type leaves

disinvite:
Defer regex parsing of each leaf in the `TYPES` section until we need it. If you're not using `datacmp` or importing to Ghidra, it turns out you can skip almost everything.

This hides the raw data in `types._keys` behind the `from_key()` function. Many of the tests need to set `_keys` values directly, so we'll have to live with the `pylint:disable=protected-access` escape hatch for now.

This squeezed out another 400-500ms for `LEGO1` on my Windows box, although this is surely less pronounced on a machine from this decade.

The `SYMBOLS` and `LINES` sections are now the slowest in the profiler. We could apply the same principle to `SYMBOLS` and see what happens. For `LINES`, we actually do need everything up front, so this is where we'll benefit most from parsing the PDB directly, either with Python or a native module.

### DIFF
--- a/reccmp/cvdump/parser.py
+++ b/reccmp/cvdump/parser.py
@@ -2,7 +2,7 @@ import re
 from pathlib import PureWindowsPath
 from typing import NamedTuple
 from .types import CvdumpTypesParser
-from .symbols import CvdumpSymbolsParser
+from .symbols import CvdumpSymbolsParser, SymbolsEntry
 from .cvinfo import CvdumpTypeKey
 
 # e.g. `     27 00034EC0     28 00034EE2     29 00034EE7     30 00034EF4`
@@ -118,7 +118,7 @@ class CvdumpParser:
         self.symbols_parser = CvdumpSymbolsParser()
 
     @property
-    def symbols(self):
+    def symbols(self) -> list[SymbolsEntry]:
         return self.symbols_parser.symbols
 
     def _lines_section(self, line: str):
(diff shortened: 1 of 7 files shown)


reccmp/cvdump/parser.py
```diff
@@ -118,7 +118,7 @@ def __init__(self) -> None:
         self.symbols_parser = CvdumpSymbolsParser()
 
     @property
-    def symbols(self):
+    def symbols(self) -> list[SymbolsEntry]:
```
disinvite, line 121: FYI: This enhances type checking for the consumer in `cvdump/analysis.py` but is unrelated to the wider PR.

disinvite:
`SYMBOLS` and `LINES` refactoring led to smaller performance improvements. (Mainly because we can't simply avoid the work as we can here.)

They might still be worth adding (in their own PR) while we work on the replacement parser (or gluecode for `pdb-rs`).

jonschz:
> Many of the tests need to set _keys values directly, so we'll have to live with the pylint:disable=protected-access escape hatch for now.

Maybe a public helper like `set_parsed_key()` with a docstring saying that this is (mostly or exclusively) supposed to be used in tests? Looks good overall.

disinvite:
> Maybe a public helper like `set_parsed_key()` with a docstring saying that this is (mostly or exclusively) supposed to be used in tests? Looks good overall.

Created #485 to fix the tests later. I think what we really need is a public API on the types database so that none of the tests are concerned with parsing cvdump text or the internal data structures. Going to expound on that in #106.

PR #486, opened 2026-07-20
Use report for progress calculation

disinvite:
This makes the following changes to `ReccmpStatusReport` and its child struct `ReccmpComparedEntity`:

- The report created by `reccmp-reccmp` now contains both matched and _unmatched_ entities. The reason is so we can use the contents of `ReccmpStatusReport` only to compute the project accuracy and project statistics. Unmatched entities are excluded from display and the serialized report files.

- When creating the `ReccmpStatusReport`, skip any entities that match the "ignore these function names" and "ignore library functions" filters if either are enabled.

- Similarly, a `ReccmpStatusReport` loaded from a file can also exclude entities by function name or the `is_library` flag. This is so you can diff against a report that includes more entities than you want and not see a lot of dropped functions.

- Store the `is_library` flag for entities in the report. A future enhancement to the HTML report could show or hide these, similar to stubs.

- Store the `function_total` value in the report. A later update will display progress/accuracy calculations in the HTML report because we now have all the data.

- Drop the redundant `report_progress_stats` function. We can get the number of implemented functions and the sum of the effective accuracy from the similar `report_function_accuracy` function. We have stopped displaying "raw accuracy" (i.e. not rounding effective matches to 100%) but we still calculate it here.

> Former changes that have been rebased out:
> - #493
> - #500


disinvite:
After some thought, this is too big to comfortably review. Let's start with #493 then rebase this one.

disinvite:
The function total now embedded in the report should help finish up #450.

reccmp/compare/report.py
```diff
@@ -46,25 +47,38 @@ class ReccmpComparedEntity:
     """True if this entity had no fixed recomp address across the
     samples combined by reccmp-aggregate."""
 
+    def is_matched(self) -> bool:
+        return self.recomp_addr is not None or self.recomp_addr_varies
+
+    def is_function(self) -> bool:
+        """Entities without a type (derived from older reports that did not
+        serialize this field) are considered functions to maintain compatibility."""
```
jonschz, line 55: nice

reccmp/compare/report.py
```diff
@@ -88,53 +102,78 @@ def has_same_source(self, other: "ReccmpStatusReport") -> bool:
         """Were both reports derived from the same reccmp target?"""
         return self.filename.lower() == other.filename.lower()
 
+    def update_function_count(self) -> None:
```
jonschz, line 105: Maybe `update_function_total`, or rename the variable to `total_function_count` for consistency?

reccmp/compare/report.py
```diff
@@ -88,53 +102,78 @@ def has_same_source(self, other: "ReccmpStatusReport") -> bool:
...
+            del self.entities[key]
+
+        # Manually decrease it because recalculating will use
+        # the higher of either the previous or current count.
+        self.function_total -= functions_removed
+
+    def asmcmp_filtering(self, nolib: bool, ignore_functions: list[str]) -> None:
+        """Helper to filter the report using the current filter options from `reccmp-reccmp`.
+        Compare with `entity_filter` in asmcmp.py that acts on the `ReccmpEntity` object.
+        """
```
jonschz, line 135: This sounds like there could be an option to unify this code, but maybe this duplication is actually better. I haven't thought too much about it

tests/test_report_version_1.py
```diff
@@ -251,7 +311,6 @@ def test_serialize_excludes_defaults():
     assert "type" not in entity
 
 
-@pytest.mark.xfail(reason="TODO: Serialize changes report timestamp")
```
jonschz, line 254: nice

jonschz (approved):
Looks good from a code perspective. I haven't tried this locally - please tell me if you'd like me to try some real-world scenarios.

reccmp/compare/report.py
disinvite: We should revisit if/when we add more filtering options for the report.

disinvite:
> Looks good from a code perspective. I haven't tried this locally - please tell me if you'd like me to try some real-world scenarios.

I ran a few just now to confirm. The function count, if calculated, should match:
```powershell
reccmp-roadmap --target <target> --verbose | Select-String "^0.* (?:fun|vto) " | Measure-Object -Line
```
(or the `grep` and `wc -l` equivalent)

PR #493, opened 2026-07-21
Use `int` as address for in-memory report

disinvite:
JSON and HTML report files expect a hex string for each address. We also use this to print to the terminal.

Instead of converting the `int` address when the in-memory report is created, delay this until we need to write a file or display something. (In the spirit of #307.)

The sticking point is the magic string `"various"` that appears for the recomp addr when the entity is created by `reccmp-aggregate`. We have a new property for this to distinguish between the recomp addr being `None` because this is an unmatched entity or because it is matched but has no fixed recomp address. We don't currently use the report to hold unmatched entities, but I want to do this in a follow-up to fix some gaps with accuracy/progress calculation. (See: #444 and #447.)

We did not have any tests specific to the version 1 report format, so I added some.

### DIFF
--- a/reccmp/compare/report.py
+++ b/reccmp/compare/report.py
@@ -7,6 +7,13 @@ from reccmp.types import EntityType
 from .diff import CombinedDiffOutput, DiffReport, RawDiffOutput, raw_diff_to_udiff
 
 
+def format_address(addr: int) -> str:
+    """This is here just to document each spot where we
+    convert an int address into a string.
+    In the future, the format may be customizable. (GH #370)"""
+    return f"{addr:#x}"
+
+
 class ReccmpReportDeserializeError(Exception):
     """The given file is not a serialized reccmp report file"""
 
@@ -18,18 +25,22 @@ class ReccmpReportSameSourceError(Exception):
 @dataclass
 class ReccmpComparedEntity:
     # pylint:disable=too-many-instance-attributes
-    orig_addr: str
+    orig_addr: int
     name: str
     accuracy: float
     # Version 1 files have no type, so it is optional.
     type: EntityType | None = None
-    recomp_addr: str | None = None
+    recomp_addr: int | None = None
     is_effective_match: bool = False
     is_stub: bool = False
     rdiff: RawDiffOutput | None = None
 
     # Legacy field for importing version 1 files (aggregate).
     udiff: CombinedDiffOutput | None = None
+
+    recomp_addr_various: bool = False
+    """True if this entity had no fixed recomp address across the
+    samples combined by reccmp-aggregate."""
 
     @property
     def effective_accuracy(self) -> float:
@@ -46,7 +57,7 @@ class ReccmpStatusReport:
     timestamp: datetime
 
     # Using orig addr as the key.
-    entities: dict[str, ReccmpComparedEntity]
+    entities: dict[int, ReccmpComparedEntity]
 
     # Only set during deserialize.
     from_version: int | None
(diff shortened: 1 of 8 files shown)


reccmp/compare/report.py
```diff
@@ -18,19 +25,23 @@ class ReccmpReportSameSourceError(Exception):
...
     is_effective_match: bool = False
     is_stub: bool = False
     rdiff: RawDiffOutput | None = None
 
     # Legacy field for importing version 1 files (aggregate).
     udiff: CombinedDiffOutput | None = None
 
+    recomp_addr_various: bool = False
+    """True if this entity had no fixed recomp address across the
+    samples combined by reccmp-aggregate."""
```
jonschz: Maybe `reccmp_addr_not_unique` or `reccmp_addr_ambiguous`?

reccmp/compare/report.py
```diff
@@ -18,19 +25,23 @@ class ReccmpReportSameSourceError(Exception):
...
 class ReccmpComparedEntity:
     # pylint:disable=too-many-instance-attributes
-    orig_addr: str
+    orig_addr: int
     name: str
     accuracy: float
     # Version 1 files have no type, so it is optional.
     type: EntityType | None = None
-    recomp_addr: str | None = None
+    recomp_addr: int | None = None
```
jonschz, line 33:
```suggestion
    recomp_addr: int | None = None
    """Can be `None` if the recomp address is not unique across an aggregate, or <potential other reasons>"""
```
maybe?

tests/test_report_version_1.py
```diff
@@ -0,0 +1,278 @@
...
+
+    # The remaining optional fields are not in the JSON.
+    assert "stub" not in entity
+    assert "library" not in entity
+    assert "effective" not in entity
+    assert "diff" not in entity
+    assert "type" not in entity
+
+
+@pytest.mark.xfail(reason="TODO: Serialize changes report timestamp")
```
jonschz, line 254: Is there a ticket for this TODO?

reccmp/compare/report.py
```diff
@@ -249,16 +258,26 @@ class JSONReportVersion1(BaseModel):
...
 def _serialize_version_1(
     report: ReccmpStatusReport, diff_included: bool = False
 ) -> JSONReportVersion1:
     """The JSON report can exclude the diff to make deserialization faster."""
     entities = [
         JSONEntityVersion1(
-            address=addr,  # prefer dict key over redundant value in entity
+            address=format_address(
+                addr
+            ),  # prefer dict key over redundant value in entity
```
jonschz: It was here before, but I don't understand the comment `# prefer dict key over redundant value in entity`

jonschz (approved):
Looks good to me. I also didn't find any regression when trying `reccmp-reccmp --json` and `reccmp-reccmp --diff`, though I didn't test a lot of scenarios.

tests/test_report_version_1.py
disinvite: It's fixed by #486, if we merge it after a rebase. I wanted this PR to just be about changing the datatype and the cleanup required by doing that.

reccmp/compare/report.py
disinvite: The `entities` `dict` in the report is keyed by the orig address. Each entity has _its own_ `orig_addr` property. When writing the report file, we use the key.

I can't think of a good reason that the values would not match. Maybe there should be an `assert` here?

disinvite: (Assert added)

reccmp/compare/report.py
disinvite: I went with `recomp_addr_varies` to be more idiomatic for now.

This could use some improvement, but it's tricky. I thought about adding `aggregate: bool` onto the report instead of this `varies` property, which is "better", but the blocker is that we are using `reccmp-aggregate` as a hack to produce the HTML file for the playwright tests. We need a generic tool like `reccmp-report` that can translate between formats and have the aggregate as its own format subset.

PR #500, opened 2026-07-25
Drop `DiffReport` struct

disinvite:
Whittling down #486 to just the new changes.

This one drops the `DiffReport` struct, a stop on the way from a `ReccmpMatch` (metadata about the program) to a `ReccmpComparedEntity` (the finished product).

#493 introduced the possibility that `ReccmpComparedEntity` objects are unmatched. This is the "new" feature that we will explore in #486, so for now the plumbing only deals with `ReccmpMatch` objects. I added some code to check the `recomp_addr` where this is necessary.

`reccmp-reccmp` allows filtering the report entities in two ways:
- Exclude library functions via `--nolib`
- Exclude functions by name match in `reccmp-project.yml`

This PR moves the filtering inside of `core.py`. (I tried something similar in #447.) The question of whether to filter the finished `ReccmpComparedEntity` objects or the `ReccmpMatch` objects comes down to whether you want the ability to actually _skip_ the expensive comparison option. If we are filtering on `ReccmpComparedEntity`, the asm rendering and diffing has already occurred. This more closely matches the current behavior.

### DIFF
--- a/reccmp/compare/core.py
+++ b/reccmp/compare/core.py
@@ -1,7 +1,7 @@
 import logging
 import difflib
 import struct
-from typing import Iterable, Iterator
+from typing import Callable, Iterator
 from typing_extensions import Self
 from reccmp.project.detect import RecCmpTarget
 from reccmp.compare.diff import EntityCompareResult, RawDiffOutput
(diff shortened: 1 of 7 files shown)


jonschz (approved):
LGTM. Didn't test for regressions myself, I presume you tried some of our projects? Especially the `dump` argument, which had some bigger code changes.

disinvite:
> LGTM. Didn't test for regressions myself, I presume you tried some of our projects? Especially the `dump` argument, which had some bigger code changes.

Yes. No diff using the `--dump` option in `LEGO1` or `BETA10`. I didn't see any `reccmp` diffs there or in `racers` either.

PR #502, opened 2026-07-29
More options for asm output

### LINKED ISSUE #275: Option to configure (or disable) asm sanitize
Example scenario: When using `reccmp` to mark STL or CRT functions, it would be helpful to just see the orig address rather than maintaining the rhythm of copying the _location_ of the address in to Ghidra, jumping there, then copying out the called address.

For decomps that are near the finish line, it may be useful to disable sanitize altogether to see what diffs remain. (This will introduce noise, see #274 for example)

disinvite:
Fixes #275. Adds these options to `reccmp-reccmp` asm generation output:
- `--no-udiff` alters `--verbose` mode so you see the entire function, including long runs of matching instruction that are hidden by the unified diff.
- `--no-placeholder` will not replace any address unless it is named. (No `<OFFSETX>`)
- `--no-sanitize` disables all address replacement.

`--no-placeholder` and `--no-sanitize` also impact the JSON/HTML reports. (Adding `--no-udiff` to the report should be a follow-up PR if there's any interest.)

Since we are adding more command line options, I added some [argument groups](https://docs.python.org/3/library/argparse.html#argument-groups) to improve the `reccmp-reccmp --help` display.


reccmp/tools/asmcmp.py
```diff
@@ -113,69 +117,93 @@ def virtual_address(value) -> int:
...
+        help="Replace undiscovered addresses with `<OFFSET>` (default: True)",
+    )
+    asm_group.add_argument(
+        "--sanitize",
+        action=argparse.BooleanOptionalAction,
+        default=True,
+        help="Replace addresses with name or placeholder (default: True)",
+    )
+    term_group.add_argument(
+        "--udiff",
```
jonschz, line 202: Unless I'm wrong, `udiff` isn't a term beyond Python's difflib. How about `--no-grouped-diff`? Open to other suggestions

reccmp/tools/asmcmp.py
```diff
@@ -113,69 +117,93 @@ def virtual_address(value) -> int:
...
+        "--placeholder",
+        action=argparse.BooleanOptionalAction,
+        default=True,
+        help="Replace undiscovered addresses with `<OFFSET>` (default: True)",
+    )
+    asm_group.add_argument(
+        "--sanitize",
+        action=argparse.BooleanOptionalAction,
+        default=True,
+        help="Replace addresses with name or placeholder (default: True)",
```
jonschz, line 199: From that description alone I would have a hard time telling how this differs from `--placeholder`. I don't have a good suggestion for a better description right now, though

reccmp/tools/asmcmp.py
```diff
@@ -113,69 +117,93 @@ def virtual_address(value) -> int:
...
         help="Don't display text summary of matches",
     )
-    parser.add_argument(
+    output_group.add_argument(
         "--nolib",
         action="store_true",
         help="Exclude LIBRARY annotations from the analysis",
     )
+    asm_group.add_argument(
+        "--placeholder",
```
jonschz, line 190: `--placeholderS` or `--address-placeholders`?

jonschz (commented):
Brief remarks (not a full review yet). I like the options and the direction

reccmp/compare/asm/parse.py
```diff
@@ -33,7 +34,21 @@ def from_hex(string: str) -> int | None:
...
+class AsmRenderOptions:
+    do_sanitize: bool = True
+    number_placeholders: bool = True
+    use_placeholders: bool = True
+
+
 class ParseAsm:
+    addr_test: AddrTestProtocol | None
+    name_lookup: NameReplacementProtocol | None
+    is_32bit: bool
```
jonschz, line 47: 32 in contrast to 16 or 64? Maybe an enum would be more appropriate?

reccmp/compare/asm/parse.py
```diff
@@ -193,6 +218,17 @@ def sanitize(self, inst: DisasmLiteTuple) -> tuple[str, str]:
 
         return (inst_mnemonic, op_str)
 
+    def fix_jump(self, inst: DisasmLiteTuple) -> tuple[str, str]:
```
jonschz, line 221: I would have expected some similar code to be removed elsewhere, and/or this code to be called from `sanitize`. Is there a new duplication now?

jonschz (approved):

reccmp/tools/asmcmp.py
disinvite: https://en.wikipedia.org/wiki/Diff#Unified_format

but this may not have much meaning to our users.

reccmp/compare/asm/parse.py
disinvite: Could add an enum to #503. `is_32bit` was done just to get basic 16-bit support working.

reccmp/compare/asm/parse.py
disinvite: The code is intentionally duplicated. With how frequently we call this method, I would expect the small overhead of a nested call to jump handling to be visible in the profiler. But I can check to see. (This module would be a great candidate for Rust conversion)

reccmp/tools/asmcmp.py
disinvite: The other option is a enum with different "replacement options" from least to most strict:
1. Placeholders, not numbered (unused option I added a long time ago, for some reason)
2. Placeholders (current behavior)
3. No placeholders
4. No replacement at all (`--no-sanitize`)

but how do you capture that in a meaningful way? And we might someday need another option that's more nuanced and doesn't fit this gradient.

While on the topic: I've never been totally happy with "sanitize" to describe the address replacement.

disinvite:
I’ve spent a few cycles trying to get the param names right. What about this direction: `--asm-label-entities` and `--asm-mask-addrs`? Both with a selection of choices instead of as booleans with the “no” prefix.

The difficult part is how to convey that disabling the name option means we fall through to the mask/placeholder option. 

Or would it would be most clear with a single param that accepts a comma delimited list of preferred label options?

PR #504, opened 2026-08-02
Support folded VTABLE annotations

ritsuwastaken:
I ran into an issue where /OPT:ICF set for MSVC seems to have folded identical vtables, so wanted to suggest this change

### DIFF
--- a/reccmp/compare/ingest.py
+++ b/reccmp/compare/ingest.py
@@ -12,6 +12,7 @@ from reccmp.formats import PEImage, TextFile
 from reccmp.cvdump import CvdumpTypesParser, CvdumpAnalysis
 from reccmp.parser import DecompCodebase
 from reccmp.parser.marker import ProjectAliases
+from reccmp.parser.node import ParserVtable
 from reccmp.types import EntityType, ImageId
 from reccmp.compare.event import (
     ReccmpEvent,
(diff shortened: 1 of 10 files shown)


disinvite:
Can you give an example of a folded vtable you see in your decomp?

It seems like this change does not impact the analysis excerpt for allowing `// VTABLE: TEST 0x1234 FOLDED` to pass by the linter without an error. Is that intended?

ritsuwastaken:
> Can you give an example of a folded vtable you see in your decomp?

hi, sure: https://github.com/ritsuwastaken/msvc_vtable_folded_example
it seems like it's enough to have /OPT:ICF for newer MSVC versions, however for older versions there should be /GL as well

> It seems like this change does not impact the analysis excerpt for allowing // VTABLE: TEST 0x1234 FOLDED to pass by the linter without an error. Is that intended?

sorry, not really. not sure what should be the proper fix

disinvite:
Thanks! For reference, Ghidra shows:
```
                    B::`vftable'                      XREF[2]: A:00401159(*), 
                    A::`vftable'                               ~A:004019e1(*)  
                    ??_7A@@6B@
                    ??_7B@@6B@
     4053c8 90 11 40 00   addr    A::`scalar_deleting_destructor'
     4053cc 00 1a 40 00   addr    A::value
                    Base::`vftable'                   XREF[2]: Base:0040117a(*), 
                    ??_7Base@@6B@                              ~Base:004019aa(*)  
     4053d0 d0 11 40 00   addr    Base::`scalar_deleting_destructor'
     4053d4 c0 19 40 00   addr    Base::value
```
The "extra" field that follows a `// VTABLE` marker is already spoken for: it indicates the virtual base class, as in ``RaceSkel::`vftable'{for `LegoAnimActor'}``. We could check for `FOLDED` as a magic string to indicate that you want to suppress the linter alert instead.

Are you able to match these folded vtables in your real project, even without this proposed change?

ritsuwastaken:
Without the change in this MR, I'm getting:
```bash
ERROR Dropped duplicate address 0x12345678 on VTABLE annotation in file: {filename}
```
and non of the entities end up in the report.

With the change, last processed one (EntityBatch.set()) gets into the report, sort of like it already works with functions.

But I don't know what is the expected behavior for both

ritsuwastaken:
@disinvite tried extending vtable matching

disinvite:
The latest change looks good. We should add some tests before merging. For example:
- `FOLDED` sets `is_folded` and not `base_class`.
- Single vtable marker for an address leaves `folded_vtables` blank
- `folded_vtables` appear in the same order as the markers (and the first one sets the primary name for the entity)
- For matching: vtable names can be loaded in any order and still match

I can add these too, up to you.

From yesterday's update: I had forgotten that we delete the duplicate marker annotations, so we have to reuse `FOLDED` to keep both.

Interesting distinction between folded functions and vtables. For functions we are matching by line number. We don't know which one the compiler will choose so we need all markers enabled. With vtables, all the names are available to match with, but `reccmp` only stores one name for an entity, so we need them all added on one or both sides to have a stable match.

ritsuwastaken:
Added tests

disinvite:
Looks great, thank you!

PR #513, opened 2026-08-14
Prefer a known orig vtable size over the recomp size estimate

sandboxmo-de:
The recomp vtable size is an estimate (next-symbol distance or section contribution, per SymbolNode.size()) that may include trailing alignment padding. When check_vtables reads the orig vtable through that window, the padding bytes after the true table scan as null pointers and produce false 'recomp vtable is larger' warnings. A data source can supply the orig vtable's true size; prefer it via any_size(ImageId.ORIG) in check_vtables and _compare_vtable. Behavior is unchanged when no orig size is available (any_size falls back to the recomp size).

Measured on an MSVC 7.0 target driven by CSV data sources: 32 of 32 'recomp vtable is larger' warnings were this false positive.

### DIFF
--- a/reccmp/compare/verify.py
+++ b/reccmp/compare/verify.py
@@ -3,98 +3,30 @@ These functions report problems with the current entities that limit or block fu
 """
 
 import logging
-import struct
-from reccmp.formats.pe import PEImage
 from reccmp.types import EntityType, ImageId
 from .db import EntityDb
 
 logger = logging.getLogger(__name__)
 
 
-def _table_has_entries_past(image: PEImage, addr: int, start: int, end: int) -> bool:
-    """Whether the vtable at `addr` has any real entry in the byte range
-    [start, end). The size that produced `end` may over-count the table by a
-    trailing alignment slot, so only a non-null pointer counts as an entry."""
-    start = 4 * (start // 4)
-    end = 4 * (end // 4)
-    if end <= start:
-        return False
+def check_vtables(db: EntityDb):
+    """Alert to vtable size information that cannot be correct.
 
-    tail = image.read(addr + start, end - start)
-    return any(addr != 0 for addr, in struct.iter_unpack("<L", tail))
-
-
-def check_vtables(db: EntityDb, orig_bin: PEImage, recomp_bin: PEImage):
-    """Alert to cases where the recomp vtable is larger than the one in the orig binary.
-
-    If a data source gives the orig vtable's true size, we can compare the two
-    tables directly. Otherwise the orig size is unknown and we estimate it from:
-    1. The address of the following vtable in orig, which gives an upper bound on the size.
-    2. The pointers in the orig vtable. If any are zero bytes, this is alignment padding between two vtables.
-
-    Both estimates read the orig table at the *recomp* size, which is itself an
-    estimate (next-symbol distance or section contribution) and may over-count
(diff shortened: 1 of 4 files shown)


disinvite:
If the recomp vtable is longer, do we still display the extra virtual functions in the diff?

sandboxmo-de:
> do we still display the extra virtual functions in the diff?

Nope, no it didn't. Fixed in latest commit; let me know if you'd like any tweaks to the logic for handling this

disinvite:
Do we still need the "Recomp vtable larger" message? What do you think about dropping it @jonschz? (I recall we had some back-and-forth about the message on some earlier PR.)

Any diff will show in the regular `reccmp` output with the option to block a PR if you run `reccmp-vtable` in CI. This PR will close two gaps: use known orig size if we have it, and drop trailing null pointers from the recomp side if the padding is part of the COMDAT size from the PDB. (Maybe we should do the same on the orig side?)

jonschz:
I'm fine with dropping the warning

sandboxmo-de:
"Recomp vtable is larger than orig vtable" warning has been removed

reccmp/compare/verify.py
```diff
@@ -3,44 +3,29 @@
...
-    We can tell by looking at:
-    1. The address of the following vtable in orig, which gives an upper bound on the size.
-    2. The pointers in the orig vtable. If any are zero bytes, this is alignment padding between two vtables.
+def check_vtables(db: EntityDb):
+    """Check vtable sizes provided by the data source.
+
+    Size differences between the orig and recomp vtables are reported by the
+    vtable comparison, so they are not checked here. The only check is that a
+    user-provided orig vtable size does not extend past the next entity in the
+    binary. If it does, the size can't be right, so warn about it.
```
disinvite: We can probably just say precisely what we're doing: checking for entities (vtables only, for now) where the known size and calculated (max) size are in conflict. And drop any reference to the earlier behavior.

tests/test_compare_output.py
```diff
@@ -567,3 +567,88 @@ def test_entity(
...
+
+
+def test_compare_vtable_recomp_longer():
+    """An extra virtual function on the recomp side should appear in the
+    diff when the orig vtable size is known."""
+    function_bytes = b"\xc3\x00\x00\x00"  # `ret` padded to 4 bytes
+    functions = function_bytes + function_bytes
+
+    vtable_addr0 = b"\x00\x00\x00\x00"
+    vtable_addr4 = b"\x04\x00\x00\x00"
```
disinvite: We should use more realistic pointer values so `vtable_addr0` doesn't match the padding bytes. I just merged in #519 to make that possible. We should change it here and in the next test.

reccmp/compare/verify.py
sandboxmo-de: Done

tests/test_compare_output.py
sandboxmo-de: Done

sandboxmo-de:
Unrelated to the content of these PRs, but would it make sense to change the squash commit message behavior for this project? Didn't realize the walls of text in the Claude commits were going to all get accumulated into the giant wall of text on merge >_<


image


I imagine "Pull request title and description" rather than "Pull request title and commit details" makes the most sense

PR #514, opened 2026-08-14
Make MSVC symbol truncation dependent on reported conversion in TYPES section of PDB

sandboxmo-de:
Compare.run() hardcoded truncate=True for match_symbols and match_functions. The 255-character truncation is a quirk of MSVC 4.x-era PDBs (C4786); on later toolchains it makes distinct long symbols collide and produces arbitrary non-unique matches. Add a truncate-symbols option to the project file target schema (default true, preserving current behavior) and plumb it through RecCmpTarget to Compare.

### DIFF
--- a/reccmp/cvdump/analysis.py
+++ b/reccmp/cvdump/analysis.py
@@ -147,11 +147,12 @@ class CvdumpAnalysis:
     @property
     def truncate_symbols(self) -> bool:
         """Whether symbol names in this PDB are truncated to 255 characters.
-        This is a limitation of the original design of MSVC 4.x-era PDBs
-        (see C4786); on later toolchains truncating makes distinct long
-        symbols collide and produces arbitrary non-unique matches. If the
-        TYPES section was not dumped we cannot tell, so assume truncation
-        and keep the historical behavior."""
+
+        MSVC 4.x-era PDBs truncate long symbol names (see C4786). Later
+        toolchains don't, and truncating their names anyway can make
+        different long symbols collide and match incorrectly. If the TYPES
+        section was not dumped, we can't tell the PDB version, so assume
+        truncation, which matches the old behavior."""
         return self.parser.is_16bit_type_pool is not False
 
     def __init__(self, parser: CvdumpParser):
--- a/reccmp/cvdump/analysis.py
+++ b/reccmp/cvdump/analysis.py
@@ -147,10 +147,11 @@ class CvdumpAnalysis:
     @property
     def truncate_symbols(self) -> bool:
         """Whether symbol names in this PDB are truncated to 255 characters.
-        This is a quirk of MSVC 4.x-era PDBs (see C4786); on later toolchains
-        truncating makes distinct long symbols collide and produces arbitrary
-        non-unique matches. If the TYPES section was not dumped we cannot tell,
-        so assume truncation and keep the historical behavior."""
+        This is a limitation of the original design of MSVC 4.x-era PDBs
+        (see C4786); on later toolchains truncating makes distinct long
+        symbols collide and produces arbitrary non-unique matches. If the
+        TYPES section was not dumped we cannot tell, so assume truncation
+        and keep the historical behavior."""
         return self.parser.is_16bit_type_pool is not False
 
     def __init__(self, parser: CvdumpParser):
--- a/reccmp/cvdump/analysis.py
+++ b/reccmp/cvdump/analysis.py
@@ -144,6 +144,15 @@ class CvdumpAnalysis:
     parser: CvdumpParser
     lines: dict[PureWindowsPath, list[LineValue]]
 
+    @property
+    def truncate_symbols(self) -> bool:
+        """Whether symbol names in this PDB are truncated to 255 characters.
+        This is a quirk of MSVC 4.x-era PDBs (see C4786); on later toolchains
+        truncating makes distinct long symbols collide and produces arbitrary
+        non-unique matches. If the TYPES section was not dumped we cannot tell,
+        so assume truncation and keep the historical behavior."""
+        return self.parser.is_16bit_type_pool is not False
+
     def __init__(self, parser: CvdumpParser):
         """Read in as much information as we have from the parser.
         The more sections we have, the better our information will be."""
--- a/tests/test_cvdump_symbol_truncation.py
+++ b/tests/test_cvdump_symbol_truncation.py
@@ -10,12 +10,12 @@ from reccmp.cvdump.runner import iter_cvdump_sections
 TYPES_16BIT = """
 *** Converting 16-bit types to 32-bit equivalents
 
-0x1000 : Length = 14, Leaf = 0x1201 LF_ARGLIST argument count = 1
+0x1000 : Length = 6, Leaf = 0x1201 LF_ARGLIST argument count = 0
 """
 
 # The same for a PDB whose type pool is already 32-bit.
 TYPES_32BIT = """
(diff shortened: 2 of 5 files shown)


disinvite:
If we take it as fact that the compiler version determines whether names/mangled-names are truncated in the PDB (i.e. whether they are Pascal strings with length as a single byte), and since we need to add something to the YML anyway to allow this behavior to be configurable, I think adding a `compiler` option to the target would add more value.

We have to keep the current behavior (truncating) as default for now, so maybe a good starting point would be an enum with `"auto"` and `"msvc-modern"`? Where `"auto"` means "autodetect". Which we do not do right now, but maybe some day. (And not in scope for this PR, for the record.)

If we go this route, let's start with just those two options to satisfy the need behind this PR. We'll figure out version numbering later.

What do you think?

disinvite:
Another simpler option: the types section begins with this message if we are reading an MSVC 4.x PDB:
```
*** TYPES

*** Converting 16-bit types to 32-bit equivalents
```

so we could add something to `/cvdump/parser.py` that assumes we will always truncate (current behavior) unless the "Converting 16-bit types" message is missing.

https://github.com/microsoft/microsoft-pdb/blob/805655a28bd8198004be2ac27e6e0290121a5e89/PDB/dbi/tpi.cpp#L321-L324
https://github.com/microsoft/microsoft-pdb/blob/805655a28bd8198004be2ac27e6e0290121a5e89/cvdump/dumppdb.cpp#L244-L246

sandboxmo-de:
> so we could add something to /cvdump/parser.py that assumes we will always truncate (current behavior) unless the "Converting 16-bit types" message is missing.

Done -- we now truncate unless the banner is missing from the `TYPES` section

tests/test_cvdump_symbol_truncation.py
```diff
@@ -0,0 +1,75 @@
...
+
+from reccmp.cvdump.parser import CvdumpParser
+from reccmp.cvdump.analysis import CvdumpAnalysis
+from reccmp.cvdump.runner import iter_cvdump_sections
+
+# The first lines of the TYPES section for an MSVC 4.x-era PDB.
+TYPES_16BIT = """
+*** Converting 16-bit types to 32-bit equivalents
+
+0x1000 : Length = 14, Leaf = 0x1201 LF_ARGLIST argument count = 1
```
disinvite: To make this accurate to real cvdump output:
```
0x1000 : Length = 6, Leaf = 0x1201 LF_ARGLIST argument count = 0
```
(same for the other samples)

disinvite:
I ran a test on MSVC 4, 6, and 7 to see what gets truncated when.

On 4.x, names in `SYMBOLS` and mangled names in `PUBLICS` are both truncated to 255.

On 6.x, they are both truncated to 247, followed by an 8 digit hash.

On 7.x, not truncated at 255. They may be truncated at some higher threshold.

So we are slightly inaccurate to say that the symbols are _only_ truncated on MSVC 4.x and (presumably) older. But I think this approach is good enough for now, because we can't generate hashes for MSVC 6.x without knowing the function they used.

tests/test_cvdump_symbol_truncation.py
sandboxmo-de: Done

reccmp/cvdump/analysis.py
```diff
@@ -144,6 +144,15 @@ class CvdumpAnalysis:
     parser: CvdumpParser
     lines: dict[PureWindowsPath, list[LineValue]]
 
+    @property
+    def truncate_symbols(self) -> bool:
+        """Whether symbol names in this PDB are truncated to 255 characters.
+        This is a quirk of MSVC 4.x-era PDBs (see C4786); on later toolchains
+        truncating makes distinct long symbols collide and produces arbitrary
```
disinvite: I don't know if "quirk" is the right way to describe it. Might be more accurate to call it a limitation of the original design.

tests/test_cvdump_symbol_truncation.py
```diff
@@ -0,0 +1,75 @@
+"""Detecting whether a PDB truncates symbol names to 255 characters.
+This is a quirk of MSVC 4.x-era PDBs, which cvdump announces by converting
+the 16-bit type pool at the top of the TYPES section."""
+
```
disinvite: I think this module should be about tests for the overall parsing of the cvdump text as a whole, versus just testing for a specific section. (i.e. `test_cvdump_types.py`). Setting the two booleans is one item we're testing here, but we're not specifically looking at whether the symbols are matched using a truncated version.

Maybe `test_cvdump_sections.py?` A lot of other name options are already in use, but let's not rename any files here.

tests/test_cvdump_symbol_truncation.py
```diff
@@ -0,0 +1,75 @@
...
+    parser = CvdumpParser()
+    parser.read_section("TYPES", TYPES_32BIT)
+
+    assert parser.is_16bit_type_pool is False
+    assert CvdumpAnalysis(parser).truncate_symbols is False
+
+
+def test_no_types_section_assumes_truncation():
+    """Without the TYPES section we cannot tell, so keep the historical
+    behavior rather than guess."""
```
disinvite: I think the docstrings should be understandable on their own instead of referencing a previous test.

tests/test_cvdump_symbol_truncation.py
```diff
@@ -0,0 +1,75 @@
...
+    parser = CvdumpParser()
+
+    assert parser.is_16bit_type_pool is None
+    assert CvdumpAnalysis(parser).truncate_symbols is True
+
+
+def test_message_belongs_to_the_types_section():
+    """The message begins with three asterisks like a section header does, but
+    it is not one. It must land in the body of the TYPES section so that the
+    parser can see it."""
```
disinvite: The docstring and test method name could use a rewrite. It's not 100% clear what we're testing. I think we could be more specific and say that we're ensuring the "Converting" line does not cause a new section split.

reccmp/cvdump/analysis.py
sandboxmo-de: Done

tests/test_cvdump_symbol_truncation.py
sandboxmo-de: Done

tests/test_cvdump_symbol_truncation.py
sandboxmo-de: Rewritten, no longer depends on the previous test

tests/test_cvdump_symbol_truncation.py
sandboxmo-de: Rewritten, lmk if you'd like to see any further changes to this test

disinvite:
My only real issue with merging this is that all the docstrings and comments have the hallmarks of AI speak. I don’t mind using it for the code, but making things human-friendly should be priority 1. Any chance you could do a pass to improve clarity? (The same feedback applies to the other PRs too.)

I don’t know any great tricks to coax Claude to do it. The ASD-STE100 meme going around does reduce the “load-bearing, it lands, byte-perfect, etc” tells, but what you get is a little too simple so it stands out for a different reason.

sandboxmo-de:
Yeah, I'll do a manual pass and rewrite some of these. Still familiarizing myself with the codebase so felt like Claude would do a better job with the docstrings than me. Will ping again after I've cleaned these up

sandboxmo-de:
Docs should be significantly better now across all 3 PRs, mix of my rewrites and Claude's rewrites

Will be stricter with doc quality for future PRs (was spending more time trying to prevent code slop than doc slop). Happy to meet whatever bar of quality you want met

Issue #509, opened 2026-08-10
New code parser


disinvite:
Tracking for multiple PRs that will implement the new code parser.

### New features
- Support all indenting styles for functions and namespace scope. The current parser uses tab stops to pair up curly brackets. (#273)
- Limited unwrapping of preprocessor (PPC) blocks, as in the following:
    - `0` and `1` expressions
      ```cpp
      #if 0
      void test_function(int x) {
      #else
      void test_function(int x, int y) {
      #endif
      ```
    - Same curly bracket sequence in every branch
      ```cpp
      #if COMPAT_MODE
        int x;
        for (x; x < 10; x++) {
      #else
        for (int x; x < 10; x++) {
      #endif
      ```
    - Disconnected curly brackets, if they are balanced across the entire file
      ```cpp
      #ifdef EXTERN_C
      extern "C" {
      #endif
      // ...
      #ifdef EXTERN_C
      };
      #endif
      ```
- Tracking (line, col) position of all elements to allow the linter to check horizontal alignment.
- More lenient parsing: the state machine design required us to abort some marker sequence instead of allowing them with a warning.
- Fixing many long-standing bugs: #55, #56, #174, and others.

PR #516, opened 2026-08-18
First round of exhaustive parser tests

disinvite:
Part of #509. I have more items I'm trying to sort into buckets that make sense. (For example: is comment/nameref parsing simple enough to be its own module, or distributed among modules for each marker type?)

Horizontal alignment and matching slash depth aren't even checked by the current parser, but the new one will make it a lot easier to add them to our syntax reporting.

Blank lines between the marker and "completion token" (my term for the comment or code line that finishes the annotation) aren't checked for all marker types. I can fix the bug pretty easily on the current parser to remove the xfails, but we may want to wait until all these new tests are in before starting the migration.

### DIFF
--- a/reccmp/parser/error.py
+++ b/reccmp/parser/error.py
@@ -57,6 +57,18 @@ class AlertCode(Enum):
     # WARN: The alias does not point to a valid marker type and has no effect.
     ALIAS_GOES_NOWHERE = 114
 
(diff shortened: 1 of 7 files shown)


jonschz:
Sorry for taking so long to respond. I don't have that much time at the moment, feel free to ping me if you want my opinion on something specific.

PR #517, opened 2026-08-18
Make `cydifflib` an optional dependency

### LINKED ISSUE #515: Cannot install
Whenever I try to install reccmp I get the following error:
```
  ╰─> [21 lines of output]
      WARNING: tool.scikit-build.metadata is deprecated; use the standard top-level [[tool.dynamic-metadata]] instead
      *** scikit-build-core 1.0.3 using CMake 4.4.2 (wheel)
      *** Configuring CMake...
      loading initial cache file C:\Users\theso\AppData\Local\Temp\tmpvnin6qxw\build\CMakeInit.txt
      -- Building for: Visual Studio 17 2022
      -- Selecting Windows SDK version 10.0.26100.0 to target Windows 10.0.19045.
      -- The C compiler identification is unknown
      -- The CXX compiler identification is unknown
      CMake Error at CMakeLists.txt:15 (project):
        No CMAKE_C_COMPILER could be found.



      CMake Error at CMakeLists.txt:15 (project):
        No CMAKE_CXX_COMPILER could be found.



      -- Configuring incomplete, errors occurred!

      *** CMake configuration failed
      [end of output]

  note: This error originates from a subprocess, and is likely not a problem with pip.
  ERROR: Failed building wheel for cydifflib
```
My cmake works in VS projects so idk what the issue is

disinvite:
Fixes #515. pyproject.toml [supports optional dependencies](https://packaging.python.org/en/latest/guides/writing-pyproject-toml/#dependencies-optional-dependencies), so if we fully migrate to that, it might make sense to have a "performance" feature set.

### DIFF
--- a/README.md
+++ b/README.md
@@ -87,6 +87,8 @@ We also have tooling to import the information from the decompilation into [Ghid
 
 We have established some [best practices](docs/recommendations.md) that have no impact on `reccmp`'s output, but have made a positive impact on the LEGO Island decompilation.
 
+[`cydifflib`](https://github.com/rapidfuzz/CyDifflib) is a drop-in replacement for Python's builtin [`difflib`](https://docs.python.org/3/library/difflib.html) module that offers better performance. If `cydifflib` is installed in the environment where you run `reccmp`, we will use it.
+
 ## Contributing
 
 Feel free to contribute to this project if you are interested! More information can be found at [CONTRIBUTING.md](./CONTRIBUTING.md).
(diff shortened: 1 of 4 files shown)


disinvite:
Is this still worth doing even though #515 is resolved? I say yes.

madebr:
I'd also keep it optional, perhaps even configurable at runtime. There will be more users running into #515. Making it optional lowers the potential support burden.

jonschz:
Adding a `performance` profile sounds good to me. `pip install reccmp` should then install what's necessary, `pip install reccmp[performance]` should also install cydifflib.

PR #518, opened 2026-08-19
roadmap.py: use csv.writer for CSV export instead of a bare join

sandboxmo-de:
`export_to_csv` wrote each row with `','.join()` and no quoting. C++ symbol names can contain literal commas — any multi-parameter template instantiation (`std::set<Foo *,FooCompare,std::allocator<Foo *> >`) or demangled multi-argument signature — so on read-back a CSV parser cannot tell the name's commas from column delimiters, and every column after the symbol shifts.

Switch to stdlib `csv.writer` with the file opened `newline=''`, matching the reader-side convention already used in `reccmp/compare/csv.py`. Fields without delimiters are written unchanged, so comma-free output is byte-identical to before.

Verified with a round-trip test (`tests/test_tools_roadmap.py`): under the previous join, a row carrying a two-parameter template name reads back as 11 columns instead of 9 and the symbol value truncates at its first comma; with `csv.writer` both round-trip intact.

### DIFF
--- a/tests/test_tools_roadmap.py
+++ b/tests/test_tools_roadmap.py
@@ -1,12 +1,4 @@
-"""Testing the reccmp-roadmap tool's CSV export.
-
-Regression coverage for the unquoted-comma bug: `export_to_csv` used to build
-each line with a bare `",".join(...)`, so any field containing a comma (most
-commonly `name`, for a multi-argument template symbol like
-`set<MxAtom *,MxAtomCompare,allocator<MxAtom *> >`) produced a row with extra
-columns once read back. In this project that silently invented dozens of fake
-"modules" on the very first parse of a real dump.
-"""
+"""Tests for the reccmp-roadmap tool's CSV export."""
 
 import csv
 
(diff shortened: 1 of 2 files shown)


tests/test_tools_roadmap.py
```diff
@@ -0,0 +1,102 @@
...
+    assert len(parsed) == 2
+    assert parsed[0]["name"] == "foo<A,B>"
+    assert parsed[0]["module"] == "ONE/a.cpp"
+    assert parsed[1]["name"] == "bar"
+    assert parsed[1]["module"] == "TWO/b.cpp"
+
+
+def test_export_to_csv_no_special_characters_unaffected(tmp_path):
+    """Values without commas or quotes round-trip unchanged. The common
+    case gets no extra quoting."""
```
disinvite: This could use another rewrite without "round-trip" as a verb. (There's another test using that term from an earlier PR, but let's not add more)

disinvite:
The three tests check the header and csv contents in three different ways. In each case we just want to know whether the same input rows can be read back again after writing the file. Maybe pick just one approach and then pull out the repeated code into a utility function?

tests/test_tools_roadmap.py
sandboxmo-de: Done

PR #521, opened 2026-08-21
Use text files for cvdump samples

disinvite:
Is it worth moving cvdump `TYPES` excerpts to their own files? I set up a very simple format for sample text files that also gives the type keys a more self-documenting name.

To demonstrate, I updated `test_ghidra_integration_type_imports.py` along with adding docstrings and (semi-redundant) comments to help things flow better visually.

My reason for doing this: I have a dozen or so samples of various inheritance patterns designed to exercise the vbase slim behavior in the Ghidra import. By necessity, the test files almost entirely consist of cvdump text, and I couldn't find a great way to structure the files that:
1. keeps relevant sample data near its tests
2. stays under 1000 lines (`pylint` limit)
3. doesn't spread loosely related tests across many files.

### DIFF
--- a/.github/workflows/format.yml
+++ b/.github/workflows/format.yml
@@ -44,6 +44,7 @@ jobs:
             '# codespell:ignore-begin.*?codespell:ignore-end *\n' \
           --ignore-regex remote-checkin-comment \
           -L SEH \
+          -S ./tests/cvdump_sample/*.txt \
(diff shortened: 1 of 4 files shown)


jonschz (approved):
Very nice! Do you want to migrate some want to migrate some other tests as well? I believe that there are more tests with long cvdump outputs.

PR #522, opened 2026-08-30
asm compare: exact-match fixes for operand equivalence, one-past-the-end references, and string extents

foxtacles:
Three fixes for false mismatches in the instruction comparison, found while verifying byte-identical rebuilds of LEGO Island with MSVC 4.2. Each one is a case where an operand pair that refers to byte-identical data (or an identical call target) scored as a difference, so a *correct* source could score lower than an incorrect one.

## The three commits

**1. `asm replacement: resolve one-past-the-end array references`**
An address exactly one past the end of an annotated DATA entity (size >= 8) now resolves to `name+size` when no other entity starts there. This is the standard compiler idiom for array end bounds (`cmp esi, offset g_array+SIZE`). Previously the orig side resolved to a raw address while the recomp side resolved to a name (or vice versa), producing a guaranteed mismatch on both correct and incorrect source.

**2. `asm compare: cross-side operand equivalence and shared table locations`**
After the initial diff, mismatching line pairs are re-examined: if two lines differ only in address operands and every operand pair shares an interpretation — the same matched entity at the same delta, or the same delta from an operand pair that already compares equal — the pair is scored as a match. The design is deliberately fail-closed:
- equivalence is strictly symmetric (a one-sided fuzzy resolution is never accepted),
- call targets and indirect slots only accept exact-entity or position-based interpretations that no annotation contradicts,
- diff opcodes are patched locally, so the ratio can only increase and rewritten lines cannot re-pair with unrelated lines.

This also shares detected jump/data table locations (function-relative) between the two disassemblies when their section layouts disagree; a table that sits before the instruction that reads it was previously disassembled as junk code, or cut the disassembly short. The re-parse is kept only if both sides converge on the same layout. Fixes cases like `__dosmaperr`, `__output` (calls to statics invisible in the PDB), and the Smacker `DoFrame` functions (0.10 -> 1.0).

**3. `compare: strip trailing nulls when naming a string entity`**
The recomp reader knows each literal's true extent from its section contribution (so embedded-null strings can be read at all) and includes the terminator; the orig reader stops at the first null. A source literal spelled `"text\0"` — which the shipped 1997 LEGO Island source provably uses in the DirectDraw error tables — therefore produced `"text\x00"` vs `"text"` entity names and a guaranteed operand mismatch. Trailing nulls are stripped when naming; interior nulls are preserved, so embedded-null matching is unaffected.

## Validation

On a rebuild of LEGO Island whose LEGO1.DLL / ISLE.EXE / CONFIG.EXE are **byte-identical to retail** (verified by hash), current master still scores 201 function rows below 100% across the three targets (198 LEGO1 / 1 ISLE / 2 CONFIG — STL `_Tree` instantiations, `ErrorToString`-style string-table functions, CRT internals). With these three commits, all three targets score every row at 100%, with zero regressions on any other row. An earlier validation against non-identical builds also showed no regressed function on any target (details in the second commit's message).

`pytest` (1076 passed), `black --check`, `pylint` (10.00), and `mypy` are clean; the new behavior is covered by `tests/test_candidate_lookup.py`, `tests/test_refine.py`, `tests/test_instgen.py`, and additions to `tests/test_name_replacement.py` and `tests/test_entity_obj.py`.

### DIFF
--- a/reccmp/compare/asm/instgen.py
+++ b/reccmp/compare/asm/instgen.py
@@ -6,7 +6,7 @@ import bisect
 import struct
 from functools import cache
 from enum import Enum, auto
-from typing import Iterable, Literal, NamedTuple
+from typing import Iterable, Literal, Mapping, NamedTuple
 from capstone import Cs, CS_ARCH_X86, CS_MODE_16, CS_MODE_32  # type: ignore
 from .const import JUMP_MNEMONICS
 
@@ -65,7 +65,20 @@ def stop_at_int3(
 
 class InstructGen:
     # pylint: disable=too-many-instance-attributes
-    def __init__(self, blob: bytes, start: int, is_32bit: bool = True) -> None:
+    def __init__(
+        self,
+        blob: bytes,
+        start: int,
+        is_32bit: bool = True,
+        seeds: Mapping[int, SectionType] | None = None,
+    ) -> None:
+        """seeds provides table locations that are known before we start
+        the analysis. A jump or data table is normally detected from the
+        instruction that reads it, but that instruction may come *after*
+        the table, in which case we have already disassembled the table
+        bytes as (junk) code. The caller can break this chicken-and-egg
+        problem by seeding the tables detected on a previous run or by the
+        disassembly of the same function in another image."""
         self.is_32bit = is_32bit
         self.blob = blob
         self.start = start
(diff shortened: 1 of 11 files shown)


foxtacles:
Separate PRs coming

PR #523, opened 2026-08-30
asm replacement: resolve one-past-the-end array references

foxtacles:
An address exactly one past the end of an annotated DATA entity (size >= 8) now resolves to `name+size` when no other entity starts at that address. This is the standard compiler idiom for an array's end bound, e.g. `cmp esi, offset g_array+SIZE`. Previously one side resolved to a raw address while the other resolved to a name (or vice versa), producing a guaranteed operand mismatch on both correct and incorrect source.

The size floor keeps small scalars out: an address just past a 4-byte entity is much more likely to be an unrelated, unannotated neighbor variable. Covered by new cases in `tests/test_name_replacement.py`, including the small-entity guard.

Found while verifying byte-identical rebuilds of LEGO Island with MSVC 4.2, where this class of false mismatch penalized correct source (mostly STL `_Tree` instantiations iterating to an array bound).

### DIFF
--- a/reccmp/compare/asm/replacement.py
+++ b/reccmp/compare/asm/replacement.py
@@ -42,11 +42,37 @@ def create_name_lookup(
 
         # We will not return an offset name if this is not a variable
         # or if the offset is outside the range of the entity.
-        if entity.entity_type not in (
-            EntityType.DATA,
-            EntityType.OFFSET,
-        ) or offset >= entity.any_size(image_id):
+        # Exception: an offset exactly one past the end of the entity is
+        # allowed, so we can identify one-past-the-end pointers (a common
+        # compiler idiom for array end bounds). This case can only be reached
+        # when no other entity starts at the given address.
+        size = entity.any_size(image_id)
+        if (
+            entity.entity_type
+            not in (
+                EntityType.DATA,
+                EntityType.OFFSET,
+            )
+            or offset > size
+        ):
             return None
+
+        if offset == size:
+            # One-past-the-end reference: a common compiler idiom for the end
+            # bound of an array. Only allow this for arrays and structs; for
+            # a scalar an address just past the entity is much more likely to
+            # be an unrelated (unannotated) neighbor variable.
+            # Without type information, fall back to a size heuristic.
+            type_key = entity.get("data_type")
+            if (
+                CvdumpTypeKey(type_key).is_scalar()
+                if type_key is not None
+                else size < 8
+            ):
+                return None
+            # Skip the data_type offset lookup (the offset is outside the
+            # type) and use the raw offset suffix.
+            return entity.match_name(f"+{offset}")
 
         type_key = entity.get("data_type")
         if type_key:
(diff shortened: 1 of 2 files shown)


reccmp/compare/asm/replacement.py
```diff
@@ -42,12 +42,32 @@ def get_name(entity: ReccmpEntity, offset: int = 0) -> str | None:
...
+            # bound of an array. Only allow this for array-sized entities;
+            # for small scalars an address just past the entity is much more
+            # likely to be an unrelated (unannotated) neighbor variable.
+            if size < 8:
+                return None
+            # Skip the data_type offset lookup (the offset is outside the
+            # type) and use the raw offset suffix.
+            return entity.match_name(f"+{offset}")
+
         type_key = entity.get("data_type")
```
disinvite, line 77: Since we are looking at the data type anyway, could we use `type_key.is_scalar()` to decide whether it's a struct/array or primitive? Instead of the size cutoff.

tests/test_name_replacement.py
```diff
@@ -86,8 +86,26 @@ def test_offset_name(db: EntityDb, entity_type: EntityType):
...
+    # One-past-the-end of an array-sized entity resolves to 'name+size'.
+    # (A common compiler idiom for the end bound of an array.)
+    assert lookup(110) == "Hello+10 (OFFSET)"
+
     # Outside the range = no name
-    assert lookup(110) is None
+    assert lookup(111) is None
+
+
+@pytest.mark.parametrize("entity_type", (EntityType.DATA, EntityType.OFFSET))
```
disinvite: Can probably ignore `EntityType.OFFSET` because that should've been refactored out in #459.

tests/test_name_replacement.py
```diff
@@ -86,8 +86,26 @@ def test_offset_name(db: EntityDb, entity_type: EntityType):
...
+def test_one_past_end_small_entity(db: EntityDb, entity_type: EntityType):
```
disinvite: We should probably test the case where there _is_ an entity at the `array+1` address.

reccmp/compare/asm/replacement.py
foxtacles: Done — the rule now asks the type system when a `data_type` is present: `CvdumpTypeKey.is_scalar()` refuses the one-past-the-end name for scalars regardless of size (a `T_REAL64` no longer slips through), and admits arrays/structs even below the old cutoff. The size heuristic only remains as the fallback for entities with no type information. Covered by `test_one_past_end_scalar_type` / `test_one_past_end_aggregate_type`.

tests/test_name_replacement.py
foxtacles: Dropped — the new tests use `EntityType.DATA` only.

tests/test_name_replacement.py
foxtacles: Added `test_one_past_end_prefers_existing_entity`: when an entity starts exactly at the end-bound address, its own name wins and the fallback never engages.

foxtacles:

PR #524, opened 2026-08-30
asm compare: cross-side operand equivalence and shared table locations

foxtacles:
> [!NOTE]
> Stacked on #523 (its commit is included here); please review only the second commit.

Two operands that point at the same thing can compare unequal today when their addresses displace differently in each image:

- one-past-the-end of an array that coincides with the start of the next variable in one image but not the other (`__dosmaperr`, `__spawnve`)
- references biased slightly before/past a variable, e.g. an array base folded with the first index, or a walking-pointer end bound anchored on a struct member (`MxSoundManager::GetAttenuation`, `Act3Cop::ParseAction`)
- references into the function's own body, e.g. its jump tables, when the function links at a different address (Smacker blob)
- calls to functions no symbol covers, e.g. statics invisible in the PDB (`__output` -> `_write_multi_char`/`_write_string`)

After the initial diff, mismatching line pairs are re-examined: if two lines differ only in address operands and every operand pair shares an interpretation — the same matched entity at the same delta, or the same delta from an operand pair that already compares equal — the pair is scored as a match. The design is deliberately fail-closed: equivalence is strictly symmetric (a one-sided fuzzy resolution is never accepted); call targets and indirect slots only accept exact-entity or position-based interpretations that no annotation contradicts; and the diff opcodes are patched locally, so the ratio can only increase and rewritten lines cannot re-pair with unrelated lines.

This also shares detected jump/data table locations (function-relative) between the two disassemblies when their section layouts disagree. A table sitting before the instruction that reads it was previously disassembled as junk code with image-specific relocated bytes, or cut the disassembly short at an undecodable byte; both Smacker `DoFrame` functions were affected (0.10 -> 1.0). The re-parse is kept only if both sides converge on the same layout.

Validation on the LEGO Island project (zero regressed rows on any target): on a rebuild whose binaries are byte-identical to retail, this plus #523 and #525 takes every remaining sub-100% row to 100% (previously 198 LEGO1 / 1 ISLE / 2 CONFIG rows scored below 100% despite identical bytes). An earlier validation against non-identical builds showed 9 improved + 7 new 100% on LEGO1 and 116 improved + 22 new 100% on BETA10. New coverage in `tests/test_refine.py`, `tests/test_candidate_lookup.py`, `tests/test_instgen.py`.

### DIFF
--- a/reccmp/compare/asm/instgen.py
+++ b/reccmp/compare/asm/instgen.py
@@ -6,7 +6,7 @@ import bisect
 import struct
 from functools import cache
 from enum import Enum, auto
-from typing import Iterable, Literal, NamedTuple
+from typing import Iterable, Literal, Mapping, NamedTuple
 from capstone import Cs, CS_ARCH_X86, CS_MODE_16, CS_MODE_32  # type: ignore
 from .const import JUMP_MNEMONICS
 
@@ -65,7 +65,20 @@ def stop_at_int3(
 
 class InstructGen:
     # pylint: disable=too-many-instance-attributes
-    def __init__(self, blob: bytes, start: int, is_32bit: bool = True) -> None:
+    def __init__(
+        self,
+        blob: bytes,
+        start: int,
+        is_32bit: bool = True,
+        seeds: Mapping[int, SectionType] | None = None,
+    ) -> None:
+        """seeds provides table locations that are known before we start
+        the analysis. A jump or data table is normally detected from the
+        instruction that reads it, but that instruction may come *after*
+        the table, in which case we have already disassembled the table
+        bytes as (junk) code. The caller can break this chicken-and-egg
+        problem by seeding the tables detected on a previous run or by the
+        disassembly of the same function in another image."""
         self.is_32bit = is_32bit
         self.blob = blob
         self.start = start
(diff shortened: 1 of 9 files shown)


foxtacles:
Revised after a deep review pass:

- **Added**: a contradiction veto for non-strict operands — two operands that each sit exactly on *different* matched entities are never equated by position (`test_refine_conflicting_exact_candidates`).
- **Added**: boundary tests for `ANCHOR_DELTA_LIMIT` (exactly at / one past) and the strict-interior `self_reach` bound, plus an audit counter: refined pair counts are logged per function.
- **Evaluated and rejected**: extending the strict-operand veto to *unmatched* PDB entities at the call target. On the LEGO Island corpus this vetoes large numbers of legitimate position-equated pairs (EH handlers and CRT internals that are named in the PDB but have no orig-side annotation to match), e.g. `_WinMainCRTStartup` drops below 100% on a byte-identical build. The matched-entity veto is the sound boundary: only a matched entity makes a cross-address claim that position can contradict.

Full validation after the revision: all three LEGO Island targets still score every row at 100% on a byte-identical rebuild, with zero regressed rows.

disinvite:
I think we should split this into at least two component PRs: one for the improved jump table (switch) detection and one for the post processing to fix the array+1 diffs.

For the second, maybe we could run ParseAsm on each single instruction with a diff and use an alternate AddrLookupProtocol and/or NameLookupProtocol function that will specifically use the preceding entity with the extra offset added, but only if it is an array or struct that would qualify. (Those may not be the actual names of the protocols but I can do only so much from the phone browser.)

Retaining the addrs before replacement/sanitization is worth exploring later as its own thing. Since we are only concerned with a specific type of diff I think it’s too complex to add now. Unless there is a nuance I’ve missed and there’s a third category of diff to solve.

foxtacles:

PR #525, opened 2026-08-30
compare: strip trailing nulls when naming a string entity

foxtacles:
The two images do not report string extents the same way, and the asymmetry makes a *correct* source score lower than an incorrect one.

In the recompiled image each literal's true extent is known from its section contribution, and `create_recomp_strings` deliberately uses it so strings containing embedded nulls can be read at all. In the original image there is no such metadata, so the reader stops at the first null. A source literal spelled `"Action not supported.\0"` therefore becomes entity name `"Action not supported.\x00"` on the recomp side but `"Action not supported."` on the orig side; both names are substituted into the disassembly, so an instruction pair referring to byte-identical data compares as a difference.

That is not hypothetical: the LEGO Island DirectDraw error tables are provably spelled with the trailing `\0` in the shipped 1997 source (each of those literals occupies `round_up(strlen+2, 4)` in `.data` while every other literal in the same object occupies `round_up(strlen+1, 4)`). Restoring that spelling reproduces all 91 string addresses exactly, yet dropped `MxDirectDraw::ErrorToString` and `MxDeviceEnumerate::EnumerateErrorToString` from 1.0 to 0.79.

Trailing nulls are storage, not content: a C string's identity for comparison ends at its terminator. Interior nulls are preserved, so the embedded-null matching that motivates using the known extent is unaffected. New cases in `tests/test_entity_obj.py` cover trailing, interior, and all-null strings.

### DIFF
--- a/reccmp/compare/db.py
+++ b/reccmp/compare/db.py
@@ -10,8 +10,23 @@ from reccmp.types import EntityType, ImageId
 
 def entity_name_from_string(text: str, wide: bool = False) -> str:
     """Create an entity name for the given string by escaping
-    control characters and double quotes, then wrapping in double quotes."""
-    escaped = text.encode("unicode_escape").decode("utf-8").replace('"', '\\"')
+    control characters and double quotes, then wrapping in double quotes.
+
+    TRAILING null characters are stripped first, because the two images do not
+    report string extents the same way. In the recompiled image we know each
+    literal's true extent from its section contribution, so a source literal
+    spelled `"text\\0"` is read back as 'text\\x00'. In the original image there
+    is no such metadata and the reader stops at the first null, giving 'text'.
+    Substituting those two into the disassembly would make an operand pair that
+    refers to identical bytes compare as a difference -- i.e. it would score a
+    CORRECT source lower than an incorrect one. Trailing nulls are storage, not
+    content: a C string's identity for comparison purposes ends at its
+    terminator. Nulls in the INTERIOR are preserved, so the ability to read and
+    match strings containing embedded nulls (the reason the known extent is
+    used in the first place) is unaffected."""
+    escaped = (
+        text.rstrip("\0").encode("unicode_escape").decode("utf-8").replace('"', '\\"')
+    )
     return f'{"L" if wide else ""}"{escaped}"'
 
 
--- a/tests/test_compare_analyze.py
+++ b/tests/test_compare_analyze.py
@@ -314,6 +314,31 @@ def test_complete_partial_strings_with_nulls(db: EntityDb):
     assert e.name == '"\\x00test"'
 
 
+def test_complete_partial_strings_trailing_null_spelling(db: EntityDb):
+    """A literal spelled "text\0" stores one more byte, which only the sized
+    (section contribution) reader can see; the unsized reader stops at the
+    first null. Both must produce the same name for the same bytes."""
+    binfile = Mock(spec=[])
+    binfile.read = Mock(return_value=b"text\x00\x00")
+    binfile.read_string = Mock(return_value=b"text")
+
+    with db.batch() as batch:
+        # Contribution-sized, like a recomp string from the PDB.
+        batch.set(ImageId.ORIG, 100, type=EntityType.STRING, size=6)
+        # No size metadata, like an orig string.
+        batch.set(ImageId.ORIG, 200, type=EntityType.STRING)
+
+    complete_partial_strings(db, ImageId.ORIG, binfile)
+
+    sized = db.get(ImageId.ORIG, 100)
+    unsized = db.get(ImageId.ORIG, 200)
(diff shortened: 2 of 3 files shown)


disinvite:
The AI slop obscures the actual point 😀

Are we saying that: 
1. Padding may or may not be created in the real image by adding trailing `\0` bytes to a string.
2. We should strip the null bytes in the "create name" function so that these cases match in reccmp. (i.e. the string would behave the same in the real program no matter how many `\0` chars are there.)
3. Any diff that can be solved by adding or removing `\0` is an alignment problem to be detected by some other tool (maybe roadmap?)

foxtacles:
Fair 😄 — concisely:

1. **Yes.** An explicit trailing `\0` makes the literal one byte longer before 4-byte alignment, so it shifts padding and every following string address. The 1997 source really spelled the DirectDraw error tables that way — isle has 92 of these in `LEGO1/mxdirectx/mxdirectdraw.cpp` and `mxdirectxinfo.cpp`, e.g. `return "Action not supported.\0";` (mxdirectdraw.cpp:1129) — and in retail those literals occupy `round_up(strlen+2, 4)` in `.data` while every other literal in the same object occupies `round_up(strlen+1, 4)`.

2. **Yes, with one nuance:** it's not about runtime behavior, it's that the two sides *name* the same bytes differently. The recomp reader knows the literal's extent from the section contribution → `"Action not supported.\x00"`; the orig reader has no size info and stops at the first null → `"Action not supported."`. Identical bytes, two names, guaranteed operand diff. Stripping trailing nulls at naming time makes them agree; interior nulls stay, since they're content (and the reason the known extent is read at all).

3. **Right — nothing is lost.** The layout consequence of a missing/extra `\0` is still fully visible without the name: all the following literals shift, and every function referencing them diffs on the addresses themselves (that's exactly how the isle spelling was found). What this PR fixes is the inversion: the *correct* spelling reproduces all 91 string addresses exactly, yet scored `ErrorToString`/`EnumerateErrorToString` at 0.79, while the wrong spelling scored 1.00.

🤖 via [Claude Code](https://claude.com/claude-code)

disinvite:
I don't love the reasoning in answer 1. Are we saying that the section contribution size is accurate with respect to `\0` padding in the source string? Or not?

I would think claude can generate enough samples to back up the claim with data either way.

foxtacles:
Yes — the section contribution is byte-accurate for the stored literal: an explicit `\0` is one more stored byte, and alignment padding is never included. Data:

**Compiler ground truth** (MSVC 4.2 `cl /c`, `.data` of the object, literal boundaries annotated):

| source spelling | stored bytes |
|---|---|
| `"abc"` | 4 — `61 62 63 00` |
| `"abc\0"` | 5 — `61 62 63 00 00` |
| `"Action not supported."` | 22 — `… 2e 00` |
| `"Action not supported.\0"` | 23 — `… 2e 00 00` |
| `"in\0side"` | 8 — `69 6e 00 73 69 64 65 00` |

The zeros between literals that pad to 4-byte alignment sit outside every symbol.

**Project scale** (LEGO1 debug PDB — the same section contributions reccmp reads): 1,593 string contributions total; **1,501 are exactly `strlen+1`** and **exactly 92 are `strlen+2`** — precisely the DirectDraw error tables (`0x100ff250 contribution=27 "Unrecognized error value.\x00\x00"`, etc.). Contribution values are unaligned (27, 71, 78, 114…), so padding is provably excluded.

So the recomp side genuinely knows `"…\0"` from `"…"`; the orig side can't (it stops at the first null), which is the whole asymmetry — and why the strip happens only at naming time.

foxtacles:
Also added `test_complete_partial_strings_trailing_null_spelling`: both readers over the same bytes (`b"text\x00\x00"`) — one sized by the contribution, one stopping at the first null — must produce the same name, with the true extent preserved on the sized entity.

disinvite:
This CSV should fix the problem in `LEGO1`, after adding the `\0` to the error strings in `MxDeviceEnumerate` and `MxDirectDraw`. You would also need to add the annotations for `CONFIG`.
[lego1-mxdirectdraw-strings.csv](https://github.com/user-attachments/files/31622913/lego1-mxdirectdraw-strings.csv)

I think we are doing everything correct without this change. The problem is that the DirectDraw error strings aren't marked with `// STRING`, so we are creating the entities via automated search and cannot guess how many trailing nulls there are. If the strings are marked, we add any `\0` chars to the orig entity size, along with the null terminator. (We may or may not be doing this correctly for widechar: I will add some tests later)

foxtacles:
Implemented your way instead — the LEGO1 CSV (thanks for generating it) plus an equivalent CONFIG one produced from the debug PDB contributions now live in the isle project as `data-sources`, so both sides derive identical names and no naming change is needed here. Validated end to end: all three targets score every row at 100% with a reccmp build that does not include this change.

PR #526, opened 2026-08-30
crt_startup: pair empty-fingerprint startup functions by array position

foxtacles:
The fingerprint matcher pairs CRT startup functions by the entities their code references. A startup function that references no known entity — a CRT-internal initializer such as `__onexitinit`, or a compiler-generated `$E` thunk registering the `CxxUnhandledExceptionFilter` pair — produces an empty fingerprint on both sides and can never be paired, so it is reported as an unmatched stub even when the two images agree completely.

The startup arrays themselves are populated in link order, which a faithful recompilation preserves. This change pairs entries whose fingerprint is empty on **both** sides by their array position, guarded three ways: both arrays must have the same length; every fingerprint match must agree with the shared order (any contradiction disables the fallback entirely); and a one-sided empty fingerprint is treated as a disagreement, not evidence. Entries with a non-empty but ambiguous fingerprint are left unmatched as before.

On the LEGO Island project this resolves the last unmatched LEGO1 functions (`Implemented 4488/4491 -> 4489/4489`; the two initializers gain compared rows and the matched thunk is excluded as usual) and both ISLE initializers, with no change to any other row. Covered by six new cases in `tests/test_analysis_crt_startup.py`, including the length, one-sidedness, and order-violation guards.

### DIFF
--- a/reccmp/analysis/crt_startup.py
+++ b/reccmp/analysis/crt_startup.py
@@ -301,6 +301,35 @@ def create_crt_matches(
         if not matched_this_pass:
             break
 
+    # Positional fallback for functions the fingerprints cannot pair: startup
+    # arrays are populated in link order, which a faithful recompilation
+    # preserves, so entries whose fingerprint is EMPTY on both sides (their
+    # code references no known entity; common for CRT-internal initializers)
+    # can be paired by their position in the array. Only do this when both
+    # arrays have the same length, every fingerprint match agrees with the
+    # shared order, and both entries at the position are still unmatched.
+    # Entries with a non-empty but ambiguous fingerprint are left unmatched
+    # as before.
+    orig_order = list(orig_array.functions)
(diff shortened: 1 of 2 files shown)


disinvite:
Is this a case where `__onexitinit` (for example) is not visible in the PDB or any recomp artifacts? So it would be impossible to solve this by adding metadata to the orig side?

foxtacles:
Mixed. `___onexitinit` is a PDB public, so orig-side metadata could match that one. The C++ init entry (the `$E` routine registering the `CxxUnhandledExceptionFilter` pair) has no symbol in the PDB or any recomp artifact — it's a static in a prebuilt CRT object — so name metadata can't reach it. The fingerprint can't either: it samples matched DATA entities only, and this routine references only two functions and an import slot. The array order is the only evidence available.

foxtacles:

PR #527, opened 2026-08-30
compare: anchor annotated statics inside preserved spans

foxtacles:
A static function (e.g. a CRT-internal helper like `_write_multi_char`) is invisible in the PDB, so an annotation for it can never be matched by name — it is reported as an unmatched stub forever, even when the rebuilt code is identical.

Its address is still fully determined by its object's section contribution: it keeps its offset relative to the surrounding, matchable functions as long as that span is reproduced. This change matches each unmatched annotated FUNCTION entity that lies between two matched functions whose span has the same length in both images, at the same offset in the recompiled image. Guards:

- the enclosing span must have exactly the same length on both sides (a changed contribution disables the anchoring);
- the anchored recomp address must be vacant — any existing entity there means the symbol table disagrees, and the annotation may then only match by name;
- the regular function comparison still scores the new match, so a wrong anchor shows up as a mismatched row instead of passing silently;
- the recomp side receives the orig-side extent bound as its size so the match is actually comparable (matches without any size are dropped from the report).

On the LEGO Island project this resolves the two remaining ISLE stubs, `_write_multi_char` and `_write_string` (both compare at 100%), with no change to any other row. Eight new cases in `tests/test_compare_mutate.py` cover the span, vacancy, stub, ordering, and multi-static rules.

### DIFF
--- a/reccmp/compare/core.py
+++ b/reccmp/compare/core.py
@@ -61,6 +61,7 @@ from .mutate import (
     name_thunks,
     unique_names_for_overloaded_functions,
     match_crt_startup,
+    match_span_anchored_functions,
     set_max_size,
 )
 from .verify import (
@@ -192,6 +193,7 @@ class Compare:
             set_max_size(self._db, img_id)
 
         match_crt_startup(self._db, self.orig_bin, self.recomp_bin)
+        match_span_anchored_functions(self._db)
         check_vtables(self._db)
         match_ref(self._db, self.report)
         unique_names_for_overloaded_functions(self._db)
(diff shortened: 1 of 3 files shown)


foxtacles:
Closing in favor of configuration: the two affected functions (`_write_multi_char`, `_write_string`) are now omitted from the ISLE report via the existing `report.ignore-functions` mechanism in the project's `reccmp-project.yml`. Their bytes remain fully verified by the project's byte-exact reproduction gate, so only per-function report granularity is lost, and no new matching heuristic is needed in reccmp. If a signature-based approach for PDB-invisible statics materializes later, that would supersede both this PR and the exclusion.

PR #528, opened 2026-08-30
Ghidra import: testing various inheritance patterns

disinvite:
Part of #106. The Ghidra importer uses the structures from the types db directly. I want to push these behind an API so we can swap out the mechanism that loads debug artifacts. But we need more tests first.

Claude created these sample .cpp files with various inheritance patterns. I compiled each one and generated the cvdump `TYPES` output with both MSVC 4.2 and 6. (They were identical.) The docstrings and comments are all mine. One of my goals for doing this was to understand how this works and why we can't assume the location of the indirect virtual base class without reading the `vbtable`. (Certainly doable in a follow-up.)

This corrects an "issue" where we created slim copies of base classes where this was not necessary. (I say "issue" because there was not an observable difference.) For example: if class D has base classes B and C, but only B uses virtual inheritance, we do not need a slim copy of C.

### DIFF
--- a/reccmp/ghidra/importer/type_importer.py
+++ b/reccmp/ghidra/importer/type_importer.py
@@ -238,8 +238,15 @@ class PdbTypeImporter:
 
         class_size: int = type_in_pdb["size"]
         raw_name: str = type_in_pdb["name"]
-        # Only a class with a virtual base of its own needs the slim treatment.
-        # A sibling base class without one fits at the offset given for it.
+        # Virtual inheritance requires that struct members are arranged in this order:
+        #
+        # 1. Members from non-virtual base classes.
+        # 2. Members from the derived class.
+        # 3. Members from virtual base classes.
+        #
+        # If this class has a direct virtual base class and we are creating it as the base for something else
+        # (i.e. if `as_base_class` is True) then we need to create a "slim" copy that contains only the
+        # non-virtual base classes and its own members.
         make_slim = as_base_class and "vbase" in field_list
         if make_slim:
             raw_name += "_vbase_slim"
--- a/reccmp/ghidra/importer/type_importer.py
+++ b/reccmp/ghidra/importer/type_importer.py
@@ -306,8 +306,6 @@ class PdbTypeImporter:
         components.sort(key=lambda c: c.offset)
 
         if make_slim:
-            # Make a "slim" version: shrink the size to the fields that are actually present.
-            # This makes a difference when the current class uses virtual inheritance
             assert (
                 len(components) > 0
             ), f"Error: {sanitized_name} should not be empty. There must be at least one direct or indirect vbase pointer."
--- a/reccmp/ghidra/importer/type_importer.py
+++ b/reccmp/ghidra/importer/type_importer.py
@@ -80,7 +80,7 @@ class PdbTypeImporter:
         return self.extraction.compare.types
 
     def import_pdb_type_into_ghidra(
-        self, type_index: CvdumpTypeKey, as_base_class: bool = False
+        self, type_index: CvdumpTypeKey, *, as_base_class: bool = False
     ) -> DataType:
         """
         Recursively imports a type from the PDB into Ghidra.
@@ -171,7 +171,9 @@ class PdbTypeImporter:
             type_index,
             referenced_type,
         )
-        return self.import_pdb_type_into_ghidra(referenced_type, as_base_class)
+        return self.import_pdb_type_into_ghidra(
+            referenced_type, as_base_class=as_base_class
+        )
 
     def _import_array(self, type_pdb: CvdumpParsedType) -> DataType:
         inner_type = self.import_pdb_type_into_ghidra(type_pdb["array_type"])
--- a/reccmp/ghidra/importer/type_importer.py
+++ b/reccmp/ghidra/importer/type_importer.py
@@ -80,18 +80,18 @@ class PdbTypeImporter:
         return self.extraction.compare.types
 
     def import_pdb_type_into_ghidra(
-        self, type_index: CvdumpTypeKey, slim_for_vbase: bool = False
+        self, type_index: CvdumpTypeKey, as_base_class: bool = False
     ) -> DataType:
         """
         Recursively imports a type from the PDB into Ghidra.
         @param type_index Either a scalar type like `T_INT4(...)` or a PDB reference like `0x10ba`
-        @param slim_for_vbase If `True`, the current invocation
-            imports a superclass of some class where virtual inheritance is involved (directly or indirectly).
-            This case requires special handling: Let's say we have `class C: B` and `class B: virtual A`. Then cvdump
+        @param as_base_class If `True`, the current invocation
+            imports a superclass to be embedded in some derived class. Virtual inheritance may
+            be involved: Let's say we have `class C: B` and `class B: virtual A`. Then cvdump
             reports a size for B that includes both B's fields as well as the A contained at an offset within B,
             which is not the correct structure to be contained in C. Therefore, we need to create a "slim" version of B
             that fits inside C.
-            This value should always be `False` when the referenced type is not (a pointer to) a class.
+            This value should always be `False` when the referenced type is not a class.
         """
         if type_index.is_scalar():
             return self._import_scalar_type(type_index)
@@ -107,17 +107,20 @@ class PdbTypeImporter:
 
         # follow forward reference (class, struct, union)
         if type_pdb.get("is_forward_ref", False):
-            return self._import_forward_ref_type(type_index, type_pdb, slim_for_vbase)
+            return self._import_forward_ref_type(type_index, type_pdb, as_base_class)
+
+        if type_category in ["LF_CLASS", "LF_STRUCTURE"]:
+            return self._import_class_or_struct(type_pdb, as_base_class)
+
+        assert (
+            not as_base_class
+        ), f"Tried to import {type_index} ({type_category}) as base class"
 
         if type_category == "LF_POINTER":
             return get_or_add_pointer_type(
                 self.api,
-                self.import_pdb_type_into_ghidra(
-                    type_pdb["element_type"], slim_for_vbase
-                ),
+                self.import_pdb_type_into_ghidra(type_pdb["element_type"]),
             )
-        elif type_category in ["LF_CLASS", "LF_STRUCTURE"]:
-            return self._import_class_or_struct(type_pdb, slim_for_vbase)
         elif type_category == "LF_ARRAY":
             return self._import_array(type_pdb)
         elif type_category == "LF_ENUM":
@@ -150,7 +153,7 @@ class PdbTypeImporter:
         self,
         type_index: CvdumpTypeKey,
         type_pdb: CvdumpParsedType,
-        slim_for_vbase: bool = False,
+        as_base_class: bool = False,
     ) -> DataType:
         referenced_type = type_pdb.get("udt") or type_pdb.get("modifies")
         if referenced_type is None:
@@ -168,7 +171,7 @@ class PdbTypeImporter:
             type_index,
             referenced_type,
         )
-        return self.import_pdb_type_into_ghidra(referenced_type, slim_for_vbase)
+        return self.import_pdb_type_into_ghidra(referenced_type, as_base_class)
 
     def _import_array(self, type_pdb: CvdumpParsedType) -> DataType:
         inner_type = self.import_pdb_type_into_ghidra(type_pdb["array_type"])
@@ -226,14 +229,17 @@ class PdbTypeImporter:
     def _import_class_or_struct(
         self,
         type_in_pdb: CvdumpParsedType,
-        slim_for_vbase: bool = False,
+        as_base_class: bool = False,
     ) -> DataType:
         field_list_type = type_in_pdb["field_list_type"]
         field_list = self.types.from_key(field_list_type)
 
         class_size: int = type_in_pdb["size"]
         raw_name: str = type_in_pdb["name"]
-        if slim_for_vbase:
+        # Only a class with a virtual base of its own needs the slim treatment.
+        # A sibling base class without one fits at the offset given for it.
+        make_slim = as_base_class and "vbase" in field_list
+        if make_slim:
             raw_name += "_vbase_slim"
         sanitized_name = sanitize_name(raw_name)
 
@@ -299,7 +305,7 @@ class PdbTypeImporter:
 
         components.sort(key=lambda c: c.offset)
 
-        if slim_for_vbase:
+        if make_slim:
             # Make a "slim" version: shrink the size to the fields that are actually present.
             # This makes a difference when the current class uses virtual inheritance
             assert (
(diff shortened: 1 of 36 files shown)


reccmp/ghidra/importer/type_importer.py
```diff
@@ -80,18 +80,18 @@ def types(self):
...
-            imports a superclass of some class where virtual inheritance is involved (directly or indirectly).
-            This case requires special handling: Let's say we have `class C: B` and `class B: virtual A`. Then cvdump
+        @param as_base_class If `True`, the current invocation
+            imports a superclass to be embedded in some derived class. Virtual inheritance may
+            be involved: Let's say we have `class C: B` and `class B: virtual A`. Then cvdump
             reports a size for B that includes both B's fields as well as the A contained at an offset within B,
             which is not the correct structure to be contained in C. Therefore, we need to create a "slim" version of B
             that fits inside C.
-            This value should always be `False` when the referenced type is not (a pointer to) a class.
+            This value should always be `False` when the referenced type is not a class.
```
jonschz: I think a pointer to a class will transitively call this. Then again, it's been a while, I don't remember the details

tests/ghidra_integration_test_setup.py
```diff
@@ -168,6 +170,39 @@ def set_up_cvdump_types(self, cvdump_types: str):
         load_cvdump_types(analysis, self.compare.types)
 
 
+def as_structure(data_type: "DataType") -> "Structure":
```
jonschz: Matter of taste, but I would move these helpers into a dedicated file since this one already has quite some complexity

tests/cvdump_sample/cpp/base_padding.cpp
jonschz, line 1: Maybe add a short README.md to the `cvdump_sample` directory, including
- how to compile these (toolchain etc.)
- how to make the `.txt` files out of them (doesn't need to duplicate the documentation in `__init__.py`, but should link to it and maybe enhance it)

tests/test_ghidra_integration_inheritance.py
```diff
@@ -0,0 +1,543 @@
...
+
+        A           size 12
+        B           size 12
+        C : A, B    size 28
+
+    Should create full-sized structs for A and B when creating C.
+    """
+    sample = load_cvdump_sample("multiple-inheritance")
+    type_helper.set_up_cvdump_types(sample.text)
+    leaf = type_helper.type_importer.import_pdb_type_into_ghidra(sample.key("class-c"))
```
jonschz:
```suggestion
    class_c = type_helper.type_importer.import_pdb_type_into_ghidra(sample.key("class-c"))
```
maybe something like this? You can make out from the "class-c" what `leaf` is supposed to be, this will just make it more explicit

tests/test_ghidra_integration_inheritance.py
```diff
@@ -0,0 +1,543 @@
...
+
+def test_duplicate_base_appears_twice(type_helper: GhidraTypeTestHelper):
+    """cvdump_sample/cpp/diamond_duplicate.cpp
+
+        A           size 8
+        B : A       size 12
+        C : A       size 12
+        D : B, C    size 28
+
+    Diamond inheritance pattern.
```
jonschz:
```suggestion
    Diamond inheritance pattern without virtual inheritance.
```

tests/test_ghidra_integration_inheritance.py
```diff
@@ -0,0 +1,543 @@
...
+def test_slim_vbase_at_offset_zero(type_helper: GhidraTypeTestHelper):
+    """cvdump_sample/cpp/vbase_simple.cpp
+
+        A                size 8
+        B : virtual A    size 16
+        C : B            size 20
+
+    B is 16 bytes because it contains its own members (8 bytes) followed by A.
+    The field list for C places C's first member at offset 8. Therefore, the
+    full-sized B will not fit and we need to create a "slim" copy.
```
jonschz:
```suggestion
    B is 16 bytes because it contains its own members (8 bytes) followed by some room to accomodate an A.
    The field list for C places C's first member at offset 8. Therefore, the
    full-sized B will not fit, hence the B parent within C needs to be a "slim" copy (i.e. without the room for A).
```
maybe? I'm not entirely happy with either formulation

tests/test_ghidra_integration_inheritance.py
```diff
@@ -0,0 +1,543 @@
...
+    leaf = type_helper.type_importer.import_pdb_type_into_ghidra(sample.key("class-c"))
+
+    assert leaf.getLength() == 20
+    assert components_of(leaf) == [
+        (0, "base", "B_vbase_slim"),
+        (8, "m_c0", "int"),
+        # A?
+    ]
+
+    slim = component(leaf, 0)
```
jonschz:
```suggestion
    slim_b = component(leaf, 0)
```
maybe?

tests/test_ghidra_integration_inheritance.py
```diff
@@ -0,0 +1,543 @@
...
+
+def test_slim_vbase_at_nonzero_offset(type_helper: GhidraTypeTestHelper):
+    """cvdump_sample/cpp/vbase_second_base.cpp
+
+        A                size 8
+        B                size 12
+        C : virtual A    size 16
+        D : B, C         size 32
+
+    C needs a slim copy because it has a virtual base. The copy is at offset 12
```
jonschz:
```suggestion
    The C parent within D needs to be a slim copy because it has a virtual base. The copy is at offset 12
```
or something along these lines

tests/test_ghidra_integration_inheritance.py
```diff
@@ -0,0 +1,543 @@
...
+    """cvdump_sample/cpp/diamond_virtual.cpp
+
+        A                size 8
+        B : virtual A    size 16
+        C : virtual A    size 16
+        D : B, C         size 28
+
+    Diamond inheritance pattern with virtual inheritance.
+    D contains only one copy of A: B and C point to the single A
+    using their vbtables.
```
jonschz:
```suggestion
    using their vbtables. Notably, `sizeof(D) < sizeof(B) + sizeof(C)`.
```

tests/test_ghidra_integration_inheritance.py
```diff
@@ -0,0 +1,543 @@
...
+
+
+@pytest.mark.xfail(reason="The importer does not place virtual base classes yet")
+def test_padding_between_two_virtual_bases(type_helper: GhidraTypeTestHelper):
+    """cvdump_sample/cpp/vbase_padding.cpp
+
+        A                           size 8
+        B                           size 1
+        C : virtual A, virtual B    size 25
+        D : virtual B, virtual A    size 32
```
jonschz, line 544: nice example!

reccmp/ghidra/importer/type_importer.py
```diff
@@ -80,18 +80,18 @@ def types(self):
         return self.extraction.compare.types
 
     def import_pdb_type_into_ghidra(
-        self, type_index: CvdumpTypeKey, slim_for_vbase: bool = False
+        self, type_index: CvdumpTypeKey, as_base_class: bool = False
```
jonschz: For style points: How about making `as_base_class` a keyword-only argument? (optional)

reccmp/ghidra/importer/type_importer.py
```diff
@@ -226,14 +229,17 @@ def _import_enum(self, type_pdb: CvdumpParsedType) -> DataType:
...
+        as_base_class: bool = False,
     ) -> DataType:
         field_list_type = type_in_pdb["field_list_type"]
         field_list = self.types.from_key(field_list_type)
 
         class_size: int = type_in_pdb["size"]
         raw_name: str = type_in_pdb["name"]
-        if slim_for_vbase:
+        # Only a class with a virtual base of its own needs the slim treatment.
+        # A sibling base class without one fits at the offset given for it.
```
jonschz: Not exactly sure what you mean by the "sibling". Maybe an example with classes A, B, C, D like in the tests would be helpful here? Also, it's not immediately clear if `a class` refers to the class currently being imported or its parent. Maybe:

```suggestion
        # Situation: `class C: B` and `class B: virtual A`
        # The next check will be `True` when the import of `A` recursively imports it `B` parent.
        # `A` might not have enough contiguous space to accomodate the "full" `B` (consisting of both `B`'s fields and `B`'s room to accommodate the virtual `C` parent).
        # Therefore, we need to import a slim version of `B`, which consists of `B`'s fields only.
        # <potential sibling sentence>
```

reccmp/ghidra/importer/type_importer.py
```diff
@@ -299,7 +305,7 @@ def _import_class_or_struct(
 
         components.sort(key=lambda c: c.offset)
 
-        if slim_for_vbase:
+        if make_slim:
             # Make a "slim" version: shrink the size to the fields that are actually present.
             # This makes a difference when the current class uses virtual inheritance
```
jonschz:
```suggestion
        if make_slim:
```
This comment no longer adds much value in my opinion

jonschz (approved):
Very nice overall! The logic change looks sound to me. This really would have benefitted from unit tests back when I wrote the code.

I really like the cvdump + source code approach. Not sure if there's a good way of keeping them in sync, but the current setting (i.e. try to remember to update both when changing any one of them) will work for me. Building the cvdump files dynamically feels overkill to me.

disinvite:
I rewrote a lot of the docstrings to try and express that:
1. Direct and indirect virtual inheritance requires that _all_ virtually inherited members appear last in the struct list
2. We choose to embed the parent structs instead of copying members, but virtual inheritance requires the slim copy of a base class without the gap left for its virtually inherited members

Let me know what you think. There's one reference to "sibling" still in there. By that I only meant that it is another base class at the same level, as opposed to a grandparent class.

tests/test_ghidra_integration_inheritance.py
```diff
@@ -1,32 +1,45 @@
 """Tests for importing classes with various inheritance patterns.
 
-Parent classes are embedded in the struct for the derived class instead of
-copying their members instead.
+By convention, the `vftable` pointer appears at offset 0 of a class with virtual functions.
```
jonschz, line 3: Whose convention? MSVC's?

tests/test_ghidra_integration_inheritance.py
```diff
@@ -1,32 +1,45 @@
...
 
-Vftable pointers are set to offset 0 of the class with virtual functions.
+To represent inheritance, we create the base class struct (or use one created earlier in the run)
+and set its position to the offset given in the PDB fieldlist. The other option is to copy members
+from the base classes with new offsets, but we don't do it this way.
```
jonschz, line 7: It's good to explain that, maybe also mention why? From the top of my head, polymorphism is one of the arguments (e.g. calling `void f(A a)` when you have an instance of `class B: A`), and also that it's easier to track where each property comes from

tests/test_ghidra_integration_inheritance.py
```diff
@@ -459,27 +505,30 @@ def test_multiple_inheritance_with_virtual_functions(
         B           size 8
         C : A, B    size 24
 
-    A, B, and C each introduce a virtual function.
-    C's virtual function is combined in A's vtable at offset 0 in C.
-    The vtable for B's functions is at offset 12 in C.
+    A, B, and C each introduce a virtual function. Where do we create the `vftable` member?
+    According to the PDB, only the fieldlists for B and C have a VFUNCTAB leaf, so only their
+    structs get the member. The effect is that C's virtual function is "merged" into A's vtable.
+    (This may or may not be correct and we should revisit when we create a vtable struct during import.)
```
jonschz, line 511: Off-topic - I believe the way it works is that C has two vftables:
- The one inherited from A: This is where A's functions are inherited / overwritten and where C's newly introduced virtual functions go
- The one inherited from B (at offset 12): This is where B's functions are inherited / overwritten

Let's say I have a function `void f(B b)`. If I call `f(c)`, then a vtordisp is inserted into that function call which introduces an offset of 12. But if `f` calls any of `B`'s virtual functions, it has to call the one that's overridden by `C`, so the overrides must take place in `B`'s vftable within `C`.

In any case, let's not dwell on it, the comment is perfectly fine

jonschz (approved):
Looks good! Feel free to merge after addressing the final comments, no second re-review needed.

tests/test_ghidra_integration_inheritance.py
disinvite: In terms of Ghidra, though, does it make a difference to the decompiler output? (copying the members versus embedding the structs)

jonschz: Yes, it does make a difference to the decompiled pseudocode:
grafik
All of the `->base` and `.base` wouldn't be there with embedding.

madebr: Embedding would make creating vtables much easier.

jonschz: > Embedding would make creating vtables much easier.

You mean, as in "introducing a structure for a vtable where all the functions have the right calling conventions"?

I thought about that before. However, I am not sure if it will do more harm than good (at least early in a decomp project): When you haven't implemented / worked out the signatures of all functions in that vtable yet, you'd lose Ghidra's own inference / heuristic on what the call signature likely is when a vfunction is invoked. The decompiler can also really get thrown off by wrong call signatures. So even if we did implement that, we may want to make it optional.

madebr: Good point. I don't know if Ghidra can even express this without confusing the decompiler.

```c
struct ParentVTable {
  void (__thiscall * destructor)(struct Parent*);
  int (__thiscall * method1)(struct Parent*, int arg1);
};
struct ParentFields {
  int m_member;
};
struct Parent {
  ParentVTable *lpVtbl;
  ParentFields fields;
};

struct ChildVTable {
  ParentVTable parent_vtable;
  int (__thiscall * method2)(struct Child*, int arg);
};
struct ChildFields {
  ParentFields parent_fields
  int m_member_child;
};
struct Child {
  ChildVTable *lpVtbl;
  ChildFields fields;
};
```

Going from a `Child *` to a `Parent *` will always require a cast.

jonschz: I created #551 to further discuss this point. Let's close the discussion here and reference the ticket in the code.

```suggestion
To represent inheritance, we create the base class struct (or use one created earlier in the run)
and set its position to the offset given in the PDB fieldlist. The other option is to copy members
from the base classes with new offsets, but we don't do it this way. See #551 for the reasons and future plans.
```

disinvite: I should have clarified my question. I was thinking specifically about polymorphism: if you don't use embedding for the struct, does _that_ affect the decompiler and give misleading or false results?

In the docstring I just wanted to note (for future reference) that there are two different ways to create the structs and one is not more correct than the other.

jonschz: > if you don't use embedding for the struct, does that affect the decompiler and give misleading or false results?

I raised that point in #551. Yes, it technically makes a difference. Let's say we have
```cpp
class B: public A {}

void f(A a);

B b;

f(b);
```
then the `f(b)` invocation will decompile to `f(b.base)` in one case and to `f((A)b)` in the other case. What I don't know (and what should be researched in #551) is whether the decompiler will actually stumble and fail or produce worse results in one of the variants.

tests/test_ghidra_integration_inheritance.py
disinvite: In practice, yes, because it's our main use case. If there's hard evidence out there that establishes what compiler does what, we should point to it. We may have to wait until we come across a compiler that does things differently that we want to support.

PR #530, opened 2026-08-31
Allow LINE markers inside function bodies again

foxtacles:
The #510 refactor replaced `marker.allowed_in_func()` with an inline tuple but dropped `MarkerType.LINE` from the set. A `// LINE:` marker inside a function body now raises a spurious `missed_end_of_function` warning and ends function tracking early, truncating the function's `end_line`.

Example from isle (`LEGO1/viewmanager/viewmanager.cpp`), which now fails `reccmp-decomplint --warnfail`:

```
 483 :  warning: missed_end_of_function  // LINE: BETA10 0x10172bbd
```

This restores `LINE` to the allowed set (the pre-#510 behavior) and adds a regression test asserting no alert and an intact `end_line`.

### DIFF
--- a/tests/test_parser_line_number.py
+++ b/tests/test_parser_line_number.py
@@ -291,3 +291,4 @@ def test_function_with_line_marker_inside(parser: DecompParser):
     assert len(parser.alerts) == 0
     assert parser.functions[0].line_number == 2
     assert parser.functions[0].end_line == 6
+    assert parser.lines[0].line_number == 4
--- a/tests/test_parser_line_number.py
+++ b/tests/test_parser_line_number.py
@@ -275,3 +275,19 @@ def test_string_line_continuation(parser: DecompParser):
     assert parser.strings[0].line_number == 2
     # TODO: enable when end_line is added
     # assert parser.strings[0].end_line == 3
+
+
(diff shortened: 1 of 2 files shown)


tests/test_parser_line_number.py
```diff
@@ -275,3 +275,19 @@ def test_string_line_continuation(parser: DecompParser):
...
+
+
+def test_function_with_line_marker_inside(parser: DecompParser):
+    """A LINE marker inside a function body belongs to that function
+    and must not end function tracking or produce a warning."""
+    parser.read(dedent("""\
+        // FUNCTION: TEST 0x1234
+        void test()
+        {
+            // LINE: TEST 0x1250
```
disinvite, line 287: May as well assert the `// LINE` position too while we are here. There is no getter in `DecompParser` that returns just the line markers; feel free to add one.

disinvite:
Thanks. That's what I get for not adding tests for `// LINE` to #516 (or ever)

tests/test_parser_line_number.py
foxtacles: Done in 291652e — added a `lines` getter matching the other symbol properties and the position assert.

disinvite (approved):

Issue #537, opened 2026-09-04
Code parser: scope applied to some comment annotations


disinvite:
In this example:
```cpp
namespace Pizza {
    // FUNCTION: TEST 0x1234
    // SampleFunction

    // VTABLE: TEST 0x5555
    // class Test
};
```
you get annotations for `SampleFunction` (no scope) and ``Pizza::Test:`vftable'`` (scope).

A coding oversight makes the behavior not consistent. Vtables and variables have scope added, other types do not.

We've used the fully qualified ``ClassName::`scalar deleting destructor'`` pattern countless times so the scale is definitely tipped towards "should not apply scope". In a situation where you have a namespace or class like this, you should probably just implement whatever is being annotated, except for synthetics like a constructor/destructor.

PR #538, opened 2026-09-04
Code parser: fix trailing white space bug (#184)

### LINKED ISSUE #184: Should trim whitespace on nameref annotations
If the name for your nameref annotation has trailing whitespace, we won't match the entity properly.

e.g.
```
// LIBRARY: 105 0x5EF7C0 SYMBOL
// _mbstowcs <-- with space
```

### LINKED ISSUE #512: Trailing space of a commented symbol annotation is not trimmed
Consider the following library annotation:
```c

// LIBRARY: MAIN 0x00402a4e
// printf                   
```

Notice how the line containing `printf` has trailing whitespace.

Running `reccmp` using this annotation results in the following diff:
```
0x40155b : ret 
0x40155c : push "HALLO 1\n" (STRING) 	(f1.cpp:26)
0x401561 : -call printf      (FUNCTION)
         : +call printf (FUNCTION)
0x401566 : add esp, 4
0x401569 : mov eax, <OFFSET10> 	(f1.cpp:27)
```

I'd expect the parser to:
1. warn about trailing whitespace (maybe optionally), and
2. trim the text for me (required)

disinvite:
Fixes #184 and #512.

While we're here, added tests for #537 since both fit into the "problems common to all nameref annotations" bucket.

### DIFF
--- a/tests/test_parser_common_nameref.py
+++ b/tests/test_parser_common_nameref.py
@@ -1,0 +1,98 @@
+"""Testing behavior common to all name annotations completed by a comment."""
+
+from textwrap import dedent
+import pytest
+from reccmp.parser.parser import DecompParser
+from reccmp.parser.marker import MarkerType
+from .parser_helpers import (
+    AnnotationType,
+    NAME_ANNOTATIONS,
+    completion_token,
+    xfail_param,
+)
+
+
+@pytest.fixture(name="parser")
+def fixture_parser() -> DecompParser:
+    return DecompParser()
+
+
+NAMEREF_TYPES = [marker_type for marker_type, _ in NAME_ANNOTATIONS]
+
+# fmt: off
+WHITESPACE_VARIANTS = [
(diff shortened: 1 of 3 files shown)


tests/test_parser_common_nameref.py
```diff
@@ -0,0 +1,98 @@
...
+
+    # We are not checking the specific nameref parsing rules here.
+    # Just make sure leading and trailing whitespace was removed.
+    assert symbol.name == symbol.name.strip()
+
+    # Undecided on whether we should warn, but we don't for now.
+    assert not parser.alerts
+
+
+# TODO: Replace with NAMEREF_TYPES when xfails are resolved.
```
jonschz:
```suggestion
# TODO #537: Replace with NAMEREF_TYPES when xfails are resolved.
```
nitpick

jonschz (approved):

PR #539, opened 2026-09-05
CodeView section contains UUID and Age

madebr:
With newer PDB's, the .debug section contains the GUID, the Age (=number of times the EXE/PDB have been compiled) and PDB filename.

This is actually a CodeView `LF_TYPESERVER2` type record.

### DIFF
--- a/reccmp/formats/pe.py
+++ b/reccmp/formats/pe.py
@@ -8,6 +8,7 @@ Based on the following resources:
 import dataclasses
 from enum import IntEnum, IntFlag
 from functools import cached_property
+from uuid import UUID
 from pathlib import Path
 import struct
 from typing import Iterable, Iterator, cast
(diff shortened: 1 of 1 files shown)


jonschz:
Is there a chance of regressions for old PDBs? Or did we just not extract the information correctly before?

madebr:
I think it was wrong before.
- [`CV_INFO_PDB20`](https://docs.rs/minidump/latest/minidump/format/struct.CV_INFO_PDB20.html): what is used in older PDBs
- [`CV_INFO_PDB70`](https://docs.rs/minidump/latest/minidump/format/struct.CV_INFO_PDB70.html): what is used in current PDBs

jonschz:
Do you have a modern PDB file at hand? I don't have a modern MSVC project set up at the moment.

madebr:
> Do you have a modern PDB file at hand? I don't have a modern MSVC project set up at the moment.

Download the [SDL3 Visual Studio development package](https://github.com/libsdl-org/SDL/releases/download/release-3.4.16/SDL3-devel-3.4.16-VC.zip). That contains 3 modern PDBs.

jonschz (approved):
I realised that I needed a modern DLL, not a modern PDB, but both were provided. Thanks! I can confirm that the new code works and the old version indeed failed to extract the PDB path correctly.

PR #543, opened 2026-09-06
Support pointers to non-exported data in `datacmp`

jonschz:
The current approach is flawed: In the LEGO1 example, we find a `_DIOBJECTDATAFORMAT *` and deduce that it points to some instance, but it actually points to an array (`_DIOBJECTDATAFORMAT[0x100]`). This cannot be reconstructed from the PDB.

Tasks:
- [x] Support annotating the type of a `// GLOBAL`
- [x] Revert deducing the type from the recomp parent type (since e.g. a pointer and an array are ambiguous)
- Add test coverage
  - [x] missing: get_by_name
- [ ] Document new features / annotations
- [ ] Rebase after #550


jonschz:
The following will work on `LEGO1`:

```
// GLOBAL: LEGO1 0x10098f80
// c_dfDIKeyboard

// GLOBAL: LEGO1 0x10097f80 TYPE=_DIOBJECTDATAFORMAT[256] NO_RECOMP_SYMBOL
// c_rgodfDIKeyboard

// If you enable this one, there will be no errors on LEGO1.
// If you disable it, you can see that the mismatches on `c_rgodfDIKeyboard` are displayed correctly.
// // GLOBAL: LEGO1 0x100db700
// _GUID_Key
```

@madebr do you want to take a look?

reccmp/compare/variables.py
```diff
@@ -321,31 +368,42 @@ def compare_variable(self, var: ReccmpMatch) -> ComparisonItem:
             # (i.e. if this is a static or non-public variable)
             # then we can only compare the raw bytes.
             compare_items = [
-                DataOffset(offset=i, name="", pointer=False) for i in range(data_size)
+                # FIXME type, temporary PoC
```
jonschz, line 371: This can probably be reverted. On the other hand, the `DataOffset` might just be redundant with `ScalarType`

reccmp/parser/parser.py
```diff
@@ -302,7 +300,21 @@ def _vtable_marker(self, marker: DecompMarker):
 
     def _vtable_done(self, class_name: str):
         for marker in self.tbl_markers.iter():
-            is_folded = marker.extra is not None and marker.extra.lower() == "folded"
+            # TODO: Rediscuss the syntax. I would actually prefer to change how we handle multiple inheritance here.
+            # Maybe
+            # // VTABLE: MYTARGET 0x1234 BASE_CLASS=MyBaseClass
+            # Would be a breaking change, to be discussed.
+            # For example, the current pattern does not work if a virtual base class is called "Folded"
```
jonschz, line 307: To be discussed. @disinvite do you have an opinion on this?

tests/test_parser_util.py
```diff
@@ -200,17 +202,17 @@ def test_marker_extra_spaces():
     """The extra field can contain spaces"""
     marker = match_marker("// VTABLE: TEST 0x1234 S p a c e s")
     assert marker is not None
-    assert marker.extra == "S p a c e s"
+    assert marker.extras == ("S", "p", "a", "c", "e", "s")
```
jonschz, line 205: Could this be a problem in some way? a key-value approach might need a way of escaping spaces, but that could also be overkill

disinvite:
Does the compiler drop unused types from the PDB? If not, users could just create whatever type/shape is needed:
```cpp
struct cKeyboardStruct {
    _DIOBJECTDATAFORMAT x[0x100]
};
```

Building a struct-name to leaf-id map doesn't add too much performance overhead. I ran a quick test yesterday with an extra regex search on each `TYPES` leaf for the `class name = ...` string.

jonschz:
> Does the compiler drop unused types from the PDB?

Unfortunately, yes. The following makes it to the PDB:
```cpp
struct cKeyboardStruct {
    _DIOBJECTDATAFORMAT x[0x100];
};

static cKeyboardStruct* keyboardStruct = NULL;
```
but the last line is necessary, which can have an impact on binary matching.

jonschz:
But even if we could make the "define your own struct in C++" approach work, it would get rid of the array type creation logic at most. The remainder of this change would still be needed.

disinvite:
Rough idea I have not thought through:
```cpp
// VTABLE: HELLO 0x1000 KEYVAL
// VTABLE: HOWDY 0x2000 KEYVAL
// VTABLE: GREET 0x3000 KEYVAL
// symbol = asdadfafadfadf
// name = Pizza:`vftable'{for XYZ}
// base_class = XYZ
```
Basically using the "extra" field to indicate that we're parsing the following comments differently.

disinvite:
> But even if we could make the "define your own struct in C++" approach work, it would get rid of the array type creation logic at most. The remainder of this change would still be needed.

True. I'm just looking for any way to reduce the surface area. And if we need to do this for a variable that has no PDB struct to attach to, we are stuck.

jonschz:
> Rough idea I have not thought through:

This approach looks promising to me. Some thoughts:
- Do we even need the `KEYVAL`? We could always try to detect an arbitrary number of key-value type comments below (regex draft: `\s*\/\/\s*(?P<key>\w+) =\s*(?P<value>(?<= )\S.*\S|\S|)\s*$`), and continue the existing parser after the first mismatch.
  - We could even start emitting deprecation warnings for the existing VTABLE pattern that they should migrate to a key-value based approach
- Is there a risk of a symbol annotation being misdetected as a key-value? This could happen if a `\s=\s` could be part of a symbol, which I don't think is possible. If there is an issue, though, the `KEYVAL` annotation in the first line won't save us (since we can't tell if the current line already is the symbol or just another key-value annotation)
- The new pattern would effectively mean "boolean flags are set in the same line as the annotation, key-value flags are set in comments below". Looks good to me, just wanted to state this explicitly so we can all agree/disagree on this pattern.

Edit: I read your comment again and now I think I understand where you are going with the `KEYVAL`. I'd prefer to keep the key-value per-annotation, even if that can lead to a bit more boilerplate. Sharing key-value attributes between multiple annotations feels more error-prone and harder to parse to me.

Edit 2: I wouldn't provide the symbol via key-value since that is against the general annotation pattern we have established:
```
// <ANNOTATION>
<TARGET>
```
where `<TARGET>` is either an actual C++ target or a comment of a target or symbol. If we introduced a `// symbol = ...` key-value annotation, we would have to conditionally parse the `<TARGET>` below, which I don't like.

disinvite:
We don't need `KEYVAL` if it we can distinguish a comment "completion token" from a comment with a key/value pair with perfect accuracy. Another option is to use an indicator character, as in:
```cpp
// FUNCTION: TEST 0x1234
// @test = 100
// @emote = 😎
// NameOf::TheFunction
```
but we know `@` can appear in symbols, and it can be the first character.

But what if there is no "completion token"? Are these still valid annotations?
```cpp
//// Completed by attributes only:
// GLOBAL: TEST 0x2000
// @name = MyGlobalVariable

//// Attribute and comment completion token in conflict:
// GLOBAL: TEST 0x3000
// @name = OtherVariable
// RealNameOfTheVariable

//// Attribute and code completion token in conflict:
// GLOBAL: TEST 0x3000
// @name = OtherOtherVariable
int CodeVariable = 5;
```

So it may be that I've introduced something that will take more effort to solve than the specific ask for the DirectInput keyboard struct.

jonschz:
Going full circle, how about one-line key-value with double quotes and backslash escaping, similar to JSON? (I have briefly considered proper JSON, but I think it's overkill and adds noise). This could look like
```
// the new thing
// GLOBAL: LEGO1 0x10097f80 TYPE="_DIOBJECTDATAFORMAT[256]" NO_RECOMP_SYMBOL
// c_rgodfDIKeyboard

// legacy vtable
// VTABLE: LEGO1 0x10097f80 MyBaseClass
class ChildClass: public MyBaseClass {};

// modern vtable
// VTABLE: LEGO1 0x10097f80 BASE_CLASS="MyBaseClass"
class ChildClass: public MyBaseClass {};

// hypothetical quote escape
// GLOBAL: TEST 0x1234 KEY_WITH_QUOTES_AND_SPACES="I said \"hello\""
```
Main benefit: This keeps syntax and parser complexity lower and is capable of representing all strings. We could use existing JSON logic to parse only the string, then we would already have support for other escape sequences like `\n` (if we ever need them, which I seriously doubt)

jonschz:
I implemented the JSON key-value approach. datacmp on LEGO1 will now pass with the following addition:
```
// GLOBAL: LEGO1 0x10098f80
// c_dfDIKeyboard

// this one can be disabled to verify that c_rgodfDIKeyboard is being compared correctly
// GLOBAL: LEGO1 0x100db700
// _GUID_Key

// GLOBAL: LEGO1 0x10097f80 TYPE="_DIOBJECTDATAFORMAT[256]" NO_RECOMP_SYMBOL
// c_rgodfDIKeyboard
```

reccmp/cvdump/types.py
```diff
@@ -11,6 +11,9 @@
     CvdumpTypeMap,
 )
 
+# TODO: Discuss if we want to split the file instead, and if so, what can/should be pulled out
+# pylint:disable=too-many-lines
+
```
jonschz, line 18: Needs opinion

reccmp/cvdump/types.py
```diff
@@ -443,10 +446,74 @@ def get(self, type_key: CvdumpTypeKey) -> TypeInfo:
...
+                return None
+
+            array_type = self.get_by_name(regex_match.group("name"))
+            if array_type is None:
+                # TODO report
+                return None
+            element_size = array_type.size
+            if element_size is None:
+                # TODO report
+                return None
```
jonschz: Address TODOs

tests/test_parser_util.py
```diff
@@ -196,21 +212,12 @@ def test_get_string_contents(line: str, expected: str):
     assert string.is_widechar is False
 
 
-def test_marker_extra_spaces():
-    """The extra field can contain spaces"""
-    marker = match_marker("// VTABLE: TEST 0x1234 S p a c e s")
-    assert marker is not None
-    assert marker.extra == "S p a c e s"
```
jonschz, line 203: semantics no longer relevant or desired. New semantics are covered by other tests.

tests/test_parser_util.py
```diff
@@ -196,21 +212,12 @@ def test_get_string_contents(line: str, expected: str):
...
-
-    # Trailing spaces removed
-    marker = match_marker("// VTABLE: TEST 0x8888 spaces    ")
-    assert marker is not None
-    assert marker.extra == "spaces"
```
jonschz, line 208: Covered by other tests

tests/test_match_msvc.py
jonschz, line 1: drive-by: better type safety. Functionality wise, this was only adapted to the signature change of `match_variables`.

reccmp/parser/parser.py
```diff
@@ -302,7 +334,32 @@ def _vtable_marker(self, marker: DecompMarker):
...
+                base_class: str | None = next(iter(marker.extra_flags))
+                extra_flags.pop()
+                logger.warning(
+                    'Legacy VTABLE base class annotation used above %s:%i. Change to `// VTABLE: %s 0x%x BASE_CLASS="%s"`.',
+                    self.filename.name,
+                    self.line_number,
+                    marker.module,
+                    marker.offset,
+                    base_class,
+                )
```
jonschz, line 358: We can move this to a second PR if you prefer, or even split into three (implement key-value, implement orig type annotation, change VTABLE annotation with legacy support).

If we keep everything in here, we should change the PR title.

reccmp/parser/marker.py
```diff
@@ -119,13 +151,39 @@ def match_marker(
...
+                        value, str
+                    ), "This assertion should never fail since the regex checks the presence of double quotes"
+                    extra_strings.append((match.group("key"), value))
+                except JSONDecodeError as e:
+                    logging.warning(
+                        "Invalid JSON in extra (last part) of annotation '%s'",
+                        raw_extra,
+                        exc_info=e,
+                    )
+                    return None
```
jonschz, line 183: Not sure if this would rather belong into `match_msvc.py` architecturally. Benefit would be that we could make decomplint failures out of these. The current approach has the benefit that all regex shenanigans are localised to this file.

reccmp/compare/db.py
```diff
@@ -20,6 +21,7 @@ def entity_name_from_string(text: str, wide: bool = False) -> str:
 }
 
 
+@dataclass
```
jonschz, line 24: generates `__eq__` which I need in a test

reccmp/cvdump/types.py
```diff
@@ -443,10 +448,87 @@ def get(self, type_key: CvdumpTypeKey) -> TypeInfo:
...
 
-    def get_by_name(self, name: str) -> TypeInfo:
-        """Find the complex type with the given name."""
-        # TODO
-        raise NotImplementedError
+    def get_by_name(
+        self,
+        name: str,
+        orig_addr: int,
+        report: ReccmpReportProtocol = reccmp_report_nop,
```
disinvite, line 455: I think we should just use the existing `logger` and not drag the report protocol into here. `reccmp-cvdump` fails with a circular import because you can't import from `reccmp.compare` and not get `.core` and all its dependencies.

reccmp/cvdump/types.py
```diff
@@ -443,10 +448,87 @@ def get(self, type_key: CvdumpTypeKey) -> TypeInfo:
...
+    ) -> TypeInfo | None:
+        """Find the class or structure with the given name."""
+
+        expected_leaf_fragment = f"class name = {name},"
+        potential_hits = [
+            self.get(key)
+            for key, (leaf, _) in self._raw.items()
+            if expected_leaf_fragment in leaf
+        ]
```
disinvite, line 514: We should cache this structure so we only pay for it once. Or not at all, if there are no name lookups.

reccmp/cvdump/types.py
```diff
@@ -443,10 +448,87 @@ def get(self, type_key: CvdumpTypeKey) -> TypeInfo:
...
+
+        # The same entry may appear multiple times (e.g. due to forward refs), so we deduplicate by key
+        # and also filter again by the name just to be sure
+        actual_hits = dict((hit.key, hit) for hit in potential_hits if hit.name == name)
```
disinvite, line 518: We don't currently do anything with the "WARNING: UDT mismatch" reports, but would that explain why we have some duplicate struct names? Maybe we could treat these cases (where the leaf does not have the `FORWARD REF` flag but the leaf id does not match the embedded UDT) as forward refs. Not sure how correct an approach that is.

reccmp/cvdump/types.py
```diff
@@ -443,10 +448,87 @@ def get(self, type_key: CvdumpTypeKey) -> TypeInfo:
...
-        raise NotImplementedError
+    def get_by_name(
+        self,
+        name: str,
+        orig_addr: int,
+        report: ReccmpReportProtocol = reccmp_report_nop,
+    ) -> TypeInfo | None:
+        """
+        Searches the type database for `name`.
+        Also supports arrays with decimal length (e.g. `MyType[20]`);
```
disinvite, line 459: Maybe this should be two functions: one to get the `CvdumpTypeKey` from the name, and one to create a synthetic array type where you pass in the key and number of elements you want. That would shift the user-provided type string parsing back to the caller.

reccmp/cvdump/types.py
disinvite: I think we can easily extract all the `read_(leaf_type)` functions and their companion regex strings into their own module.

disinvite:
Since we are adding new options to the markers anyway, would it be better to have "create synthetic matches from this variable" as one of the options and do it in the `Compare` pipeline instead of in the variable comparator? To match the current PR behavior, we would need to create synthetic matches recursively, although this is perhaps not what the user intends in every situation.

reccmp/parser/parser.py
disinvite: I think a split makes sense if it's not too much trouble.

jonschz:
> Since we are adding new options to the markers anyway, would it be better to have "create synthetic matches from this variable" as one of the options and do it in the `Compare` pipeline instead of in the variable comparator?

The difficulty is that we only find out the recomp address of the synthetic match when `datacmp` is run on the normal match that references a synthetic one. We'd have to annotate 
`c_dfDIKeyboard` that it _contains_ synthetic matches, and then always run a datacmp-like comparison on all entities with that annotation. Not sure if this is what you meant. Sounds like a lot of effort to me, especially since some datacmp-like logic will have to be duplicated or generalised to the Compare step. Might be the less hacky approach, what I don't like is the two new annotation types needed for that.

reccmp/cvdump/types.py
jonschz: The forward refs were the reason I introduced this code since the string match catches them. I don't have any new insight about the UDT mismatch. The warning about double matches is merely a precaution. I'd keep the other handling as it is.

jonschz:
Maybe a better alternative: 
```
// GLOBAL: LEGO1 0x10097f80 TYPE="_DIOBJECTDATAFORMAT[256]" SYNTHETIC_MATCH_FROM_POINTER="0x10098f94"
// c_rgodfDIKeyboard
```
The requirement would be that there is a different, valid global covering the `SYNTHETIC_MATCH_FROM_POINTER` address. We could then introduce a compare step that
- goes over all globals with that flag,
- searches for another matched global covering the referenced address
- get the recomp address from the match
- create a new entity based on that

We could then repeat this step until the number of matched globals no longer increases, which would cover recursive structures.

reccmp/cvdump/types.py
disinvite: #549

disinvite:
Rough outline of what I was thinking for matching inside the main compare loop: https://github.com/isledecomp/reccmp/compare/master...disinvite:reccmp:y-match-recurse  Some overlap with the problem in #385. (Ignore the weird tests for now. I didn't find a way to make them more idiomatic before moving onto something else.)

We will eventually want to do a similar kind of recursive matching with functions too.

jonschz:
> Rough outline of what I was thinking

Alright, I did understand you correctly. Let me summarize:

- We agree that the creation of the "synthetic matches" should be done in the main compare loop
- We just need to agree/decide how we want to annotate it

## Option 1 ([mentioned here](https://github.com/isledecomp/reccmp/pull/543#issuecomment-5739156523))

```cpp
// GLOBAL: LEGO1 0x10098f80 RECURSE
// c_dfDIKeyboard

// GLOBAL: LEGO1 0x10097f80 TYPE=_DIOBJECTDATAFORMAT[256] NO_RECOMP_SYMBOL
// c_rgodfDIKeyboard
```

- ➕ easy to implement, doesn't introduce annotation parsing complexity, PoC exists
- ➖ requires 3 new annotation types, one of them being on a type whose own matching behaviour doesn't actually change
- ➖ `c_rgodfDIKeyboard` (the line below the new annotation) is only a name with no meaning, in contrast to every other line below a marker
- ➖ the fact that `c_rgodfDIKeyboard` is matched based on `c_dfDIKeyboard` is implicit, which can lead to an (admittedly quite theoretical) issue outlined below:
  <details>
  Consider the following orig and incorrect recomp:

  ```cpp
  // orig
  const int DATA_1_WITHOUT_SYMBOL = 1;
  const int DATA_2_WITHOUT_SYMBOL = 2;
  const int* FIRST_TO_BE_MATCHED = &DATA_1_WITHOUT_SYMBOL;
  const int* SECOND_TO_BE_MATCHED = &DATA_2_WITHOUT_SYMBOL;
  
  // recomp
  const int DATA_1 = 1;
  

  // GLOBAL: TEST 0x1100 DATA_TYPE="int" NO_RECOMP_SYMBOL
  // DATA_1_WITHOUT_SYMBOL

  // GLOBAL: TEST 0x1000 RECURSE
  const int* FIRST_TO_BE_MATCHED = &DATA_1;

  // Note the incorrect recomp; this should be &DATA_2
  // GLOBAL: TEST 0x1004 RECURSE
  const int* SECOND_TO_BE_MATCHED = &DATA_1;
  ```

  Now whether `DATA_1` in the recomp would be mapped to orig `DATA_1_WITHOUT_SYMBOL` or `DATA_2_WITHOUT_SYMBOL` would depend only on the order in which `FIRST_TO_BE_MATCHED` and `SECOND_TO_BE_MATCHED` are processed in reccmp. Whoever gets processed first gets to decide what `DATA_1` will be mapped to. This isn't a problem if the decomp is correct, but if it is wrong, the behaviour can flip between compilations.
  </details>

## Option 2 ([mentioned here](https://github.com/isledecomp/reccmp/pull/543#issuecomment-5726409158))

```cpp
// GLOBAL: LEGO1 0x10098f80
// c_dfDIKeyboard

// GLOBAL: LEGO1 0x10097f80 TYPE="_DIOBJECTDATAFORMAT[256]" SYNTHETIC_MATCH_FROM_POINTER="0x10098f94"
// c_rgodfDIKeyboard
```

- ➕ easy to implement, doesn't introduce annotation parsing complexity
- ➕ requires only 2 new annotation types
- ➕ the pointer used for matching is explicit and stable (in contrast to the implicitness issue in Option 1)
- ➖ `c_rgodfDIKeyboard` (the line below the new annotation) is only a name with no meaning, in contrast to every other line below a marker

## Option 3 (new)

```cpp
// GLOBAL: LEGO1 0x10098f80
// c_dfDIKeyboard

// Option 3.1
// NORECOMPSYMBOL: LEGO1 0x10097f80 TYPE="_DIOBJECTDATAFORMAT[256]"
// c_dfDIKeyboard+0x14

// Option 3.2
// GLOBAL: LEGO1 0x10097f80 TYPE="_DIOBJECTDATAFORMAT[256]" SYNTHETIC_MATCH_FROM_POINTER
// c_dfDIKeyboard+0x14
```

- ➕ requires only 2 new annotation types (either two extras, or one extra and one new annotation type)
- ➕ The logic "the first line annotates 'where in orig', the second line annotates 'where in recomp'" is preserved
- ➕ the pointer used for matching is explicit and stable (in contrast to the implicitness issue in Option 1)
- ➖ the parsing logic for the second line might be more complicated

reccmp/parser/parser.py
jonschz: #550

disinvite:
I see "recurse" (or whatever its final name will be) serving two purposes:

- As a last resort. There is no other way to match the entities and it must be done indirectly.
- As a convenience. I want to match a web of entities from CRT or a library and I have high confidence that they will match or accept the risk that they will not.

In the case of `c_rgodfDIKeyboard`, or any entity where there is no name available from an "official" source like a PDB, the `// GLOBAL` annotation is there to:

- Flag it as a variable to be compared by `datacmp`.
- Give it some name, even though matching with that name is not possible.

This is one reason I wanted to add #371. A name that doesn't match anything may be intended by the user. For example: if a team has names for entities exported from Ghidra/IDA at the start of a decomp.

Either way, the name has to be there because this is the pattern we've established for the markers/annotations. It isn't possible to set attributes and no name in the current design.

Do we need the `no_recomp_symbol` option? It looks like it's used to:

- Suppress the "no match" warning (already hidden behind `--debug`)
- Be explicit about which addresses can be matched from another entity with the "recurse" option.

For the second case, do we need this protection if we acknowledge that the "recurse" option should be used sparingly and with intention?

The risk would be that you create a diff that is not exposed by any of the tools. If you match an address using "recurse" and the indirect address has no entity type (to flag it as variable) and no data type or size (so we know how many bytes to read) then you have a hidden diff.

jonschz:
> As a convenience. I want to match a web of entities from CRT or a library and I have high confidence that they will match or accept the risk that they will not.

Just to be sure I understood your use case correctly here: Do you want to do some "recursive auto mode" matching here, i.e. "these two entities match, try to match everything in it recursively"? If so, how do you want to infer the types of the new synthetic matched data? Just looking at our one example of `c_rgodfDIKeyboard`, the inferred type was `_DIOBJECTDATAFORMAT` but the correct type would have been `_DIOBJECTDATAFORMAT[256]`. I personally think that would be a high risk operation with lots of potential for false negatives. If teams don't want to bother with match errors in some library code, they can always just choose to ignore those globals or functions. I think it would also be difficult to document what an acceptable use case for `RECURSE` would be, and how a user can tell that it's doing the right / expected thing.

The approaches I suggested try to stick to reccmp's current philosophy of only matching what is explicitly annotated or can be inferred beyond reasonable doubt. That's the main reason I introduced `no_recomp_symbol` - it makes it explicit that this entity must be inferred. Options 2 and 3 I mentioned go even further and say from _where_ exactly it should be inferred, which is the closest we can get to a regular annotation (regular: the data found in recomp at X matches orig at Y; here: if you dereference the data found in recomp at X, it matches orig at Y).

disinvite:
> Do you want to do some "recursive auto mode" matching here

I was thinking about that, but let's simplify the matter and say that "recurse" only applies to the pointers from the struct in question. (Or, the variable itself, if it is a pointer.)

That means the name "recurse" no longer applies. 😃

> How do you want to infer the types of the new synthetic matched data?

Those would have to be annotated _if_ you want `datacmp` comparison. The user may not care and just wants a function that references this pointer to match. In the case of `c_dfDIKeyboard` we were forced to omit an annotation to avoid a CI blocker from `datacmp`.

> `The inferred type was ``_DIOBJECTDATAFORMAT`` but the correct type would have been ``_DIOBJECTDATAFORMAT[256]``

When creating new matches during the "recurse" step, we should only add the pointee type _if_ there is no existing value on the orig side. Otherwise, the user-defined `_DIOBJECTDATAFORMAT[256]` will be replaced by `_DIOBJECTDATAFORMAT`.

What if we require that new matches created by the "recurse" step have an "unmatchable" recomp entity? Meaning: either the entity does not exist in recomp at all, or it has no name or symbol.

jonschz:
> we should only add the pointee type if there is no existing value on the orig side

problem is that the automatic type could be incomplete (e.g. an array that's too small) and `reccmp`'s output would tell you "everything is alright and working fine", creating false negatives. I almost fell into this trap with my first draft in this PR. I'd prefer to have mandatory type annotations for synthetic matches, i.e. keep this new feature narrow and keep the risk of false negatives low. If the user just wants to force a match, they can pick an arbitrary type and then add the synthetic match to the ignore list. Then we're close to the Option 1 I mentioned above. I wouldn't mind going that direction, though I'd still prefer Option 2 since it requires no changes to how `c_dfDIKeyboard` is processed at all. But I also don't want this discussion to go on much longer, so let's make a decision.

PR #547, opened 2026-09-13
Introduce code tokenizer for scope detection and naming

### LINKED ISSUE #434: Function match fails because of namespace
Consider a global with bitfields, and a function that accesses it:
```c
struct GlobalState {
 unsigned char a : 1;
 unsigned char b : 1;
 unsigned char c : 1;
};

// GLOBAL: A 0x10007930
struct GlobalState g_state = {
  1,
  0,
  1
};

// FUNCTION: A 0x10001030
int get_a() {
  return g_state.a;
}
```

Then comparing that function fails because of namespaces:
```
$ reccmp-reccmp --target A -v 0x10001030
[INFO] Parsing /tmp/bitfield/build/a.pdb ...

---
+++
@@ -0x10001030,7 +0x10001030,7 @@
0x10001030 : push ebp 	(src.c:19)
0x10001031 : mov ebp, esp
0x10001033 : -mov al, byte ptr [GlobalState::g_state (DATA)]
           : +mov al, byte ptr [g_state (DATA)] 	(src.c:20)
0x10001038 : and al, 1
0x1000103a : and eax, 0xff
0x1000103f : pop ebp 	(src.c:21)
0x10001040 : ret 


get_a is only 85.71% similar to the original, diff above
```

Reproducing project: [bitfield.zip](https://github.com/user-attachments/files/28274400/bitfield.zip) (Built using MSVC6 RTM)
Run `01-build.sh`, `02-detect.sh` and `03-compare.sh`.

disinvite:
Part of #509.
Closes #434.

This is the first phase of implementing the token-based code parser for `reccmp` annotations. Here we use the tokens only to pair up curly brackets into scopes and find the correct name for attached classes, structs, or namespaces. The next phase will use tokens for all parsing of marker comments and the attached completion token. Adding this will require a total rewrite of `parser.py` and dropping tests are no longer relevant. To (hopefully) provide an easy review, I opted for the multi-phase approach that adds some value (and performance) but keeps the core file mostly intact.

The main unit is the `CodeToken`, an alias around the tuple of `(start_pos, end_pos, TokenType)`. Using `namedtuple` added too much overhead, and I found it was faster to read the desired index directly instead of destructuring the tuple. (e.g. `(_, __, token_type) = token`. Both of these concessions hurt readability, but we need to maintain performance at scale because this is one of our core operations.


tests/test_parser_lexical.py
```diff
@@ -0,0 +1,267 @@
...
+
+@pytest.mark.xfail(reason="TODO: #509")
+def test_string_completion_token_line_continuation(parser: DecompParser):
+    """Should support strings broken onto multiple lines with a line continuation mark.
+    The string text should not include any escaped newlines."""
+    parser.read(dedent("""\
+        // STRING: TEST 0x1234
+        const char* g_msg = "Hello \\
+        World";
+        """))
```
madebr, line 126: Perhaps also add a test with the string containing an escaped double quote?
e.g.
```c
        // STRING: TEST 0x1234
        const char* g_msg = "Hello \"\\
        World";
```

jonschz:
Just curious: How does the performance compare to a dataclass with `frozen` and `slots` enabled?

reccmp/parser/tokenizer.py
```diff
@@ -0,0 +1,522 @@
...
+L\"[^\"\n\\]*(?:\\.[^\"\n\\]*)*[\"\n]|
+\"[^\"\n\\]*(?:\\.[^\"\n\\]*)*[\"\n]|
+L\'[^'\n\\]*(?:\\.[^'\n\\]*)*['\n]|
+\'[^'\n\\]*(?:\\.[^'\n\\]*)*['\n]|
+\#\s*(\w+)(?:[^\n\\]+|\\\n|\\)*
+""",
+    flags=re.X | re.DOTALL,
+)
+
+CodeToken = tuple[int, int, TokenType]
```
jonschz, line 53: maybe add a doc comment below what the components mean, and that a tuple is used for performance reasons

disinvite:
> Just curious: How does the performance compare to a dataclass with `frozen` and `slots` enabled?

Slower, but we are talking about small time costs overall.

Averages from 10 iterations of tokenizing `LEGO1` source files:
```
tuples
AVG:     0.148348s    (0.146s, 0.155s)
MB/s:    19.682263

dataclass with slots
AVG:     0.181629s    (0.179s, 0.187s)
MB/s:    16.075687

frozen dataclass with slots
AVG:     0.255094s    (0.253s, 0.258s)
MB/s:    11.446056
```

I'd rather keep the faster tuple approach for now and revisit when we start moving code into C or Rust modules.

jonschz (approved):
I didn't read all the details, the code, integration, and tests look good.

* Usual question: Did you check for any change in behaviour on isle (LEGO1 + BETA10) and racers?
* Is there any particular part you want me to look at specifically?
* I don't think we need much robustness regarding syntactically invalid C++ code since that shouldn't compile anyway. Nevertheless, have you thought about how we should handle it? E.g. non-closed scopes.

disinvite:
I'll do another check of a few different decomps before merging. I have been checking `LEGO1` and `BETA10` through the whole process. If you like the direction then I think we are good. The various components speak for themselves, and I added as many of the tricky scenarios as possible to the tests.

As for handling broken input: we require that the file can compile. The user should be able to solve any parsing error by just editing the file.

The problem is preprocessor directives. There's at least one decomp that uses `#if 0` blocks with annotations, and if you're going to support that then you need to support `#if 1` too (for flip-flopping which block is enabled). That's done and will be coming in a future (smaller) PR after some cleanup.

We may find a situation where a user has to compose the file in a particular way to get all the curly brackets to pair up because we can't evaluate PPC expressions beyond `0` and `1`. Hopefully that doesn't happen, but if it does, we will still be in better shape than the current design that is full of holes.

PR #550, opened 2026-09-19
[breaking] refactor marker extras to keys with optional values

jonschz:
This is technically a breaking change for virtual inheritance VTABLE markers that contain special characters, e.g.
```
// VTABLE: TEST 0x1234 MyTemplateClass<int *>
```
which have to be changed to
```
// VTABLE: TEST 0x1234 BASE_CLASS="MyTemplateClass<int *>"
```
but I don't think those are widely used (if at all).

Split off from #543, required for non-exported data annotations.

The modified VTABLE annotation now supports
- folded vtables and virtual inheritance at the same time
- virtual base classes that are named "Folded" (for whatever reason)


tests/test_parser_util.py
```diff
@@ -196,21 +212,12 @@ def test_get_string_contents(line: str, expected: str):
...
-def test_marker_extra_spaces():
-    """The extra field can contain spaces"""
-    marker = match_marker("// VTABLE: TEST 0x1234 S p a c e s")
-    assert marker is not None
-    assert marker.extra == "S p a c e s"
-
-    # Trailing spaces removed
-    marker = match_marker("// VTABLE: TEST 0x8888 spaces    ")
-    assert marker is not None
-    assert marker.extra == "spaces"
```
jonschz, line 208: covered by other tests (https://github.com/isledecomp/reccmp/pull/543#discussion_r3998901331)

reccmp/parser/marker.py
```diff
@@ -119,13 +149,39 @@ def match_marker(
...
 
-    marker_type, target_name, offset_str, extra = match.groups()
+    marker_type, target_name, offset_str, raw_extra = match.groups()
     marker_type = resolve_alias(marker_type, target_name, aliases)
 
+    extra_strings: list[tuple[str, str]] = []
+    extra_flags: set[str] = set()
+    if raw_extra is not None:
+        if not FULL_MARKER_EXTRA_REGEX.match(raw_extra):
+            logging.warning("Invalid extra (last part) of annotation '%s'", raw_extra)
```
disinvite, line 163: Seems like this and other problems in the function should cause an `AlertCode.INVALID_EXTRA` error. We can't add one to the parser from here, but what if we raise an exception instead and then set the error in the caller?

reccmp/compare/db.py
```diff
@@ -20,6 +21,7 @@ def entity_name_from_string(text: str, wide: bool = False) -> str:
 }
 
 
+@dataclass
```
disinvite: I think this is leftover from the split. The test that needs `__eq__` isn't here.

disinvite:
Still processing this. I'll add more tomorrow.

I think we should try to find a way around `MarkerDict`. That may not be easy with the parser still being as state-dependent as it is. I ended up eliminating it in my tokenizer-parser branch (intended [follow-up](https://github.com/isledecomp/reccmp/compare/master...disinvite:reccmp:tokenizer-2x) to #547) in favor of grouping markers by category.

reccmp/parser/marker.py
jonschz: That would work. Any reason why we wouldn't want to just provide the callback as an argument to `match_marker`? It is available on the level above. Not sure if you made a comment in that direction on the other PR.

jonschz:
> That may not be easy with the parser still being as state-dependent as it is.

I wouldn't mind waiting with this task until your refactor has progressed further.

